PR #5, opened 2019-01-07
Permissions

salman-amjad:
These changes enables LAN server to authenticate with the remote server and filter data that gets synced.


mcccclean (changes requested):
I've left a few suggestions around things I'd change - I've only really had a brief look I'm sorry (interrupted by Tamanu meeting!) - so my feedback is mostly just at the code style level.

I'll be in on Wednesday to have a proper look at how this PR fits in to the wider system.

mcccclean (changes requested):

PR #1045, opened 2021-05-05
Fix migration

kael89:
Old habits die hard - we use singular table names in Tupaia...

I wondered how this worked the first time! 🤔 My guess is that because I was using `SQLite` locally, the migration didn't actually run and the database was updated based on the model changes


stretchkennedy (approved):
Oh no! I totally missed this. LGTM 👍 

The dev should use postgres by default, as of earlier this morning when I merged the PR.

PR #1266, opened 2021-07-26
Reconcile patient additional data

mcccclean:
Hopefully the final piece of the https://github.com/beyondessential/tamanu-backlog/issues/776 puzzle.

Main changes:
- add a new scheduled task to reconcile duplicate patient additional data records
- add a new `conflicts` field to the PatientAdditionalData model to track conflicts encountered during auto-merge

RAYG changes:
- fixed an issue where scheduled task names were displaying incorrectly on prod
- refactored the delete record method to be Model.delete(id) rather than Database.delete(model, id)
- made fetches for PAD records go through the patient model rather than bespoke queries each time

### DIFF
--- a/packages/lan/__tests__/apiv1/Patient.test.js
+++ b/packages/lan/__tests__/apiv1/Patient.test.js
@@ -69,7 +69,8 @@ describe('Patient', () => {
       expect(result).toHaveSucceeded();
 
       const id = result.body.id;
-      const additional = await models.PatientAdditionalData.findOne({ where: { patientId: id } });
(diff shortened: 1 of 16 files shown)


stretchkennedy (commented):
Couple more observations:
- we need to remove `conflicts` from the payload that's sent to lan servers; this is hopefully as easy as adding it to `excludedColumns`
- the lan server exports records first and then imports them second, so it's going to attempt to push the deleted-on-the-sync-server undeleted-on-the-lan-server records first and then pull them - no idea how this'll go down

PR #1281, opened 2021-08-03
889: Fix bug in report when there are multiple Fiji covid lab requests on same date for the same patient

chris-bes:


bgoepfert (approved):
Lgmt - i have restarrted the build in codeship (it was just a intermittently failing test that we are aware of that failed it)

PR #1370, opened 2021-09-02
#TAM-135 Added docker image doc

kujo4pm:
Docker documentation for the open source repo. 

Would like to check to make sure those steps are actually correct. @stretchkennedy 

Once this is approved I will chuck a copy of this file into our open source repo so it'll be visible ASAP - instead of relying on a release.


stretchkennedy (changes requested):
Mostly LGTM 👍  One question that needs clarification - I'll stamp it once that's addressed.

Also, am I right in assuming the audience is developers, sysadmins, and/or people who can get a dev to help them? If we're writing for a less technical audience we'll need a lot more detail and context, I think.

stretchkennedy (approved):

PR #1644, opened 2021-12-07
TAN-1064: introduce fix for error constraint

zavjs:
- fix error constraint on complete survey
- fix function runCalculations to correctly calculate fields.


stretchkennedy (changes requested):
I know this a big ask, but I don't feel confident in this functionality without tests around it. Would you be able to either 
- set up jest similarly to the other repos and write a test for it?
- create a card to add tests around this and assign me to it, then leave a comment explaining the inputs and outputs of the function?

PR #1694, opened 2022-01-13
WAITM-22 Respect Multiselect Changes

EMcQ-BES:
WAITM-22

Most of this was already working, just an issue with the `useEffect` causing the initial values (usually empty) to be loaded after every change

### DIFF
--- a/packages/desktop/app/components/Field/MultiselectField.js
+++ b/packages/desktop/app/components/Field/MultiselectField.js
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


mcccclean (approved):
I think there's still a minor issue here (even though it's not one that has been introduced with this PR) -- a component that uses `useState` to track its value internally needs to have a `useEffect` so that it can respond to changes in the `value` prop, so that forms can control it (most commonly with a "Reset form" button, but also any other dynamic logic).

I think most/all instances of this component wouldn't run into this issue anyway because in a survey you're remounting & unmounting these components as you skip through the pages.

Marking this as approved anyway (as like I said, the bug isn't new to this PR), but if you've got the space to have a quick look at while you're here that then it might save us a headache in the future.

mcccclean (approved):

PR #1755, opened 2022-02-03
RN-354 tamanu s3

IgorNadj:
Add s3 as a report recipient destination

### DIFF
--- a/packages/sync-server/app/env.js
+++ b/packages/sync-server/app/env.js
@@ -1,0 +1,9 @@
+import config from 'config';
+
+export const setupEnv = () => {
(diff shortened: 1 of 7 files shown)


edmofro (approved):
Glad this was straightforward - all looks good to me @IgorNadj! Nicely done, fast clean work :-)

tnguyen14 (approved):
I think this is fine. I'm not sure if there are other env vars that are needed when the sync server starts, and how they're being set currently. Maybe @stretchkennedy might know. If there aren't any, then this is fine for now.

PR #2190, opened 2022-05-05
WAITM-164 Print Lab Requests

EMcQ-BES:
- Add printout modal for lab requests
- Swap lab request buttons header to dropdown
- Swap new lab request finalise buttons to dropdown
- Allow linking directly to labRequest modal
- Set finalise and print to open print modal


stretchkennedy (changes requested):
Think we might be able to move some of the logic closer to where it's used, but let me know if you don't think it's worth refactoring and I can stamp this.

stretchkennedy (approved):
LGTM 👍

PR #2206, opened 2022-05-09
WAITM-165 Print Imaging Requests

EMcQ-BES:
- Make a generic `SimplePrintout` component
- Add `ImagingRequestPrintout`
- UI updates to add print buttons


mcccclean (approved):
Two comments but none blocking approval. (oh and that console.log but that doesn't need re-review)

mcccclean (changes requested):
One last change request (and pardon the long turnaround on this) - but it strikes me as a bit unusual that the printout components are responsible for displaying their own loading state, even though it's the containing component that is responsible for managing that loading state.

Basically I reckon:

```javascript
<MyComponent isLoading={isLoading} otherProps={otherProps} />
```

should generally become:

```javascript
isLoading ? <LoadingIndicator /> : <MyComponent otherProps={otherProps} />
```

mcccclean (approved):

PR #2272, opened 2022-05-26
WAITM-197 Sync notes with their parent records

EMcQ-BES:


stretchkennedy (changes requested):
Love the cleanup with `createNote`! 😄 

Couple of questions, and you might need to look at `getNotes` again (unless I'm missing something).

stretchkennedy (approved):

PR #2293, opened 2022-06-03
A couple of fixes for printing

EMcQ-BES:
- Add printing to changelog
- Copy updates
- Remove prescriber id until we add it to database somewhere
- Fix crash on new prescription
- Refresh lab request after adding a note


edmofro (changes requested):
Good changes, one suggestion

edmofro (approved):

PR #2307, opened 2022-06-07
Additional print fixes

EMcQ-BES:
- Fix usage of `refreshLabRequest` to use the full function name
- Copy updates on medication printout


tcaiger (approved):
LGTM

PR #2313, opened 2022-06-08
TAN-1104: Create unique indexes for permissions

tnguyen14:
- Minor changes to sync logic to make things clearer/ ignore `note` column
- Create unique indexes for permissions table


NavarroEmilioLuis (commented):
Hey, judging from the test, the indices are fine, I was kinda worried about the one that includes `objectId` had that `where` clause where it's null. But I guess that's fine?

In any case I think that you should consider renaming/dropping the name for the indices altogether, just for consistency. Also it seems that the migrations can't really run atm.

Lastly I think you'll need a code owner to review but I'll bump this again!

stretchkennedy (changes requested):
Mostly LGTM. The `log.info` output will potentially output a huge amount of data in some environments, and I've left a few style suggestions.

NavarroEmilioLuis (changes requested):
Just in case, reposting here as well:

Yikes, I know what's happening now. This branch doesn't include these changes: https://github.com/beyondessential/tamanu/pull/2312

That's why the migration fails to run. I'd suggest to bring that in, because otherwise it'll have a very wacky behavior, which will probably require yet another migration to solve that. With the field names fixed, the default index name should be all snake_case as expected, so you can also change it here.

Also, the merge should be now against branch `staging-1-18-0`!

Also, an example of index in migrations can be found in `migrations/068_addVdsNcSignersPeriods.js` and it doesn't need the explicit name. However I don't really feel strong about this, if you want to use it explicitly that's fine 👍

stretchkennedy (approved):
Fixed everything I mentioned, so I'm approving it. Should also get Emilio's approval before merging.

NavarroEmilioLuis (approved):
Looks good! I'm sorry for the trouble but I wanted to point out that the index name still has camelCase in it - I don't think it's a big deal, if you want to leave it as it is that's cool. Let's sort the merge plan in and we're OK to move on!

PR #2318, opened 2022-06-09
Attach initial notes to triages and lab requests

EMcQ-BES:
We have fields for initial notes when creating these records but they don't actually get created in the db

WAITM-219
WAITM-197


tcaiger (approved):
LGTM

PR #2381, opened 2022-06-20
Update CHANGELOG.md note to remove removeDuplicatedPatientAdditionalData

nick-beyondessential:
Add note to remove removeDuplicatedPatientAdditionalData scheduled task from Samoa


edmofro (changes requested):
@nick-beyondessential thanks for keeping things updated - could we keep things like this in the [release candidate PR](https://github.com/beyondessential/tamanu/pull/2378) from now on?

PR #2431, opened 2022-07-04
EPI-62 Admin importers

passcod:
Heavily refactors the importer internals, eliminating all the sync-calling code and reducing the amount of boilerplate. A notable change is that because we know we're running directly on Postgres, we run the entire import in a transaction, and dry runs simply rollback the transaction.

Significant changes:
- **Some** errors will drip feed (ie come one at a time) instead of showing up all at once, especially at the far end of the process when it's purely relational database checks. However most things should be checked and validated in bulk before that. From local testing it's exceedingly rare to have drip fed errors tbh.
- Some errors may be **false positives** because we try to go through as much of the data as possible, discarding the erroring bits but keeping on going for as long as we can. So e.g. if a vaccine refdata errors, the scheduled vaccine row that references it will throw a foreign key error — in this case a false positive that will go away once the first error is fixed.
- Upserted rows aren't returned, only stats ("this many rows in this model created, this many updated").
- Instead of an array that defines the order to load stuff in to avoid errors, a new structure defines the dependencies of data, and then the code resolves (very dumbly, no sat solvers were harmed in the making of this movie) the required order, which makes adding stuff a lot more explicit and less mistake-prone.
- The normalisation code for sheet names and foreign key column names supports a wider range of formats than before, including more plurals and casings/spacings.
- Fixed the in-code todo to check the `type` of reference data on foreign keys — as a result had to explicitly allow e.g. vaccine schedules to refer to either `drug` or `vaccine` refdata.
- Added capability to have more precise schemas for certain types of reference data. See for [manufacturers](https://github.com/beyondessential/tamanu/pull/2431/files#diff-ccfed23ae43c338d39a3c8d2485994a2e904befa39539702be0c8a1756a9439fR30-R33) for example.
- Removes the ability to ignore errors when importing: this was needed to work around foreign keys not resolving against the database, and as this PR does resolve FKs properly, this is no longer needed. This "feature" could have also been (mis)used to "ignore" certain rows in the source data by making them fail format validation and then ignoring errors, but that seems very brittle so I haven't made an effort to keep it working.

Some test data:
- [ManufacturersRefdata.xlsx](https://github.com/beyondessential/tamanu/files/9177443/ManufacturersRefdata.xlsx)
- search for "tamanu demo reference data" in google drive
- the test files in `packages/sync-server/__tests__/importers/refdata-*.xlsx`

Outstanding things that will be done in follow up PRs:
- EPI-73 fixing the other importers (surveys, permissions) and their test suites
- Permission check on the importer endpoints (no ticket, dependent on sync-server permission work)


mcccclean (changes requested):
Mammoth effort on this! Great stuff.

mcccclean (approved):
Code-wise this all LGTM but I'm a little bit reluctant to merge it without the permissions + surveys functionality intact; they're both critical and while I'm confident there's enough time to finish implementing those before 1.20, merging really locks us in to delivering that in a way that's a little intimidating.

How do you feel about putting this in a feature branch? (Or declaring this one the feature branch that the perms + surveys stuff subsequently gets merged into?)

PR #2433, opened 2022-07-04
WAITM-294 Allow configured reports to run in local time zone on sync

EMcQ-BES:
- Add a field to config to define local time reports
- When spawning a new report sub process, set sequelize to read in the local time zone

### DIFF
--- a/packages/shared-src/src/services/database.js
+++ b/packages/shared-src/src/services/database.js
@@ -57,6 +57,7 @@ async function connectToDatabase(dbOptions) {
     host = null,
     port = null,
     verbose = false,
+    timeZone = null,
   } = dbOptions;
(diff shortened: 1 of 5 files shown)


nick-beyondessential (approved):

mcccclean (changes requested):
Pushing back gently on the specific approach of how the TZ gets passed to the database init stuff, but open to be persuaded.

I'd also point out that it's not guaranteed that reports run in a separate process! The `config.reportProcess.runInChildProcess` flag defaults to true, but can be set to false, and when it's false it just uses the existing DB connection. I don't think we need to fully support that circumstance as part of this PR, but it'd be good to add a check against this and throw an error when we're trying to run a `localTimeReport` when `runInChildProcess` is false, save us tripping over that in future.

PR #2580, opened 2022-08-09
EPI-83: New sync model on mobile

chris-bes:
### Changes:
- Integrate the new sync model change from sync server/facility server to mobile (please read card description for more details)

### Screenshots:
We no longer have sync channels, so the sync screen progress will be a bit different than before. We have 3 stages:
- Stage 1: Pushing all records (no screenshot coz it's often too quick)
- Stage 2: Pulling all records:
Image from iOS (4)

- Stage 3: Saving records (will show separate screens for CREATE/UPDATE/DELETE)
Image from iOS (8)
Image from iOS
Image from iOS (5)

- Sync finished
Image from iOS (9)

### TODO:
- Add more documentation
- Write tests
- Clean up the branch a bit more


edmofro (changes requested):
First card back at BES, we throw you a doozy, and you've absolutely SMASHED it!!

Amazing work @chris-bes 

Two general things I'd like to suggest we change, both in the direction of simplicity:
- Progress message: the end user doesn't really care what records are being saved, and now we've moved away from channels it's easier just to use an overall total for the progress of saving the changes. Although it's useful during our diagnostics sometimes to know what model the user was up to when it failed, hopefully we can get that same utility from sending the logs to Honeycomb instead. 
- Failure handling on saving incoming records. Now that the sync is atomic, we should fail the whole lot if there's any failure, so let's not go to the effort (and code complexity) of trying each record individually, and instead just fail if we hit any error. No need to pass the `failures` array back up the chain, just throw the error and catch it at the highest level, and use that error message.

PR #2945, opened 2022-10-10
Staging fix: TAN-1852: Fix permission revocation during import

mcccclean:
### Changes

Same PR as https://github.com/beyondessential/tamanu/pull/2944 but against staging this time.


passcod (approved):

edmofro (commented):
The squash and merge on dev and here is going to cause a different git history and conflicts huh? Easily resolvable conflicts (because same exact change both places), but conflicts all the same, which means merging from staging to dev will first involve updating staging from dev and resolving conflicts locally via a third branch?? What do you think about a cherry pick onto staging instead?

PR #3050, opened 2022-10-26
Sepi alastair emilio ethan killer feature

NavarroEmilioLuis:


NavarroEmilioLuis (commented):
Just leaving random notes for the future just in case we actually want to do something serious out of this

PR #3206, opened 2022-11-23
Staging fix EPI-171: Remove writeToPatient work around

chris-bes:
### Changes
- Remove writeToPatient work around

### Checklist
- [x] Code


edmofro (commented):
Ok so this one I'm not sure about. It's unrelated to anything going out this release, so isn't a regression, and Da only found the positive result on desktop because there was a separate (actually related to the new code) bug on mobile that was showing null...I actually think that's not really a bug, it's expected (even desired) behaviour on desktop. Although we do now write to PAD directly from survey responses, there may be historical ones we haven't migrated? So will leave this in the hands of WAITM and @tcaiger, but either way I don't think it needs to rush into this staging branch. Thanks though @chris-bes, you've done a stellar job of following up on issues as they've come up today.

PR #3214, opened 2022-11-24
TAN-1911: Write script to rectify Fiji Aspen db record times

stretchkennedy:
### Changes

**Don't merge!** This is a script to be run manually in Fiji Aspen to set all the timestamps for everything backwards. We need to do this for complicated client-relationship reasons that Megs understands better than I do. It's in a PR because this way it's easier to review.

Because it's changing updatedAt and createdAt, our plan is to:
1. Take the central server down
2. Apply this to the facility and central DBs so everything ends up in the same state
3. Turn it back on

Review covers both "is this doing what the card asks?", "has Kennedy missed anything really important that makes this a bad idea?", and "is the operational plan sound?"

### DIFF
--- a/manual_fiji_dst_rectification.sql
+++ b/manual_fiji_dst_rectification.sql
@@ -1,4 +1,11 @@
 -- encounters
+SELECT COUNT(*)
+FROM encounters
+WHERE updated_at >= '2022-11-13T01:00:00+12:00'::timestamptz AND
+      updated_at < '2022-11-14T07:30:00+12:00'::timestamptz AND
+      created_at >= '2022-11-13T01:00:00+12:00'::timestamptz AND
+      created_at < '2022-11-14T07:30:00+12:00'::timestamptz;
+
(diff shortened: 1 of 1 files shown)


mcccclean (approved):
Code all LGTM (assuming it lines up with Aspen's requirements, which I'm not entirely familiar with).

My only question would be, do we want a way to back out of this easily? Obviously we could revert to a backup if anything goes wrong, but thinking about the scenario where we notice something was awry two weeks from now it'd be a real pain.

I was thinking we could like, construct a bespoke table at the start of this query that tracks table name + id, and then use 

```sql
INSERT INTO aspen_affected_records 
'xyz', id from (
  UPDATE xyz
  SET [[ do the date adjustment here ]]
  RETURNING id
);
```

Something like that??? Haven't checked syntax or anything. Then we'd have something to draw on if anything did go wrong. But... entirely possible I'm being overly paranoid here.

PR #3300, opened 2022-12-08
EPI-301: Use Github Actions for CI

passcod:
### Changes

- Replaces the CI (tests) portion of Codeship with [GHA](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) instead.
- Leaves the CD (packages and deploys) part relatively intact (see rationale below).
- Speeds up cold runs by about 15 minutes, and hot runs by about 5 minutes.
- Migrations are tested on both PostgreSQL 12 and 14 (instead of just 14). Trivial to add more versions.
- Safe and structured concurrency: each parallel job is completely isolated, so there's no longer issues running e.g. mobile and desktop tests in parallel

- Granular and clear summary of what's running/succeeding/failing, both on the dedicated checks page and directly on the PR. This brings even more perceived speed: the mobile tests finish 3 minutes after push, and lints after 5 minutes, and that's immediately visible even as the rest of the run is still pending.

- The lint and coverage (as we've discussed, only really used as a "tests have passed" indicator) messages are gone, essentially replaced by the above. The `!lint-fix` command is still there, somewhat less discoverable now.

- Can restart just the failing tests:

- Can launch a CI run manually from the Github UI on any branch/tag, even without a PR. Sometimes useful for experiments.
- Ongoing CI runs on the same branch/PR are cancelled if a new commit is pushed / a new run is started.

#### Why only CI?

Instead of moving everything over at once, let's merge just the CI part for now! First, that avoids a stale PR, but most importantly it lets us see whether this is working out for us while having a relatively painfree rollback if we want.

One thing that changes is that the CI and CD are decoupled: packages will be built and deploys launched almost immediately, and even if the test suite fails. If we only merge good stuff to the main branches then there should be no issue (and it will improve velocity for testers a bit). We can rollback or rollforward (move CD sooner) if we have problems.

### Checklist

- [x] Code is finished


edmofro (changes requested):
This is great! The GitHub Actions format is really easy to follow, and some nice patterns (like the matrix thing, I love that).
Only one suggestion requires a response, which is to add comments into the new bash script to make it easier to follow

edmofro (approved):

PR #3310, opened 2022-12-08
EPI-399: Fix intermittent failing FHIR patient test

NavarroEmilioLuis:
### Changes

Current dev shows a build that failed filtering by `given`, this fixes that and more...

This was a pretty rough thing to investigate but there are some weird issues when querying the results through Sequelize. Notice that some of the updated tests are still skipped.


NavarroEmilioLuis (commented):
Left some notes of what got changed as I think the reasoning is not that obvious. Pretty much everywhere there are some other fields that we weren't considering on the tests, causing it to produce unreliable results.

The tests usually follow the same format, since the order is mixed between two fields, we'll usually go with 3 rows of data, using an obvious ordering but switching the second record fields. i.e.:
row   | field one | field two |
   1          A              E 
   2          F               B
   3          C              G

However, some tests in the sorting section are still failing because of the way it matches unnested fields (from what I could observe): it creates a list and then matches every record respectively, sometimes including the same row twice or more because of it. There's also a very weird error with family descending that doesn't happen backwards.

Anyway, all tests are still skipped for now but they probably now have the proper info we should expect.

passcod (approved):
sorry this got lost!

PR #3337, opened 2022-12-13
EPI-305: Mark patient for sync on creating encounters etc.

edmofro:
### Changes

This was a missed feature from the first pass at v2 sync, d'oh

Also cleaned up the `/patient` endpoint so that it doesn't deal with marking for sync, as it was always a bit of a weird responsibility but definitely makes sense to have its own endpoint now it is backed by a db table.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
I have a couple of questions, the latter one might be the only important one as it might make some tests unreliable

NavarroEmilioLuis (approved):
Great!

PR #3339, opened 2022-12-13
Add some fields to HTTP logs

passcod:
### Changes

This started as just something to add (the last bit of) the user ID to the http log, but I added a bunch of other fields and fixed others:

- User ID (only the part after the last `-`, which makes it shorter and, with UUIDs, still relatively collision-proof). Makes it easier to:
  - (on Facility) estimate how many users are using Tamanu right now
  - (on Facility) filter by one particular user to see "a session"
  - (on Central) see which facility is talking to the server (assuming each facility has its own login)
  - _note_: this is probably the most contentious of these new fields
- Processing time (was: "response time"). This is an existing field but "renamed" (though it didn't really have a name before). Essentially that field measures the time it takes for the server to process the request, from when we start receiving it, to when we start sending the response.
- Sending time. This is the time between starting sending the response and finishing sending the response.
- Content Type. This is the second bit of the content-type response header. So generally it will be `json`, but could be `pdf`, `png`, `plain`, maybe even `html`, on occasion.
- Bytes received. This is actual bytes read from the socket for this request, without depending on or trusting request headers. Note this includes headers.
- Bytes sent. The same but for bytes written. This replaces reading the `content-length` header, which generally didn't work.

A new logging line would look like this:

```
2022-12-13T03:02:05.267Z http: ::1 GET /v1/patient?page=0&rowsPerPage=10&order=asc&matchSecondaryIds=true - status=304 bytes-recv=2125 bytes-sent=3887 content=json time-proc=10.251ms time-send=0.307ms user=147e86e3c14b
```

Here broken down across lines for clarity, showing an unauthenticated request (login) and then with the user ID (user/me):

```
2022-12-13T03:02:05.188Z
	http: ::1 POST /v1/login
    status=200
    bytes-recv=720
    bytes-sent=2997
    content=json
    time-proc=268.100ms
    time-send=2.201ms

2022-12-13T03:02:05.193Z
    http: ::1 GET /v1/user/me
    status=200
    bytes-recv=1394
    bytes-sent=3592
    content=json
    time-proc=2.489ms
    time-send=0.385ms
    user=147e86e3c14b
```

### Checklist

- [x] Code is finished

### DIFF
--- a/packages/shared-src/src/services/logging/color.js
+++ b/packages/shared-src/src/services/logging/color.js
@@ -6,8 +6,8 @@ const {
   color,
 } = config?.log || {};
 
-const colorise = color 
-  ? (hex) => chalk.hex(hex) 
+const colorise = color
+  ? (hex) => chalk.hex(hex)
   : (ignoredHex) => (text => text);
 */
 // TEMP: Disable chalk & its import as it seems to not run correctly on AWS
@@ -19,4 +19,5 @@ export const COLORS = {
   blue: colorise('729fcf'),
   red: colorise('ef2929'),
   yellow: colorise('e9b96e'),
+  magenta: colorise('ff00ff'),
 };
--- a/packages/shared-src/src/services/logging/console.js
+++ b/packages/shared-src/src/services/logging/console.js
@@ -22,7 +22,7 @@ const additionalDataFormatter = (obj = {}) => {
 const logFormat = winston.format.printf(({ level, message, childLabel, timestamp, ...rest }) => {
   const restString = additionalDataFormatter(rest);
   if (restString === '') {
-    return `${COLORS.grey(timestamp)} ${level}:${childLabel ? `${childLabel} - ` : ''}${message}`;
+    return `${COLORS.grey(timestamp)} ${level}: ${childLabel ? `${childLabel} - ` : ''}${message}`;
   }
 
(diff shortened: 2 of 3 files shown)


NavarroEmilioLuis (approved):
Changes look good to me. Probably better idea is to have @mcccclean or @stretchkennedy to bounce if there might be something extra or missing in there (but that can be done later I guess)

mcccclean (changes requested):
Nice stuff! One small change and it's good to go.

I have one concern that is like... this change isn't _super_ useful until the data is being sent to honeycomb, just because remoting in to look at logs directly is a hassle + we don't have the aggregation tools handy. But the way that this logs everything doesn't send to honeycomb very nicely? Everything ends up preformatted in the `message` field rather than being sent as structured data. But that issue predates this PR, and I don't think it's really exacerbated here, so I don't consider it a blocker, the work to change it to structured will be the same whether it's done now or later.

Also worth noting that we're not sending http logs to honeycomb at all yet, but that's largely because we don't have a good read on what our logs volume is yet as there're still a lot of servers left to hook up to it and we don't want to accidentally smash our usage limits, it's definitely still the plan to add them eventually.

mcccclean (changes requested):
Yeah agreed, definitely a step forward, structured data can happen later for sure.

Change request was just the http syntax error - will approve as soon as tests pass.

mcccclean (approved):

PR #3340, opened 2022-12-13
TAN-1877: update encounter summary report p2

Alastair-L:
### Changes

Card: https://linear.app/bes/issue/TAN-1877/updates-to-encounter-summary-report
This is part 2. Part 1 was here: https://github.com/beyondessential/tamanu/pull/3291

This includes the following changes:
1. **Separating the encounter type column into three new columns with to show Triage, Inpatient, Outpatient (highlighted in green).**
The logic here is that we have the most recent encounter type of each category "Triage, Inpatient, Outpatient".
2. **There seems to be a bug in the report where if a patient has moved departments or locations the move is shown in the report twice (Two patients showing this bug are highlighted in red). Separate the department and location 'assigned' times into separate columns.** 
This PR fixes the third of three causes of this (see the previous PR for details)
6. **Include patient insurance details in the report** 
It was decided to do this with config - so that other deployments would be able to provide their own patient insurance details ids (or other fields)

Note: there is an SQL injection if an attacker manages to modify the config file, however there are other reports with the same vulnerability, and the config file contains the database password anyway, so I think at that point we're doomed either way. Please feel free to push back on this if you disagree though.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


nick-beyondessential (approved):
Codeowner viewed

PR #3345, opened 2022-12-14
WAITM-525: Get latest encounter for vitals

tcaiger:
### Changes

When submitting a survey or vitals on mobile, we try to add the survey response record to an existing encounter if possible and if not we create one.

However the current logic for getting an encounter is a bit strange as follows:
- Select all the encounters after 3am today for the patient
- Regardless of whether there is an endDate or not
- Get one at random of them

This change updates the logic to be:
- Select all the open encounters for the patient (ie. ones where there is no end date)
- Get the latest one of them (ie. Order by startDate)

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/mobile/App/models/Encounter.ts
+++ b/packages/mobile/App/models/Encounter.ts
@@ -22,8 +22,6 @@ import { SYNC_DIRECTIONS } from './types';
 import { getCurrentDateTimeString } from '~/ui/helpers/date';
 import { DateTimeStringColumn } from './DateColumns';
 
-const TIME_OFFSET = 3;
-
 @Entity('encounter')
 export class Encounter extends BaseModel implements IEncounter {
   static syncDirection = SYNC_DIRECTIONS.BIDIRECTIONAL;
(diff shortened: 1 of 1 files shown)


mcccclean (changes requested):
We still need the 3am filter, I think. The logic is:
- Tamanu mobile doesn't have the auto-discharger task running locally, the only way an encounter gets closed is by syncing up to the central server, getting closed there, and then syncing back down
- A mobile client might go several days in offline mode
- So, if I show up for a vaccination on Monday and then a screening survey on Thursday, my Monday encounter will still be open, but my screening survey should be attached to a new encounter, not Monday's one
- So, we need some logic that only selects an open encounter _from today_

Alternatively we could build an encounter-discharger task to mirror the one that happens on the sync server but it'd be a bit awkward - the logic would be happening in two places, and the timing of it would be a bit unreliable as people don't have Tamanu open on their devices 24/7. 

Might need to consult Megs on exactly what we need out of this as it seems like there're some conflicting requirements happening.

PR #3349, opened 2022-12-14
Staging fix: use patient.id differently in PatientIssue than other models

edmofro:
BUT WHY 😱


mcccclean (changes requested):
Good find but I think this fix is too weird to just implement without any other mitigations.

I reckon it needs:
- short explanatory comment acknowledging that it's to get around a bit of typeorm strangeness
- linear ticket to investigate that strangeness
- link to the ticket in the comment

mcccclean (approved):
Heyyy great result!

PR #3351, opened 2022-12-15
SAV-23: Clean up lab request suggesters

mcccclean:
### Changes

The LabRequest form previously used a nonstandard approach for getting its data values, going through some bespoke endpoints instead of the suggester API.

A couple of these calls had already been switched to the suggester API, so this PR deletes those old endpoints, and moves the last one (the lab tests themselves) over to the suggester API as well. Currently lab tests only use the `/all` endpoint so in the interest of treading lightly I only implemented that part. If we need to do text search for lab tests in future we can implement it then, but they're already filtered by category so it's unlikely we'll need to in the near future.

### Checklist

- [x] Code is finished
- ~[ ] Tests are written~ No tests as it's just paying off tech debt

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM left two minor suggestions (though we still probably need to remove some unit tests)

PR #3366, opened 2023-01-05
EPI-110: Modal extension on record death workflow

NavarroEmilioLuis:
### Changes

This includes allowing a partial workflow: being able to register a death with only the time of death + clinician and allowing the user to come back and fill the rest of the information later.

Ended up being a little more laborious than expected because of the weird structure but hey, should be good. Don't panic over the lines of code, there were a couple of files that got renamed or moved code to its own file.


stretchkennedy (approved):
LGTM, couple of nits that I'm still happy to merge with

PR #3369, opened 2023-01-06
EPI-157 + EPI-188: Fix modal sort, correct copy, and fix PDF

stretchkennedy:
### Changes

This PR updates the copy (which had three different words to refer to the `requestedDate` column), adds a sort order, and rewrites the PDF to use normal tables instead of grid layout.

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM.

Left two nits (not sure if they're even that):

- Wondering if there's any reason to use `paddingLeft` inline instead of declaring with the component.
- Manually having to declare the width seems a bit obnoxious? ie. having to add another column means doing the calculation and manually changing it everywhere. I wonder if we could have something that calculates the width based on the number of columns, each column could have a default size and we would only have to specify the ones that take up extra amount.

But I understand if it might be a bit pointless if we're not planning on touching this component too much, so I'm happy for this to be merged as is - if you think it's worth adding maybe give it a go.

PR #3372, opened 2023-01-08
EPI-136: Add missing modifier on datetime field

passcod:
### Changes

Add `saveDateAsString` so it works properly

### Checklist

- [x] Code is finished


dannash100 (commented):
🎄  Thank you for this

dannash100 (approved):

PR #3373, opened 2023-01-08
CI: Add dummy job for tests-pass status

passcod:
### Changes

Stuff learned from OSS! This is a dummy job with a set name that depends on all the tests we require to pass. This makes it simple to add just this job to github's branch protection requirements, and the actual set of requirements needed to achieve a pass is then essentially defined in code. This also future-proofs changes to the CI config by decoupling the exact test layout from the PR requirements in less accessible github config.

image

image

<sup>(if we enable this, let's _not_ require branches to be up to date... fine for small projects but makes it really tough to merge things at scale)</sup>

image

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,3 +215,15 @@ jobs:
 
     - run: yarn test
       working-directory: packages/mobile
+
+  # Dummy job to have a stable name for PR requirements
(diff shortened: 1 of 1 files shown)


edmofro (approved):
Nice :-)

PR #3392, opened 2023-01-12
Master to Staging: 1.23.3 hotfix

EMcQ-BES:
### Changes

- Bring the 1.23.3 hotfix change onto staging
- Fixup version number clashes


dannash100 (approved):
Kind of hard to be ontop of exactly the moves here but ill give the tick :)

PR #3393, opened 2023-01-12
SAV-51: staging-fix: don't override incorrect cred error message

dannash100:
### Changes
Saw this same logic used in mobile to deal with 401 error messages
on login endpoint it refers to incorrect credentials
and when a user is logged in an expired token will also send 401

### Checklist
Staging fix
- [x] Code is finished

_Upon merging:_

### DIFF
--- a/packages/desktop/app/api/TamanuApi.js
+++ b/packages/desktop/app/api/TamanuApi.js
@@ -232,7 +232,7 @@ export class TamanuApi {
     }
 
     // handle auth expiring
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (commented):
This solves it partially but it strikes me as weird that we have to come up with the error messages in the frontend? Aren't we sending that exact message from the facility server too?

NavarroEmilioLuis (approved):
Having a second look, we've been doing this for way too long. This is fine 👍

PR #3397, opened 2023-01-13
Maybe fix second facility deploys

passcod:
Not sure this will work, but it's the only way I can maybe guess that [what we're seeing](https://beyondessential.slack.com/archives/C03KJSEPV9A/p1673578360093069?thread_ts=1673575730.362339&cid=C03KJSEPV9A) could be happening.


EMcQ-BES (approved):
Admittedly my bash isn't the strongest, but I think this looks good

PR #3400, opened 2023-01-13
EPI-136: Use separate completed_at datetime field instead of created_at

passcod:
### Changes

Created_at doesn't survive sync intact. Also lets us use a DATETIMESTRING type instead of a TIMESTAMPTZ.

### Checklist

- [x] Code is finished
- [ ] Tests are written


NavarroEmilioLuis (approved):
Makes sense! Kept thinking a bit about having the default value for `completedAt` but I guess it doesn't hurt?

PR #3401, opened 2023-01-13
SAV-48 + SAV-95: More new sync unit tests

chris-bes:
### Changes
- More new sync unit tests in both facility + central
- Checklist is in the ticket + some more tests added along the way

### Checklist
- [x] Code is finished
- [x] Tests are written

_Upon merging:_


edmofro (changes requested):
Thanks so much for this @chris-bes! I have a few very minor suggestions inline, but would also like to talk about some advanced tests. In general I think these are good targeted unit tests, but in being so targeted they focus a little heavily on the implementation rather than the behaviour. There are some interaction behaviours that would benefit from tests that sit at a slightly higher level: e.g. concurrent syncs actually occurring and not disrupting each other or other additional changes. These are the sorts of edge cases we spent quite a while fixing during the sync rewrite, and tests could help from letting them slip back in. Would suggest we have a call for a design session on some of these more advanced cases - and if it turns into enough to warrant another ticket, I'm happy to do that too

edmofro (changes requested):
Thanks for this amazing effort @chris-bes, I really really appreciate your hard work here 🙏

edmofro (approved):
Looks really, really good to me

Thanks for your patience on reviews and commitment to the cause!!

edmofro (approved):

PR #3405, opened 2023-01-15
WAITM-605: Refresh token for central server and utilize in mobile central server connection. 

dannash100:
### Changes

Add refresh-token flow to sync-server for use by Mobile client

- Use a long living refresh token to keep central server connection alive when connected via mobile
- Revolve refresh token for added security when it is used to reissue token ( This step is optional and might be worth some discussion - weither refresh token needs to be regenerated on every hit of `/refresh` endpoint). 
- Add tests for CentralSyncManager

Rethinked flow:
- Add issuer, audience and 32bit int jwtid to jwt shape
- Refresh token in db store bcrypt hashed RandomBase64String - along with a composite unique index on `user_id` and `device_id`.
- Refresh token transported using jwt containing refreshId and userId and signed with a seperate secret
- Refresh token contents are verified and audience, issuer are verified
- userId is extracted from contents - `refresh_tokens` table is queried by userId and deviceId of request - and refreshId is verified like a password using `bcrypt.compare` against hashed refresh_id.
- If all looks good new access token is issued and refresh token is rotated

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
It feels a _little_ off to me that we're storing the entire text of the token in the db. We're able to put custom fields inside the token itself, it feels like we should take advantage of that to store a token group ID (which is stored in the token and db) instead.

This would also give us the advantage of being able to invalidate tokens a group at a time rather than only individually (see [this OAuth article](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#Refresh-Token-Automatic-Reuse-Detection)).

This isn't my area of expertise though - would be interested to get a thought from @passcod on this maybe?

passcod (changes requested):

edmofro (approved):
Codeowner stamp on the places this intersects with sync files. This is a beast PR man, amazing work 🤩 
Still requires re-review from @passcod or @mcccclean

mcccclean (changes requested):
Mammoth effort, unfortunately a few last change requests from me. Mainly from wanting to avoid committing clearly-unfinished stuff to the codebase. It's absolutely valid that sometimes we don't really know exactly what's going to go in a particular piece of functionality but I think "placeholder" and "todo" are always red flags.

(if a TODO really does need to be committed to the codebase, its comment should include a linear ticket code just so that its context not completely lost forever)

mcccclean (changes requested):
Great stuff! just one last gotcha and it'll be a :+1: from me

Edit: there's also some relevant test failures that are going on due to other tests initialising CentralServerConnection without a parameter. I think for our purposes here it's fine to just update CentralServerConnection's parameter to be optional rather than updating every test site as deviceId is only relevant in a handful of cases.

passcod (changes requested):
A few notes, but LGTM in general

mcccclean (changes requested):
Just reposting this note as I think it got lost in the mix - 

> There's also some relevant test failures that are going on due to other tests initialising CentralServerConnection without a parameter. I think for our purposes here it's fine to just update CentralServerConnection's parameter to be optional rather than updating every test site as deviceId is only relevant in a handful of cases.

passcod (approved):
verifying that my comments were addressed and standing by my last lgtm

mcccclean (approved):
LGTM :+1: - finally made it! nice work @dannash100

PR #3406, opened 2023-01-16
EPI-329: Cannot export from non general patient listing

NavarroEmilioLuis:
### Changes

Currently we are unable to use memoized components (React.memo()) on table accessors****. The problem here is that it returns a specific type of React component that is not callable as a function which produces an error. Data cells in table are usually pretty small so we can probably avoid using React.memo for the time being without too much trouble.

**** After having a closer look I noticed that:
- It's not entirely React.memo() usage, but only referencing it directly (i.e. `accessor: MemoizedComponent`).
- React.memo() can be used if the accessor is a function that returns said component (i.e. `accessor: props => <MemoizedComponent {...props} />`).
- Accessors cannot use components with hooks if the components are referenced directly, they need to be returned from a function instead.

This only covers the export button not doing anything, I couldn't replicate the styles getting messed up.


tcodling (approved):
Looks good to me!

PR #3411, opened 2023-01-16
WAITM-600 Add activity monitor, auto refresh token, auto logout inactive

EMcQ-BES:
### Changes

- Add `react-idle-timer` library to desktop app
- Update localisation to include `features.idleTimeout`
- Enable `/refresh` endpoint on lan
- Remove `TamanuApi` level auto-refresh (use the activity refresh instead)
- Detect user idle
  - Send `/refresh` request when user is active
  - Log user out if idle too long

### Checklist

- [X] Code is finished


tcaiger (approved):
This looks really clean and seems to work perfectly for what we need. I left a couple of comments which are just random thoughts and so no action needed.

PR #3417, opened 2023-01-17
EPI-299: Fix settings table

passcod:
### Changes

- Fixes the Setting update function to work without making lots of duplicates.
- Adds indices and constraints to enforce that better at the database level.
- **Changes** the behaviour of `Setting.set` to make more sense and allow keys to be deleted
	- given: `set('a.b', { a: 1, b: 2 })`
	- previously: `set('a.b', { a: 1, c: 3 })` would result in `{ a: 1, b: 2, c: 3 }`
	- now: `set('a.b', { a: 1, c: 3 })` results in `{ a: 1, c: 3 }` (`b` subkey is deleted)
- Adds two convenience SQL functions: 
    - `setting_get` which given a _leaf_ key will grab that setting (and do the boring stuff of checking `deleted_at` and co)
    - `setting_on` which given a _leaf_ key that's a boolean will return a SQL boolean.
    - leaf key limitation is due to it being a little complex to rebuild the nested object entirely in SQL... but could be lifted if needed / if someone has a bright idea how to do it.

So you can use settings in SQL! I made those for a next PR but I think they would be probably very useful for e.g. reports.

Also adds the `first()` and `last()` aggregates because I find those useful and more efficient than the workarounds but can remove if that's deemed out of scope.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM but maybe someone else should pick it up since it's so SQL intensive? Will leave a comment in the channel.

Also left some unimportant observations.

PR #3419, opened 2023-01-18
Github Actions: Fix tests-pass job

passcod:
### Changes

On advice from Github support, the dummy job needs to be set to always run, so it can then evaluate if it should fail the build. Otherwise a skipped or cancelled dependency, such as happens when the build is cancelled in favour of a new commit on the PR, would cause the tests-pass job to pass the build unwantedly.

Incidentally this construction also lets us be more flexible with the fail condition if we wish to: for example, we could have the storybook job run even though it's broken and specifically allow it through here.

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,7 @@ jobs:
 
   # Dummy job to have a stable name for PR requirements
   tests-pass:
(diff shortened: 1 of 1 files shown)


dannash100 (approved):
Loving github actions implementation so so much. blown away by how good it has become!

PR #3425, opened 2023-01-19
Fix down migration for PAD IDs

passcod:
### Changes

Found while debugging another issue. This fixes a down migration pretty far back.

### Checklist

- [x] Code is finished


EMcQ-BES (approved):
lgtm

PR #3450, opened 2023-01-26
NOT-1028: Upgrade to storybook 6, fix CI job and fix/update most tests to code changes

dannash100:
### Changes
Upgrades and fixes storybook without any other changes to babel between desktop and shared-src.
Quite alot of breaking changes in that major version to v6 see the restructure of all the config boilerplate

- Upgrade to v6
- Utilize webpack5 builder and manager
- Various fixes for the load of errors in CI step `storybook-verify`
- Utilize new addon for controls ( example in Button )
- Fix a tonne of stories as they are from days of connect Redux HOC and needed to be reworked due to patterns of datafetching and state management changes.

Screen Shot 2023-01-26 at 4 06 35 PM

With controls addon which i enabled for buttons as example
Screen Shot 2023-01-27 at 10 08 24 AM

### Checklist

- [x] Code is finished

_Upon merging:_


tcaiger (approved):
The changes look good to me and are contained to desktop storybook so should be safe I reckon

mcccclean (changes requested):
LGTM generally but it seems like the fix isn't quite enough to get the CI side of things over the line - change request to switch that back off before merging just to avoid misleading :x:'s on dev.

mcccclean (approved):

mcccclean (changes requested):
Worried that we're papering over a couple things that we don't actually want to ignore; just worth one more pass to see if these're properly fixable first.

mcccclean (approved):

PR #3456, opened 2023-01-29
EPI-270: Add ethnicity extension to patient data

NavarroEmilioLuis:
### Changes

In the end this wasn't so big of an update. I spent quite some time trying to get a unit test that covers this but "ran out of time" trying to mock config. It seems that everything that's imported from `shared` won't read any changes made to the config after the tests have started. Importing from `shared-src` would work but since this is inside a model, it has to be connected to the database and even then, the model will error with the custom types.

After talking with @mcccclean, he suggested we could either use an external function and just test the function based or we could pass the config as an argument. Since the latter involves touching a class method that's inherited, I think we probably want the first approach.

### DIFF
--- a/packages/shared-src/src/migrations/1674517247633-addExtensionToFhirPatients.js
+++ b/packages/shared-src/src/migrations/1674517247633-addExtensionToFhirPatients.js
@@ -1,0 +1,13 @@
+const TABLE = { schema: 'fhir', tableName: 'patients' };
+
+export async function up(query) {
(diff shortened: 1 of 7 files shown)


mcccclean (changes requested):
LGTM! Just one comment suggestion

nick-beyondessential (approved):
Codeowner viewed

mcccclean (approved):

PR #3458, opened 2023-01-29
No issue: get rid of iOS in mobile repo

NavarroEmilioLuis:
### Changes

Just cleaning up I guess. There are other things that are probably outdated but I'm just doing removing this bit that we already know it's just sitting there. I've also updated the slab docs that reference it.


tcaiger (approved):
I'm just picking this up from Daniel. Thanks for doing this. It makes no sense at all to have all this ios stuff cluttering up the repo. I'm in two minds about completely removing the references to "is IOS". On the one hand it removes confusion and on the other hand it is defensive programming which doesn't hurt so I think we just go with the quick win and merge this as is rather than combing the code for all the references to the is IOS and is ANDROID checks. Thanks again 🙏

PR #3468, opened 2023-02-01
WAITM-523 Don't use routed modals in encounter actions

EMcQ-BES:
### Changes

- Remove routed modal functionality from EncounterActions
- Allow passing in initial value for ChangeEncounterType modal/form

### Checklist

- [X] Code is finished
- [X] Testing notes added to the Linear issue

_Upon merging:_


tcaiger (changes requested):
This looks great - I'm so happy to finally be rid of the connected modals. I like the way that you've set up the ENCOUNTER_MODALS types for controlling the modal open state. There is just some additional clean up that we should do since this is the only remaining place that was using the connected modal. We also need to:
- delete the connectRoutedModal component from Modal.js
- remove support for `:modal` from usePatientNavigation from all endpoints, and
- remove the `:modal` param from all routes in PatientRoutes.js

tcaiger (approved):

PR #3478, opened 2023-02-03
Remove @injectConfig decorator and decorators from babel

passcod:
### Changes

Decorators cause us grief or spurious errors in some places (swc, vs code, possible storybook...) and don't provide enough value to keep.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Looks good!

edmofro (approved):
Codeowner stamp - I prefer it this way too as the decorators were unfamiliar, so this feels much less magic :-)

PR #3479, opened 2023-02-03
EPI-299: Fix field types on settings model

passcod:
### Changes

That was missed in an earlier pull request, and makes sync fail without it.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Great... I think I remember seeing those changes already though, did they get lost in a merge?

edmofro (approved):
Codeowner stamp

PR #3480, opened 2023-02-03
Sync: Skip calling the db when deleting nothing

passcod:
### Changes

If there's nothing to delete, we can skip a call to the database; it's very minor but saves a network roundtrip!

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Good stuff

edmofro (approved):
Codeowner stamp - thanks for this

PR #3481, opened 2023-02-03
Honeycomb: tracing

passcod:
### Changes

- Adds [OpenTelemetry](https://opentelemetry.io/) tracing support (via Honeycomb).
  - Leaves the logging honeycomb integration in place.
- Uses auto-instrumentation (enables HTTP client/server and Postgres tracing)
- ~~Adds manual instrumentation for sync and ScheduledTask, as a starting point~~
  - Splits manual instrumentation out of this PR, pending API improvements (like handling logging and tracing at the same time)
- Adds end-user context (from the facility server) and api-user context (from the central server)

See [my demo from the last Q&A](https://docs.google.com/presentation/d/1PyjR6gOeC3Hq7bCj0HbJUsqcVrvpOTrOf8cD8aiIbQ0/edit?usp=sharing).

Screenshot from 2023-02-03 14-28-00

#### Primer for span-based tracing / observability:

Basically, logs record points in time:

```js
    await this.sequelize.transaction(async () => {
      if (incomingChangesCount > 0) {
        log.debug(`FacilitySyncManager.runSync: Saving a total of ${incomingChangesCount} changes`);
        await saveIncomingChanges(this.sequelize, getModelsForDirection(this.models, SYNC_DIRECTIONS.PULL_FROM_CENTRAL), sessionId);
      }

      log.debug(`FacilitySyncManager.runSync: Setting the last successful sync pull time to ${pullTick}`);
      await this.models.LocalSystemFact.set('lastSuccessfulSyncPull', pullTick);
    });
    await this.centralServer.endSyncSession(sessionId);

    const elapsedTimeMs = Date.now() - startTime.getTime();
    log.info(`FacilitySyncManager.runSync: Finished sync run in ${elapsedTimeMs}ms`);

    // clear temp data stored for persist
    awaitdropSnapshotTable(this.sequelize, sessionId);
```

```
2023-02-03T22:50:50.146 FacilitySyncManager.runSync: Saving a total of 123 changes
2023-02-03T22:50:51.034 FacilitySyncManager.runSync: Setting the last successful sync pull time to 4567
2023-02-03T22:50:51.075 FacilitySyncManager.runSync: Finished sync run in 2808ms
```

To interpret logs you need to do math (read the difference between timestamps) and you need to understand the code. You have to mentally place the log entries in the idea of the code you have, and reason about the control flows as you reconstruct what happened, without getting confused by multiple threads potentially happening at once in a heavily asynchronous application.

Span-based tracing instead records:
- when a unit of work starts and ends
- useful data associated with that work
- what other work happens during that unit in the flow of the application

It calls these nested units of work "spans" of time or code, and associated data "attributes" of the spans. When a developer augments a piece of code with spans, they "instrument" that code. Spans can also contain point-in-time "events" to record important arbitrary places that don't map well to having their own span.

Behind the scenes, this "telemetry" is emitted ("exported") as structured data marking the start, end, associated data, and parentage of each span... then the observability backend (in our case, Honeycomb) processes all of it and presents it (e.g. as waterfalls) and/or allows it to be queried and/or slices it in various ways to obtain useful information (e.g. "what's the average time spent generating vaccine certificates, grouped by which nurse is logged in, in the last 24 hours").

_Distributed_ tracing improves on this concept by sending tracing information along when making calls between services, so that a called endpoint can hook up its spans to the span ID of the calling service. When both services send their traces to Honeycomb, it can reconstitute the full trace seamlessly:

Screenshot from 2023-02-03 14-26-04

#### Names of attributes

You may notice that attributes are named in a dotted format, with names that may seem unfamiliar or anti-thetical to our clarity principles:

```
net.host.name
service.environment
app.sync.sessionId
enduser.id
```

There are actually "[semantic conventions](https://opentelemetry.io/docs/concepts/semantic-conventions/)" for these fields:

- [`net.host.name`](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/) is the conventional name for the local hostname
- [`enduser.id`](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#general-identity-attributes) is the conventional name for the authenticated user ID
- [`service.`](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service) is the namespace for data related to the current application
- [`deployment`](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/deployment_environment/) is for the environment or tier
- `app` is an unofficial namespace for non-conventional data

#### Config changes

Under the `honeycomb` key:
- the `dataset` disappears, as that is now generated from environment/code information, so that each dataset contains the telemetry data for a single "service" — Honeycomb is happy to bridge spans across these, and indeed is how it expects tracing data to be given.
- a new optional `sampleRate` appears: this can be used to only send a fraction of traces to Honeycomb. E.g. for busy applications, sending 10% or 1% of the interactions can still provide a lot of information without bursting limits or adding too much overhead to a production deployment. For now this is set to `1` for 100% of traces sent.

#### Status

This integration is ready to use can be merged today.

- ~~Currently the full queries are recorded with database spans. This is very useful for debugging, but could contain PII in production environments. Can we do something about this?~~
  - **Yes**: in production, `INSERT`s and `UPDATE`s are redacted. In dev/testing, all queries are reported, and also the results are reported, for maximum usefulness.
- ~~Are we still getting the equivalent of the "log-based" information we care about? Should we send _both_ logs and traces?~~
  - Partially reverted the changes so we _do_ send _both_ logs and traces.

#### Future improvements (outside of this PR)

- Instrument more of the code.
- OpenTelemetry can also run in the browser. We could instrument the desktop app and complete our view of Tamanu.
- Instrument the meta server.

### Checklist

- [x] Code is finished

_Upon merging:_


mcccclean (changes requested):
This is really neat but I think it adds a lot of doubling-up in the code -- anyone writing or maintaining a task needs to go through both the logging API and the trace API for every event; I think that could create headaches for a junior or contractor new to the project for eg.

Probably worth pointing out that not every server will be hooked up to honeycomb (for eg Aspen, whose day-to-day maintenance etc is meant to be their responsibility; this will come up again in future deployments and I imagine the us/them splits will become more pronounced as we mature), so we can't treat traditional logging as legacy/deprecated - it's still core functionality that we have to continue to support.

So I think we'd be best off with some kind of wrapper object that is responsible for managing both. A task would initialise the object with a task name and have a single-line call to send events to it as they happen; from there it'd be the wrapper object's responsibility for sending those events to the console and to honeycomb. 

If it makes that process any smoother I think we're also fine to change up exactly _what_ gets logged to the console, for eg we can totally make the console logs a bit more mechanistic, `taskStarted.processPatientCommunications` rather than `"Started processing patient communications..."`.

mcccclean (approved):
LGTM.

Just checking is it going to mess things up if the default `honeycomb.dataset` (which is not overridden anywhere to my knowledge) changes away from `tamanu-general`? Won't that start sending things to a different place in honeycomb? (I guess that's fine I just want to be intentional about it)

anilvkaran (approved):

edmofro (approved):
Stamping the bit in `CentralSyncManager`

PR #3482, opened 2023-02-03
EPI-211: FHIR Job system

passcod:
### Changes

Jobs are submitted to a table, there's a whole mechanism for dealing with dropped jobs, the system can scale to multiple workers if we want to. Job submits can be done from JS, or, importantly for the FHIR work, trivially from SQL (e.g. in triggers).

Slab page with long form documentation: https://beyond-essential.slab.com/posts/worker-and-scheduled-jobs-pxxxxgb6

<details>
<summary>Pretty sequence diagram</summary>

image

</details>

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
I have a couple concerns about this PR but first up is an architectural question of how specific this Jobs system is designed to be - with its current name, it suggests it should ultimately form the backbone of the central server, replacing the ScheduledTasks system entirely, as just the way to do basically any piece of work outside of the things that happen immediately in a request handler.

If that's the case, it feels like it's got a worryingly large amount of config + functionality hidden inside SQL. I feel like a junior or someone new to the project should be able to add a new task to that ecosystem pretty easily, but I'm honestly not sure what that pattern looks like myself, and if something started misbehaving I'm not sure where I'd start looking. Just feels like a very high bus factor.

I think there's definitely a lot of room for improvement on the ScheduledTask system, and it's probably right to assume its true form is a jobs queue not unlike this, but I worry that how it's expressed in this PR isn't really appropriate to the shape of the team that'll be using it.

Otherwise, if they're meant to exist in parallel long-term - how does the a Job differ from a ScheduledTask? Like if I had a new feature next week where I need to turn a couple of Foos into Bars every x minutes, which system does that go in? Is this one meant for jobs that manifest as a hundred thousand very bitsy record-level tasks, where ScheduledTask is still the domain of half-hourly processing-heavy reports and cleanup jobs? Maybe it just needs a rename to clarify that?

mcccclean (changes requested):
LGTM; significantly less scared of the complexity of it when it's contained to the FHIR system. Just a couple comment-level changes and then good to go :+1:

mcccclean (approved):
:pray:

nick-beyondessential (approved):

PR #3483, opened 2023-02-03
EPI-343: Use encounter endDate for Medici report

passcod:
### Changes

- Do some formatting of the query to make it easier to read
- Use binds instead of replacements so postgres can better plan/optimise the query
- Add `dischargeDate` to the report
- Filter from encounter end date
- Exclude encounters which don't have an end date

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM I have a couple of questions but they're actually more business oriented? We can solve them later and ask in standup I guess

PR #3489, opened 2023-02-06
SAV-85: Add handling of error messages on sync session failure

NavarroEmilioLuis:
This should be part of 1.24.2 (I think alongside of other fixes).


edmofro (approved):
looks great!

edmofro (approved):

PR #3490, opened 2023-02-07
EPI-359: Upgrade @react-pdf/renderer to v3.1.3

passcod:
### Changes

This pulls in @react-pdf/yoga v4, which is fixed to no longer use asm.js, thereby eliminating the error. Also it reportedly runs at least twice as fast.

### Checklist

- [x] Code is finished

_Upon merging:_


dannash100 (approved):
Nice! When doing that experimentation for building for web I also had to update this package to get past an error

PR #3493, opened 2023-02-07
SAV-76: Only run imaging results data migration on central

stretchkennedy:
### Changes

Prevents the data migration (but not the schema migration) for the new `imaging_results` table from running on facility servers, to allow them to sync down the migrated data from the central server.

For a justification of why we want to do this, including why it involves importing config and why we're changing an existing migration, see the card.

### Checklist

- [x] Code is finished

_Upon merging:_

- [x] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed


passcod (approved):

edmofro (approved):
Title of the pr is flipped, but otherwise lgtm

PR #3495, opened 2023-02-07
EPI-351: Update Medici report API

NavarroEmilioLuis:
### Changes

The specs have changed and now we want to:
- Get rid of `period.start` and `period.end` params being mandatory, they should be optional.
- Set a default pagination limit to 100.
- Order by an encounter dispatch datetime (end date/discharge date).


stretchkennedy (changes requested):

edmofro (approved):
LGTM, pending CI checks :-)

PR #3453, opened 2023-01-27
WAITM-608: Lab request cancellations

tcaiger:
### Changes

- Add cancellation option and modal to lab request view (as long as it's not already published)
- Migrate lab_request_logs to be a string so that it can have new status types
- Make record read only if is cancelled
- Update styles to match design

### Checklist

- [x] Code is finished
- [ ] Tests are written
- [x] UI Screenshots added to the Linear issue
- [ ] Testing notes added to the Linear issue

_Upon merging:_

- [ ] Update the changelog (find it [here](https://github.com/beyondessential/tamanu/pulls?q=is%3Apr+is%3Aopen+release+in%3Atitle))
  - [ ] with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)
  - [ ] any config/settings changes and manual deployment steps
  - [ ] any db schema or other changes that could impact downstream data analysis
- [ ] Update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly) or [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q) as needed
- [ ] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed

PR #3452, opened 2023-01-27
WAITM-609: Imaging request cancellations

tcaiger:
### Changes

- add cancellation button and modal for imaging requests
- filter out deleted imaging requests in the imaging requests fetch endpoints
- update styles of imaging request view and table to match designs
- refactor config for imaging requests so that it's all in one place

#### Screenshots:

**Imaging Request Cancellation Modal**
Screen Shot 2023-01-27 at 3 16 09 PM

**Display Status in Table with Updated Status Tags**
Screen Shot 2023-01-27 at 3 15 57 PM

**Read Only View of a Cancelled Request with Attached Note**
Screen Shot 2023-01-27 at 3 15 39 PM

### Checklist

- [x] Code is finished
- [x] Tests are written
- [x] UI Screenshots added to the Linear issue
- [x] Testing notes added to the Linear issue

_Upon merging:_

- [ ] Update the changelog (find it [here](https://github.com/beyondessential/tamanu/pulls?q=is%3Apr+is%3Aopen+release+in%3Atitle))
  - [ ] with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)
  - [ ] any config/settings changes and manual deployment steps
  - [ ] any db schema or other changes that could impact downstream data analysis
- [ ] Update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly) or [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q) as needed
- [ ] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed

PR #3497, opened 2023-02-07
EPI-354: Update mapping for new Imaging Request statuses

passcod:
### Changes

Adds FHIR support for the statuses added in WAITM-608 (#3453) and WAITM-609 (#3452).

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (changes requested):
This looks good to me I just think you entered one in error? Sorry, couldn't resist. Also I think that it's a good choice for deleted to map to revoked, and also to return unknown status otherwise. Though I would say it might be good to keep the spreadsheet up to date given it's supposed to be the source of truth

NavarroEmilioLuis (approved):
Finally good to go

PR #3498, opened 2023-02-07
SAV-80: Improve patient search performance

NavarroEmilioLuis:
### Changes

Really not pleased with the results as the endpoint got even messier - still, creating a different endpoint would end in so much WET(ness) that it didn't made much sense either. For what I could investigate, these are the least amount of joins that we need for proper functioning of the "all patients" table.

I tried checking the performance of the query but couldn't get a clean baseline to begin with, it seems that it does improve it by a little, but I would've imagine it improving a lot more. Other patient listings do seem to work faster all the time, probably because of the filter in encounters so maybe that's the actual biggest bottleneck.

### DIFF
--- a/packages/lan/__tests__/apiv1/Patient.search.test.js
+++ b/packages/lan/__tests__/apiv1/Patient.search.test.js
@@ -381,9 +381,10 @@ describe('Patient search', () => {
       });
     });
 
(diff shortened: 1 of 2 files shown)


chris-bes (approved):
LGTM! Just waiting for the final call from Edwin

PR #3506, opened 2023-02-08
EPI-361: Use EXPLAIN instead of EXPLAIN ANALYZE for report import

stretchkennedy:
### Changes

`EXPLAIN ANALYZE` actually runs the query, which can take a while for big reports.

We don't use `PREPARE` because that leaves the query lying around until it's deallocated, and because the `-v` output might be useful to someone at some point.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


dannash100 (approved):
My bad on this one thank you @stretchkennedy

PR #3517, opened 2023-02-10
EPI-211: FHIR jobs

passcod:
### Changes

- Enables FHIR materialisation by default (as there are calls to make it a more stable target for reports etc...)
- Adds a scheduled task (and associated config) to put in jobs for creating any missing FHIR resources from the state of the main upstream table for each resource. That will provide both the initial population of FHIR resources into the system, and will fill any gaps that may occur for whatever reason (SQL triggers should be pretty solid in this regard but you never know).
- Adds a set of worker jobs (see #3482) for materialising FHIR resources:
  - `fhir.refresh.allFromUpstream` is issued from SQL triggers, and is responsible for "reverse mapping" the updated table row to any "FHIR upstreams", and submitting a `fhir.refresh.fromUpstream` job for the affected upstream/resource pairs.
  - `fhir.refresh.fromUpstream` performs the actual materialisation.
  - `fhir.refresh.resolver` is a post-processing step that needs to run after any materialisation does, but can cover any amount of refreshes, so it's more efficient as a separate job that runs less often than after every `fhir.refresh.fromUpstream`.
  - `fhir.refresh.entireResource` provides a way to request the refresh of an entire FHIR resource. This is useful for debugging and development, but in production its primary use is to request a full refresh of one or more resources as a post-deploy task, for example when a materialisation has changed or a field was added. It could also be useful for stress testing.
- Finally, adds a set of SQL triggers on various tables that all submit a job into the system described above whenever a row on one of select tables is updated, inserted, or deleted.

Okay so the way "reverse mapping" works is that we ask the FHIR Resource to give us a sequelize query object that, for any particular table and row in that table, would return all the rows of the "upstream" table that are affected by a change to that row. Then we take that query object and we ask sequelize to give us the SQL to execute that query. Sequelize has no publicly available way to do that, so of course we do the sane thing (cough cough) and tap into its internals to get us what we want. We then mash that SQL into an INSERT so that it runs directly on the database and inserts a whole bunch of jobs into the FHIR jobs table, and then the rest of the system is a lot more straightforward and just does basic materialisation stuff.

_Why_ do we do this horrible song and dance with SQL instead of getting IDs out of the database with normal sequelize queries and then reinserting them? Well, I think by asking this question it's rather obvious that my answer is "performance and efficiency" but why do I care about performance in this case? Well, because it turns out that a small change in the db could balloon to absolutely inane amounts of rematerialisations. For example, let's say we have 100,000 Pfizer vaccines that are materialised, and we change the name of the Pfizer vaccine in the reference data. You got it! That's 100,000 rematerialisations to go.

Sequelize has no good way to deal with inane amounts of data being shoved at you by Postgres. It doesn't have streaming or cursors\*. Sure, we can paginate, but that a) runs the risk of missing things in the gaps! (Like if more data is added to the first "page", shifting everything.) and b) just adds a ton of overhead.

So at that point I had three options:
- hook us up to the internals and get the damn SQL out and run it entirely in SQL
- go directly with the pg module to we can stream queries... except that I still need SQL to build and run that query!!!
- use something that can actually do all three of the streaming, SQL output, and not be a knob at complex queries, like `knex`... but then that would mean not being able to use the models!!!

Yeah, I went for the first option, as horrible as it seems, because the other ones are worse.

\* if you search you might find a Sequelize "cursor" library. That's not actually doing cursors: it's doing pagination. Cursors are actually a SQL thing that happens in the backend. You do a DECLARE query and then every FETCH query gets you a new page of information from the database, with a consistent view.

Streaming is another concept: basically postgres sends you each row of a result as a separate message in the protocol so you can handle one row at a time if you want, instead of buffering all the rows in memory and then only returning the results once they're all there, which is obviously a problem if you're requesting millions of rows.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
First go looks fine, still have to look closer to that SQL and the model files!

NavarroEmilioLuis (approved):
This looks good to me, but might not be a bad idea to get someone else to look at the more convoluted files. How solid is the test coverage for these?

nick-beyondessential (approved):

PR #3524, opened 2023-02-13
SAV-89: Patient was discharged automatically but shows current on desktop facility server

chris-bes:
### Changes
- Added hook to discharge outpatient encounters post persisting so that pull phase in the same session can also pull the updated encounters down

### Checklist
- [x] Code is finished

### DIFF
--- a/packages/shared-src/src/models/Encounter.js
+++ b/packages/shared-src/src/models/Encounter.js
@@ -12,6 +12,7 @@ import { dateTimeType } from './dateTimeTypes';
 
 import { Model } from './Model';
 import { onSaveMarkPatientForSync } from './onSaveMarkPatientForSync';
+import { dischargeOutpatientEncounters } from '../utils/dischargeOutpatientEncounters';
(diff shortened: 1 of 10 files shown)


edmofro (changes requested):
Amazing work, looks great! As always with backend work (and especially backend work that is in response to a bug), could use unit tests. To avoid blocking the hot fix could you spin them out into a separate card?

One request to change to using a cursor for the query batch pagination, will approve after that

NavarroEmilioLuis (commented):
Not officially reviewing but upon a glimpse I have two comments, both concerning the complexity this adds:

- Would the double batching be absolutely necessary?
- I wonder if we want to keep discharging on sync, it seems to me that it only affects whenever sync is not working (and then we have bigger issues). Discharging would naturally happen in a schedule anyway...

Also sorry for bringing this up just now, seems like it was quite some work.

edmofro (approved):

PR #3526, opened 2023-02-13
Redux refactor: replace all connect/connectApi with hooks

mcccclean:
### Changes

This takes the last components that're still using the old `DumbComponent + connect(etc)(DumbComponent)` pattern and brings them in line with the rest of the codebase.

### Checklist

- [x] Code is finished


dannash100 (changes requested):
Some stuff to check out and just a lint pass :)
Love the changes tho - cool to get rid of the last of these

dannash100 (approved):

PR #3527, opened 2023-02-13
Redux refactor: auth

mcccclean:
### Changes

We've had a `useAuth` hook around for a while now to access info about the currently logged-in user, but we had these parallel redux selectors as well. This switches things over to uniformly use `useAuth`, as well as updating `useAuth` itself to be an actual context so that we can replace its redux internals with state hooks in the future more easily.

### Checklist

- [x] Code is finished


dannash100 (approved):
Just needs lint but looks great - non auto-solvable lint issue on dev causing issues for us all i think

PR #3528, opened 2023-02-13
Redux refactor: patient info

mcccclean:
### Changes

This builds on the work to replace `connectApi` with the react hooks approach, following through on creating a usePatient context and utility functions to access + manipulate common patient info rather than manually going through a manual `dispatch(reloadPatient())` and `useSelector(state => state.patient)` every time.

### Checklist

- [x] Code is finished


dannash100 (changes requested):
I have went looking for usePatient so many times so glad this getting moved.
1 important breaking thing thrown up by linter.
Then just a lint pass

PR #3530, opened 2023-02-13
SAV-93: Stop multi-step transactions skipping sync

edmofro:
### Changes
- Take a shared lock from every create/update trigger that sets the updated_at_sync_tick to e.g. "6"
- Take an exclusive lock so from the facility sync manager so that it waits for all those transactions to complete

~Note that this is dependent on advisory locks being available across sessions, which @passcod is concerned is not the case, so will be looking into that shortly.~ Looks fine

### Checklist

- [x] Code is finished
- [x] Tests are written
- [ ] ~UI Screenshots added to the Linear issue~

_Upon merging:_


chris-bes (commented):
New commit looks great! I guess another point would be to add tests for Central similar to Facility. But I'm happy for it to be separate into another card since this one is urgent

stretchkennedy (approved):
Oh man, this is a nasty edge case. PR LGTM 👍

PR #3531, opened 2023-02-13
SAV-94: Check for timezone consistency on server startup

mcccclean:
### Changes

Adds a check on server startup to check:
- timezone according to node
- timezone according to postgres
- timezone according to config
- timezone according to remote central server (checked on facility server only)

If there's any mismatch it will throw and abort startup. I've included a config flag to make it just print a warning instead so that it doesn't cause any deployments/updates to go horribly wrong.

Postgres was actually not reporting a timezone at all when queried from the server process even though I could query it fine from psql. So... this PR also provides an explicit timezone parameter to the db connection, which I think guarantees that config + postgres will always agree. Possibly a little redundant, and it means that `config.countryTimeZone` is no longer an optional parameter. (that's probably a good thing though)

### Checklist

- [x] Code is finished

_Upon merging:_


passcod (approved):

passcod (approved):
Anyway this PR looks fine as is.

anilvkaran (approved):

PR #3534, opened 2023-02-13
Release v1.24.3

chris-bes:
### Changes
- SAV-89: Patient was discharged automatically but shows current on desktop facility server
- SAV-93: Stop multi-step transactions skipping sync
- SAV-117: Sync cleanup task should run in task process
- SAV-118: Connect to sync session while polling
- SAV-119: Sync QOL improvements
- SAV-127: Limit the number of concurrent snapshot processes for pull
- SAV-128: Snapshot records in chunks
- SAV-135: PAD cannot be upserted using importer


edmofro (approved):
Approving in advance in case I'm not available, but this should wait until it's tested on staging and my other fix is also in.

PR #3546, opened 2023-02-16
EPI-270: Include handling for non-existent PAD

NavarroEmilioLuis:
### Changes

An overlook from my part, it's highly unlikely a patient doesn't have a PAD but it's technically possible, this is to cover for that scenario


tcodling (approved):
LGTM :)

PR #3547, opened 2023-02-16
EPI-376: DB-defined reports should use ISO 9075 dates

stretchkennedy:
### Changes

There's a flag `filterDateRangeAsStrings` that toggles whether reports for the human-readable date "23/01/2023" get dates in the format `2023-01-22T13:00:00.000Z` or `2023-01-23` — no, that's not a typo, the first date has the wrong day. Because the first format can be on the wrong day depending on user timezone, and because there's no way to work out what the right day would have been, we're going to enable `filterDateRangeAsStrings` for all database-defined reports while there are relative few and we've still got control over them.

This is causing bugs in Fiji Aspen, and hotfixing it will allow us to fix them reliably. Our existing reports mean that a release step will have to be added to 1.24.3 to import a new report version without hacky workarounds.

### Checklist

- [x] Code is finished

_Upon merging:_

### DIFF
--- a/packages/lan/app/routes/apiv1/reports.js
+++ b/packages/lan/app/routes/apiv1/reports.js
@@ -46,6 +46,7 @@ reports.get(
         id: version.id,
         name: r.name,
         dataSourceOptions: version.queryOptions.dataSources,
+        filterDateRangeAsStrings: true,
         dateRangeLabel:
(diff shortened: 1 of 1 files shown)


dannash100 (approved):
Yes feel this should have been there all along. Thank you for this 🙏

PR #3550, opened 2023-02-17
EPI-355: Add visibility status to notes and patient death data

NavarroEmilioLuis:
### Changes

Avoid hard deleting records in facility server, instead setting a visibility status field to remove them.

There are a couple of unit tests that use `getNotePages` and `findX` methods, I'm not sure they need to be updated with the where clause as in [7c2156c](https://github.com/beyondessential/tamanu/pull/3550/commits/7c2156c8be4c7fe4aa26dd7cec2c472e1fa871da) because they would essentially be current notes anyway, plus that's not what is being tested.


tcodling (approved):
Looks good!

A bit picky but I noticed that with the models there are different imports for the Sequelize datatypes being used. However this seems like a more general inconsistency over this whole folder and could be be tweaked eventually since both imports do the same thing as far as I know + these were already imported so will leave it. 

Might not even be worth mentioning but I was just desperate to find something in one of my reviews 😆

PR #3564, opened 2023-02-20
EPI-270: Update FHIR NZ ethnicity extension mapping

NavarroEmilioLuis:
### Changes

It was noticed that one of the IDs was off, this is solving that problem. There is a broader discussion regarding having reference data IDs hardcoded but it's out of scope to bring this into a configurable value (by any means).


kujo4pm (approved):
Looks good to me. Listed in Aspen Reference Data as:`ethnicity-ITaukei`

PR #3565, opened 2023-02-21
SAV-65-Account for no location group

tcodling:
### Changes

Fix for encounter record where it breaks on encounters with no encounter group


chris-bes (approved):

chris-bes (approved):
lgtm!

PR #3574, opened 2023-02-22
SAV-65: Fix toasts for encounters without discharge

tcodling:
### Changes

Fix for the error toasts showing up on any old encounter without a discharge record + open encounters that don't have a discharge record yet.


tcaiger (approved):
Looks good to me :)

PR #3579, opened 2023-02-23
Update versionCompatibility for central server

mcccclean:
### Changes

VS Code hadn't saved this file oops


edmofro (approved):
Ah! I thought it seemed like one of these were missing, figured it was because we weren't bumping mobile but the gut was not wrong.

PR #3581, opened 2023-02-23
WAITM-479 Force PDF pages into single column

EMcQ-BES:
### Changes

- Change the PDFPreview to a flex column
- Add some spacing between pages

### Checklist

- [X] Code is finished


dannash100 (approved):
:shipit:

PR #3589, opened 2023-02-24
No-Issue: Fix simple get list count

tcaiger:
### Changes

There was an issue where the number of records being returned was less than the count being returned when fetching Lab Request records. It was caused by the difference in how the  sequelize findAll and count method works. The findAll method returns the labRequest records with the associated tests included as an array while the count method returns the raw sql query results which have has multiple rows for a single lab request if there are multiple lab tests associated with it. Fixed it by adding distinct to the count query.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (approved):
LGTM! Nice work!

PR #3592, opened 2023-02-24
SAV-65: Encounter record table tweaks

tcodling:
### Changes

Various small tweaks to Encounter Record based on testing results.

- [x] No space after “am” and “pm” as per figma design like "10:07am" not "10:07 am"
- [x] Don’t show discontinued medications
- [x] Filter out cancelled, deleted and entered in error lab requests/imaging requests
- [x] Shouldn't it be "Route of administration" as per testing notes?
- [x] Diagnosis should have column “Type” instead of “Certainty”
- [x] Based on what we need to order the rows in tables?
- [x] “Hospital admission” instead of “Admission”
- [x] Encounter Record” button should be “Encounter record” as per figma design
- [x] Filter out cancelled/entered-in-error lab/imaging requests
- [x] Filter out discontinued medications

### DIFF
--- a/packages/desktop/app/components/PatientPrinting/modals/EncounterRecordModal.js
+++ b/packages/desktop/app/components/PatientPrinting/modals/EncounterRecordModal.js
@@ -1,6 +1,8 @@
 import React from 'react';
 
 import { NOTE_TYPES } from 'shared/constants/notes';
+import { LAB_REQUEST_STATUSES } from 'shared/constants/labs';
+import { IMAGING_REQUEST_STATUS_TYPES } from 'shared/constants/statuses';
 
 import { EncounterRecord } from '../printouts/EncounterRecord';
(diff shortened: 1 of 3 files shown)


mcccclean (approved):
Approving as the line-by-line of this PR definitely represents an improvement.

But at an architectural level I think the current approach has a flaw that should be fixed next time we pick up work on it: the complexity of the data fetching and filtering going on here could be almost completely removed if we just built a dedicated endpoint for it on the facility server.

That way this modal could just do one single `encounter/:id/summary` request, and we could tailor that endpoint to fetch and return exactly the information we want. A db query with a couple joins is much more straightforward (both in terms of efficiency and code complexity) than a bunch of parallel web requests followed by filtering and processing. On top of that, it means we can attach unit tests to that endpoint, so we can stay on top of things breaking or changing.

PR #3598, opened 2023-02-26
WAITM-655 Add panel requests table

EMcQ-BES:
### Changes

- New migrations
  - Add panel request table
  - Add FK column to lab requests

### Checklist

- [X] Code is finished


dannash100 (approved):
💯 Thank you looks great

PR #3610, opened 2023-02-28
EPI-369: Implement new design for fhir tables

NavarroEmilioLuis:
### Changes

Ready for review, whole lot of changes but tests are passing! I'm not sure if `formatDateTime` should keep using `format` from `date-fns` (exports time with local computer time zone) or change to use `formatInTimeZone` from `date-fns-tz` (exports time with configured time zone).

Also probably needs to get rid of a bunch of unused stuff but didn't want to add that kind of clutter yet.

Lastly this needs a big warning on deployment that the migration(s?) might take a while. Maybe we want to do some test runs and time how long it takes.


passcod (commented):
Early LGTM (so far!) :)

passcod (changes requested):
This looks delightfully straightforward, which hides how much work I know went into this! A few suggestions but this is very good.

passcod (approved):

PR #3612, opened 2023-02-28
Disable autovacuum on temporary snapshot table

passcod:
### Changes

As we know we’re going to drop that table some minutes later, we can disable autovacuum on it, to avoid unnecessary background processing. See https://www.postgresql.org/docs/12/sql-createtable.html#RELOPTION-AUTOVACUUM-ENABLED

### Checklist

- [x] Code is finished

_Upon merging:_


edmofro (approved):
Great call!

PR #3568, opened 2023-02-22
SAV-128: Snapshot pull in chunks

chris-bes:
### Changes
- Snapshot pull in chunks

### Checklist
- [x] Code is finished

PR #3614, opened 2023-02-28
SAV-128: Revert chunking to check performance

edmofro:
### Changes

On the Aspen clone instance, we found that the snapshotting process took more than 50% longer with the chunk size set to 10,000 compared with 100,000,000 (1hr 40 vs. 1hr). Because the latter is essentially "off", I'd like to check whether it's even faster with the ordering etc. removed entirely. So, this reverts https://github.com/beyondessential/tamanu/pull/3568

That would leave us with just the concurrent snapshots config switch, but that may be enough. 

Will revert this if the performance doesn't significantly improve.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/sync-server/app/sync/snapshotOutgoingChanges.js
+++ b/packages/sync-server/app/sync/snapshotOutgoingChanges.js
@@ -15,13 +15,11 @@ const snapshotChangesForModel = async (
   sessionId,
   facilityId,
   sessionConfig,
-  config,
 ) => {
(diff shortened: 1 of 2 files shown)


chris-bes (approved):
I'm surprised that the build is green, I think there are also some tests modified for this new change so I'd imagine they have to be reverted too.

PR #3616, opened 2023-02-28
EPI-310: Migrate ImagingRequest.id to UUID

passcod:
### Changes

- Copy `id` to `display_id`
- Make new UUIDs for `id`
  - These are made deterministically using `uuid_generate_v5` which takes a "namespace" and a value and creates a new UUID from their SHA1 (unfortunately can't use other hashes, but that's good enough for this). The namespace needs to be a UUID, to create one we make another v5 from the nil UUID and the string `imaging_requests`, which works out to the constant value `5d0f0330-1bd7-59bd-a57e-b3297dd9d86c`.
  - Because the conversion is deterministic from `display_id`, the `id` value will be identical on central and facility
- Migrate all the foreign keys on other tables and the types of those foreign keys
  - imaging results
  - imaging areas
  - notes!
  - FHIR service request
    - which we also make a foreign key
    - this may need undoing at some point to multi-upstream that resource but for now it's good
- Also change some of the lan/ui code to use display ID as needed

### Checklist

- [x] Code is finished

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/ImagingRequestsTable.js
+++ b/packages/desktop/app/components/ImagingRequestsTable.js
@@ -35,7 +35,7 @@ export const ImagingRequestsTable = React.memo(({ encounterId, searchParameters 
   const imagingTypes = getLocalisation('imagingTypes') || {};
 
   const encounterColumns = [
-    { key: 'id', title: 'Request ID' },
+    { key: 'displayId', title: 'Request ID' },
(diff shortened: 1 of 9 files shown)


stretchkennedy (approved):
LGTM other than the one linting error for the TODO (which looks like it's already been todone)

PR #3617, opened 2023-02-28
SAV-128: Reinstate chunking

edmofro:
Reverts beyondessential/tamanu#3614, as the performance of snapshot time on Aspen Clone was exactly the same with it out (1 hour), as with it on but tuned to a value so high it has no effect. This also sets the default to a very high value.


chris-bes (approved):
LGTM!

PR #3620, opened 2023-02-28
TAN-2068: UX/UI improvements to "All patients" search fields

renatokoyama:
### Changes
https://linear.app/bes/issue/TAN-2068/uxui-improvements-to-all-patient-search-fields-and-table
- Changed styles for all patient search to match designs here:
https://www.figma.com/file/sy6gyLBPoSXuJNq5lEEOL8/Tamanu---UI-Pattern-(Updated-2022)?node-id=6858%3A92295&t=TeRlwDXnZ2thTcrA-0
- Added a new field to filter by gender
- Created advanced filters button

After:
image
image

https://user-images.githubusercontent.com/17033058/221986835-14ea18c0-4659-434a-a225-043f37d8ed63.mov

Before:
image

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Couple changes I'd like to make to this one - plus a general note that for consistency it's preferable to define the styled components at the top of a file and then use those, rather than providing `style={}` props inline.

mcccclean (changes requested):
I'm afraid I have to push back a bit harder on the use of `style=` props - in pretty much every single case we should be going with a component styled by `styled-components`. It avoids the css becoming intermingled with functionality and data, and helps keep the codebase consistent.

Regarding flexibility - I'm not actually sure that being able to inject arbitrary CSS is really a benefit! The app is built on a consistent pattern of designs, so I think developers _should_ experience some friction when it comes to introducing small variations between styling options. In cases where the design calls for minor variations in a component, we should do that via adding descriptive props to the component that let it alter its own style rather than injecting CSS directly.

mcccclean (changes requested):
One last change request :pray: - I think this note got lost in amongst other stuff earlier, sorry!

We should be favouring using `styled-components` rather than the `style=` prop wherever possible. It keeps things consistent with the rest of the codebase, and helps keep styling stuff distinct from data & functionality (as well as being very slightly more performant).

There're a few `style=` parameters getting passed in this PR - so, my last note is to replace those with the `styled(x)` pattern.

mcccclean (approved):
:tada:

PR #3643, opened 2023-03-02
Revert "EPI-310: Migrate ImagingRequest.id to UUID (#3616)"

passcod:
This reverts EPI-310


edmofro (commented):
This is a bigger change to revert than I thought, and I see why you did it and totally agree. I wonder if instead of fully reverting it, we could keep the guts (moving the non UUID ids across to displayId, and generating UUID style strings for the id field) but just change the data type of `id` back to `TEXT`? I.e. rather than reverting this PR add a new migration to change the data type

PR #3644, opened 2023-03-02
SAV-1: Age with months formatting tweaks

tcodling:
### Changes

Some small formatting alterations to the new "Age with months" field for surveys. This will only show months if the patient is less than a year old and will also not show a plural if there is only one month or year


chris-bes (changes requested):
LGTM! I'm just being OCD for those 2 minor suggestions 😅

chris-bes (approved):

PR #3645, opened 2023-03-02
Revert EPI-310: back to varchar only

passcod:
Alternative to #3643


edmofro (commented):
I like this one. Is there anything else we need to change, like that upstream uuid thing I spotted in the other revert?

edmofro (approved):

PR #3652, opened 2023-03-05
SAV-106: New general vaccination certificate

tcodling:
### Changes

This card is for the new general vaccination certificate. This uses the same button as before but now generates a certificate with no QR code that lists all vaccinations. The removal of the QR code also allows the certificate to have a new style (https://www.figma.com/file/sy6gyLBPoSXuJNq5lEEOL8/Tamanu---UI-Pattern-(Updated-2022)?node-id=7223%3A114922&t=VjZw0iuC8VmorcBG-1).

This was an interesting task as due to the fact this needs to be emailed, we generated it on the server like before. This meant I couldn't use any of the existing components for the desktop client and needed to do all the styling via react-pdf components with inline styling objects. **I am very open to suggestions if the reviewer thinks there might be a better approach to styling this as I know it isn't the slickest :)**

This is essentially the code from the previous COVID-19 certificate copied with all of the QR code related stuff stripped out and redesigned according to Figma. Due to the complexity of the COVID-19 QR code and the fact we can't change any of the style of the certificate we decided it's best not to interfere and renamed all of the previously used components specific to the QR code for the old certificate to have "COVID" at the start. I then used the original generic names for this certificate with the idea we could hopefully reuse some of the components if we need to generate more certificates on the server in the future.

### Checklist

- [x] Code is finished

_Upon merging:_


chris-bes (changes requested):
Code looks really great! I just have 2 tiny suggestions for you

mcccclean (changes requested):
Couple changes of my own to request (pardon the double-review, forgot to :caught: it!)

mcccclean (approved):

chris-bes (approved):
LGTM!

PR #3655, opened 2023-03-06
Add Scheduled task as a tracing entrypoint

passcod:
### Changes

Re-adds the manual instrumentation of ScheduledTask, this makes each run of a scheduled task a new trace, completing the Tamanu picture. Notably this makes it possible to get a complete trace for a sync session.

### Checklist

- [x] Code is finished

### DIFF
--- a/packages/shared-src/src/tasks/ScheduledTask.js
+++ b/packages/shared-src/src/tasks/ScheduledTask.js
@@ -39,7 +39,7 @@ export class ScheduledTask {
       if (!outcome) {
         // We expect the subtask will have caught & logged all its own errors
         this.log.info(`ScheduledTask: ${name}: Not running (subtask failed)`);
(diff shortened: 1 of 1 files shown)


mcccclean (changes requested):
Nice work, will be great to get proper tracing on the tasks :pray:

mcccclean (approved):
LGTM!

Re consolidating the logging items, I'm generally pretty wary of "I'll get this in now and do the clean version later", as code cleanup tasks tend to take a lower priority than other tasks -- if the cleanup is imminent, then we may as well wait for it; if it's not then we need to be ok with the PR in its current state.

That said, yeah, it's all contained to one file, so I _am_ ok with the PR in its current state! Still would prefer the event stuff to be consolidated, but I'm not going to block approval over it.

PR #3658, opened 2023-03-06
Extract @tamanu/tracing package

passcod:
### Changes

- Extracts the tracing integration from shared into its own package.
- Adds a test for serviceName()
- Replaces all the various opentelemetry dependencies with a single `@tamanu/tracing` dependency.

### Checklist

- [x] Code is finished
- [x] Tests are written


mcccclean (changes requested):
Seems like the linter has caught a few instances of areas that need updating to use the new package? Otherwise LGTM

PR #3660, opened 2023-03-06
Run CI without a pre-build step

passcod:
### Changes

Now that we've got lots of runs let's see if _not_ sharing the shared build saves us time or not.

Answer: it does. Previously:
- Build + pack + upload = 5 minutes
- Download into each step = 30 seconds
- Total 5 minutes 30 seconds (plus worker scheduling time)

With this PR:
- Yarn = 2-3 minutes
- Build shared = 30-50s
- No upload nor download time
- Total 2-4 minutes

Not only does this mean we save a minute and a half on each whole run, but we also get partial results even faster (everything save for lan+sync finishes inside of 5 minutes, whereas before they were just starting after 5 minutes).

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
WET for speed, nice catch!

PR #3666, opened 2023-03-06
Add private:true to packages that miss it

passcod:
### Changes

What it says on the can.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
🎉

PR #3667, opened 2023-03-06
NOT-1032: Configure Terser to leave classnames intact

passcod:
### Changes

Configure Terser (webpack's minimiser) to leave classnames and function names alone.

### Checklist

- [x] Code is finished


mcccclean (approved):
Nice! LGTM

PR #3668, opened 2023-03-06
Enable babel-loader caching

passcod:
### Changes

This makes (local) builds of sync-server source up to 300% faster.

### Checklist

- [x] Code is finished


mcccclean (approved):
:tada: hell yeah

PR #3265, opened 2022-12-01
NOT-988: Use swc in place of babel where possible (excludes mobile)

dannash100:
### Changes
A piece of work for the task NOT-988

Use `@swc/cli`, `@swc/core`, `@swc/jest`, `@swc/register`, `swc-loader` in place of babel to greatly improve speed of compilation see
https://swc.rs/

Benchmark on compiling shared
before:
```
Successfully compiled 530 files with Babel (3329ms)
Successfully compiled 530 files with Babel (4174ms).
Successfully compiled 530 files with Babel (3068ms).
```
after:
```
Successfully compiled: 530 files with swc (149.82ms)
Successfully compiled: 530 files with swc (185.2ms)
Successfully compiled: 530 files with swc (131.03ms)
```

Places where it wasn't appropriate to replace right now
- Mobile (react native)
- ESlint parser
- Storybook

In swapping out `@babel/jest` due to the pattern of `jest.spyOn` on an `import * as module from 'module'` breaking ES specification

Stage III decorators with metadata (i.e context.kind) I couldn't get to work after sinking hours into it. (see injectConfig on SyncManagers

Had to clean up some module import export syntax in few places - edge cases like import * as from json file

All other packages have all references to babel are ported - this also reduces a bit of boiler plate in some places

### Checklist

- [x] Code
- [n/a] Tests
- [n/a] UI Screenshots
- [n/a] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- [n/a] Include any config changes in the release PR changelog
- [ ] Update the [release PR](https://github.com/beyondessential/tamanu/pulls?q=is%3Apr+is%3Aopen+release+in%3Atitle) changelog with a ticket reference (e.g. TAN-001 or WAITM-001 will be automatically linked)
- [ ] Add testing notes to the Linear issue

PR #3669, opened 2023-03-07
Build shared with swc

passcod:
### Changes

Partially pulled from #3265.

- This is an in-place replacement for babel. It doesn't change anything about how we use shared, just improves its own build.
- Uses [swc](https://swc.rs/) instead of babel
- Changes slightly the way shared's package.json is handled: it's the original, copied to src/ just before compilation with the `name` changed. That behaves somewhat better with dependencies/dependents.
- Leave class names unmangled ([NOT-1032](https://linear.app/bes/issue/NOT-1032/disable-class-name-mangling))
- Moves the types (mostly there to help vscode, not compilation) so swc doesn't complain 

Before:

```
Successfully compiled 598 files with Babel (2779ms).
```

After:

```
Successfully compiled: 598 files, copied 1 file with swc (247.37ms)
```

### Checklist

- [x] Code is finished


mcccclean (approved):
Yesss this package clean up work has been such a long time coming :pray: :pray: :pray:

dannash100 (approved):
Amazing work - great improvements and cleanups from my source pr. 💯

PR #3672, opened 2023-03-07
Revert "Build shared with swc"

passcod:
Reverts beyondessential/tamanu#3669

This may have broken dev.


edmofro (approved):
Stamping as a straight revert, though I haven't checked if it broke dev

PR #3680, opened 2023-03-08
Script test-all and lint-all

passcod:
### Changes

This makes the `yarn test-all` and `yarn lint-all` top-level scripts read from the workspaces package list and select all packages that have a test/lint script so that if there's new packages it doesn't need changing.

In CI the list is hardcoded because it spins up different/parallel runners but locally those scripts are/can be used. It also makes those scripts a lot less scary!

### Checklist

- [x] Code is finished


mcccclean (changes requested):
Functionality LGTM but the change request is to split out the repeated code as in https://github.com/beyondessential/tamanu/pull/3679#pullrequestreview-1331674407

mcccclean (approved):

PR #3681, opened 2023-03-08
Add common eslint config for ts packages

passcod:
### Changes

This moves the typescript eslint config used in CSCA to the top level and merely references it from the package. That lets future typescript packages share the config easily instead of having their own copy.

It also reviews some of the settings to align with recommendations.

### Checklist

- [x] Code is finished


dannash100 (approved):
Ahh this is so great as the eslint config for mobile has felt all over the place, I had been trying to fix some small things here and there that annoyed me.

dannash100 (approved):
Love it! just some random questions and things

PR #3683, opened 2023-03-08
Build shared with swc, take II

passcod:
### Changes

This one is verified to work in dev! It replicates the old shared/package.json and only substitutes swc for babel.

Previous PR: #3669, revert #3672.

### Checklist

- [x] Code is finished


dannash100 (approved):
Working!! thats amazing, looks great. Mabye we shud do a post after merge re `import * as` situ incase it trips someone up

PR #3686, opened 2023-03-08
Fix yarn lockfile being different before and after shared build

passcod:
### Changes

This was a temporary change I'd made while trying out things that I forgot to remove before merging #3667.

Because shared is in the dependency list, before it's built yarn pulls it from registry, and after it's built, yarn pulls it from local, and as that's different packages the lockfile doesn't work anymore.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
🎉

PR #3689, opened 2023-03-08
Add common tslint config

passcod:
### Changes

As continuation of #3681

### Checklist

- [x] Code is finished


mcccclean (approved):
:tada: love this! so glad that someone's finally picking up all these build/workspace/etc cleanup tasks

PR #3699, opened 2023-03-09
WAITM-601: Mobile logout from inactivity text spacing tweak

dannash100:
### Changes
- Fix spacing on logout from inactivity so it doesn't push other elements down

Screenshot_20230310_091521_Tamanu Mobile

### Checklist

- [x] Code is finished
- [x] Tests are written
- [n/a] Testing notes added to the Linear issue

_Upon merging:_


tcodling (approved):
sweet pr as per usual

mcccclean (changes requested):
Whew, beast of a PR - I do have some notes though, hopefully this doesn't result in too much of a rewrite...!!!

mcccclean (changes requested):
Oh, wait, actually, jokes aside... I realised I DO have a note about the approach in this PR! It kind of kicks the can down the road a little bit, where if we change the text (for eg via translation) or use the display to show a different error message, this bug will return.

Rather than have the element appear/disappear and have the other elements adjust their margins to account for it, I think it'd be more of a robust solution to display the error message inside a fixed-height container which is always there. That way we don't need to do any calculation of individual heights of elements, and the "eHealth patient record" section doesn't need to be aware of whether we're showing an error message above it.

mcccclean (approved):

PR #3702, opened 2023-03-09
EPI-369: butcher the codebase

NavarroEmilioLuis:
### Changes

This is just separating the cleanup work from the actual card work, to give a cleaner review diff. Tests were failing on first commits because of the linter, fear not 😆

On restoring `sequelizeType`: kinda needs to be there to match the other composite type utils, though were not currently using them, the bridge is not really straightforward. I'm not sure if I should also nest all fields inside of a `params` variable like before instead of top level. [See this commit where it was changed](https://github.com/beyondessential/tamanu/pull/3610/commits/ad7260bc86152a5935adcd59a576e9110952c368).

Lastly, though pgType and sequelizeType aren't being used, the parser (and stringifier) are probably still being used?


passcod (changes requested):
Only a couple notes. I'm of two minds to leaving the composite stuff in now... but with this PR decoupling it completely from FHIR I think if we want it out we can easily do it in a later PR.

passcod (approved):
Your tests aren't passing, but LGTM otherwise

PR #3704, opened 2023-03-09
Merge csca and mobile eslint configs into one common config

passcod:
### Changes

As continuation of #3689 and #3681

### Checklist

- [x] Code is finished


mcccclean (approved):
LGTM! It seems to me like there're a few items that could be further consolidated, but it's still beneficial as-is (and I'm happy to assume there's reasons to not consolidate those)

PR #3726, opened 2023-03-14
WAITM-670: Fix mobile dropdown overlay on surveys

tcaiger:
### Changes

There is an issue where the menu in dropdowns on mobile surveys is hidden behind the next dropdown if there are 2 in a row. This is actually an issue that is existing everywhere as far as I can tell but has never come up because we don't have any examples of multiple dropdowns in a form anywhere outside of surveys.

I spent more than a couple of hours trying to find a good solution for this but manually setting the z-index in the survey form is the best I could come up with. The library that we use for the dropdowns is quite basic and doesn't support portals which would probably be the proper solution to this issue. Changing out the library for a more established one could be a long term solution but I didn't consider it in scope for this ticket.

Screenshot of fix
Screen Shot 2023-03-14 at 3 34 04 PM

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/mobile/App/ui/components/Forms/SurveyForm/FormFields.tsx
+++ b/packages/mobile/App/ui/components/Forms/SurveyForm/FormFields.tsx
@@ -57,14 +57,12 @@ interface FormFieldsProps {
   note: string;
 }
 
-export const FormFields = ({
-  components,
-  note,
-  patient,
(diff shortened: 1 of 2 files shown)


EMcQ-BES (approved):
LGTM

PR #3728, opened 2023-03-14
WAITM-479 Always scale PDF pages by width

EMcQ-BES:
### Changes

- Don't try to be clever, just always scale the pages to fill the width of the modal


tcaiger (approved):
Oh yeah that works perfectly :)

PR #3736, opened 2023-03-15
Fix settings tests (and code)

passcod:
### Changes

- Add FHIR settings to tests (what broke the tests)
- Show which settings are facility-only and which are global in listings

### Checklist

- [x] Code is finished
- [x] Tests are written


NavarroEmilioLuis (approved):
LGTM

PR #3738, opened 2023-03-15
EPI-310: Imaging ID regression fixes

passcod:
### Changes

- Two missing ID field changes in a view and a report
- A missing change in the request creation form meant that short IDs were getting written to the `id` field and a random UUID was getting generated for the `display_id` field database-side.
- A manual database query was applied to staging to correct the records created wrongly from that last missed change: `update imaging_requests set display_id = id, id = display_id where length(display_id) = 36;`

### Checklist

- [x] Code is finished


dannash100 (approved):
Looks good!

PR #3741, opened 2023-03-15
Tests: Create facility from config if not already seeded

passcod:
### Changes

Something frustrating to debug: when running tests for `lan` locally, you needed to make sure that your `serverFacilityId` line in `lan`'s `local.json` was commented out, otherwise you'd get foreign key errors on the `patient_facilities` table on every test.

This creates the facility if it wasn't already seeded, dropping the requirement for the above.

### Checklist

- [x] Code is finished


tcaiger (approved):
Thank you thank you

PR #3743, opened 2023-03-15
EPI-350: Discharge records are not being created

NavarroEmilioLuis:
### Changes

Main change is to refactor the way we create `Discharge` records: currently it's just one single place and it happens inside `onDischarge` method of an encounter, which ends up being a side effect of updating said encounter.

After a brief chat with McLean, we concluded that the `Discharge` record is supposed to be just extra information about a discharge and an encounter's `endDate` is what tells us if its actually discharged. Still, including a note (aside from a system note) is enough "additional information" to have.


dannash100 (approved):
Looks great 💯 amazing work

PR #3751, opened 2023-03-17
No issue: staging fix - solve encounter view desktop error

NavarroEmilioLuis:
### Changes

This is probably caused by the procedurally generated data: lab requests are expected to have a `requested_by_id`, I think that we currently only create lab requests via the lab request form, and that is a mandatory field. However, at database level we do not enforce this, so I think that it's OK to not expect it all the time.

Also no need for card or release notes given that this is a new component created for this release.


dannash100 (approved):
Yes yes :D Thank u heaps

PR #3756, opened 2023-03-19
SAV-65: Small label change

tcodling:
### Changes

Tiny label change for table column on encounter record summary


NavarroEmilioLuis (approved):
Thanks for the changes, it's a bit nitpicking given that it works but we'll benefit from the tidy up in the long run. I have a last one of such requests but this one is just a click away, sorry for the hassle.

PR #3764, opened 2023-03-20
Hotfix v1.24.5

edmofro:
### Changes

In Nauru, found that db queries were timing out despite not much going on CPU or memory wise. Looking at pg_stat_activity, found a bunch of advisory locks idle in transaction. We use advisory locks to represent when a sync session is actively snapshotting. The corresponding sync sessions weren't actively being connected to though, and there were no active db queries that looked like they were doing snapshotting, so there should be no need to hold an advisory lock. 

Looked in the code and found that there are a few lines sitting outside of the `try/finally`, and if those throw an error, the finally that releases the advisory lock would never be hit. This means any e.g. timeout errors that result from actual resource max out will compound, as they'll leave around an idle transaction that is not doing any work, but is hogging a db connection.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tamanu",
-  "version": "1.24.3",
+  "version": "1.24.5",
   "description": "This repo contains all the packages for Tamanu",
   "main": "index.js",
   "homepage": "https://github.com/beyondessential/tamanu.git#readme",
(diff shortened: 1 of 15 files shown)


mcccclean (changes requested):
One tiny change...!

PR #3786, opened 2023-03-22
EPI-310: Fix imaging request searching

passcod:
### Changes

- Do case-insensitive startsWith queries the right way, with `ILIKE`. This lets sequelize correctly write columns and avoids ambiguities (which was the original bug getting fixed here).

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Makes perfect sense. If I recall correctly we still had to support SQLite back then.

PR #3799, opened 2023-03-23
SAV-167: Can only use encounter actions on current facility

tcodling:
### Changes

Currently you can edit the details on an encounter from any facility even if not the in facility where the encounter was created. This can cause issues as the lists for moving location, moving department, changing clinician, etc are filtered by the current facility which could get confusing and cause unwanted behaviour.

Here are a couple examples where it can go wrong:
- A patient triaged in Facility A has their location updated by a user connected to Facility B. This swaps their encounter to be under Facility B and removes them from the Facility A emergency dashboard. Admitting that patient to hospital from Facility A moves the patient back on the Facility A inpatient list but their encounter stills displays the facility as Facility B.
- A patient admitted to Facility A has their department updated by a user connected to Facility B server. Their encounter facility is still listed as Facility A however they are now only viewable on Facility B inpatient listing.

### Checklist

- [x] Code is finished

_Upon merging:_


tcaiger (approved):
Nice one! I tried to break it and couldn't so it must be perfect

PR #3809, opened 2023-03-24
No Issue: Fix flakey Patient.search test

Alastair-L:
### Changes
image

- The issue was that the autogenerated patient first name was picked up by the search functionality by default.


mcccclean (changes requested):
I see how this solves the immediate issue, but I worry that it's a bit of a blunt-instrument way to go about it. 

That is, if it's bugged behaviour to omit a lastName field on a test data item, then we should fix that bug rather than require devs to add a lastName field to everything even when it isn't relevant.

I've got two alternative approaches that I'd propose:
1. Update the code that loops over `searchTestPatients` so that it handles filling that field out, defaulting everyone's `lastName` to `'ignored-name'`, and from there it's the responsibility of the records that want to use a random name to generate that manually in the `searchTestPatients` array
2. Fix the name sort test so that rather than expecting a particular name to be result 1, it just expects its index in the result array to be earlier than another comparison record. (this one takes a bit more thought but it'd be my preferred approach - generally I think it's better when a unit test is written to cleanly accommodate a bunch of other random data in the db rather than requiring things to be set up just-so. Certainly not always possible or convenient but I reckon it's doable here).

PR #3813, opened 2023-03-26
Bump mobile to 1.24.58

mcccclean:
### Changes

No functionality changes, just clearing up an ambiguity where another v1.24.57 already exists elsewhere/undeployed in a hotfix branch so bumping this one to disambiguate.


edmofro (changes requested):
It won't block, but will log warnings if we don't also update `versionCompatibility.js` in the sync server 🙏

edmofro (approved):

PR #3820, opened 2023-03-27
Fix patient filter+sort flaky test

mcccclean:
### Changes

Thanks to @Alastair-L for identifying the root cause & first pass :pray: - https://github.com/beyondessential/tamanu/pull/3809

Fixes an intermittent test failure on dev. The randomly generated surnames of other patients were occasionally polluting the data for the tests that expect some surnames in a particular order.

I tried addressing my own notes on Alastair's PR but they ended up quite complex! There's a note in the source about benefiting from keeping the list of test patients intermingled, but I think preserving that for the specific behaviour that this test suite isn't really worth it. 

So this represents a different tack, just separating out the filter+sort tests to run on entirely separate patient lists.

### Checklist

- [x] Code is finished
- [x] Tests are written


stretchkennedy (approved):
LGTM 👍

PR #3829, opened 2023-03-28
SAV-108: Hide button rather than disable it

tcodling:
### Changes

I initially designed the new covid-19 vaccine certificate button to be greyed out when no certifiable vaccines have been administered on the patient. We actually want it to not even show if there are no certifiable vaccines.


tcaiger (approved):
amazing

PR #3836, opened 2023-03-28
EPI-234: Extra lab request statuses for DiagnosticReport

passcod:
### Changes

Not technically what the ticket is for but the ticket is for something which doesn't even exist yet (materialisation of Lab Requests into ServiceRequests). This is merely related.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
LGTM

PR #3839, opened 2023-03-29
Staging fix SAV-185: Immunisation register breaks when filtering by Area or Department

chris-bes:
### Changes
- Include sql join for departments and location groups when there are filters

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/views/patients/PatientListingView.js
+++ b/packages/desktop/app/views/patients/PatientListingView.js
@@ -143,7 +143,7 @@ export const PatientListingView = ({ onViewPatient }) => {
         <PatientTable
           onViewPatient={onViewPatient}
           fetchOptions={{ matchSecondaryIds: true }}
-          searchParameters={searchParameters}
+          searchParameters={{ isAllPatientsListing: true, ...searchParameters }}
           columns={LISTING_COLUMNS}
         />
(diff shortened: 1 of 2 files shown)


NavarroEmilioLuis (changes requested):
Okay this is something that got overlooked somehow🤔

I agree in what you're saying: let's leave the refactor for later. Still, I think that we should continue to work in this! Aspen probably doesn't use it so we got lucky - this table should've been tested too.

I think that a proper fix would be to change the heuristic used to tell if it's the all patients listing: instead of just relying on sending `facilityId` let's actually send a param from the front end that says it is, something like:

```
// Check if this is the main patient listing and change FROM and SELECT
// clauses to improve query speed by removing unused joins
const isAllPatientsListing = filterParams.isAllPatientsListing;
```
And that should allow us to get the other join + cleaner? I believe there is probably a more important lesson here (that we can save for the refactor) which is that for such a convoluted endpoint, let's not depend on heuristic and send a param that identifies each different table, so it knows how to handle special cases.

Thanks for getting this Chris, sorry for the hassle!

NavarroEmilioLuis (approved):
Thanks!!

PR #3866, opened 2023-04-04
Fix settings subcommand

passcod:
### Changes

Really silly mistake.

### Checklist

- [x] Code is finished


stretchkennedy (approved):

mcccclean (approved):
The mistake highlights a bit of a DRY violation, could do something like...

```javascript
const logWithSeparator = async val => console.log(`-------------------------\n${await val}`;

// then
  .action((...args) => logWithSeparator(listSettings(...args))
```

but I don't feel strongly about this, happy as-is.

PR #3868, opened 2023-04-04
Set postgres application_name

passcod:
### Changes

The `application_name` of a connection shows up in postgres stats and activity views. For our deployment pattern of a database per application, most connections will be from tamanu, but this may be useful to e.g. differentiate from backups, psql, and any other connections.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
LGTM 👍 

I wonder if we should have a different application per subcommand? That's probably unnecessary work though.

PR #3870, opened 2023-04-04
NASS-717: Dynamic selection field

tcodling:
### Changes

We are introducing a new type of select field to the app that dynamically changes based on the number of available options to select. This will take either an array of objects as options or a suggester. If the options array is <= 7 we will display the input as a select field and if above we will display as an autocomplete field. Same goes for if a suggester is passed, we query it and if the first page of results is less then 7 we will display those as the select options otherwise we will display as an autocomplete field like usual with suggester logic.

### Checklist

- [x] Code is finished

_Upon merging:_


dannash100 (changes requested):
Some questions, but good job getting this done so quick! also story looks great

dannash100 (approved):
Beautiful!

PR #3876, opened 2023-04-04
EPI-164: Expose lab requests as service requests

NavarroEmilioLuis:
### Changes

Core work was changing the FHIR models API to be able to work with more than one upstream model. After that was done, I then started doing a bit of a refactor to the files since they were getting long - since this is not the main goal, I'm concentrating on moving on and create the new mapping while only refactoring the service request files.

After that's done and reviewed (to get some feedback on the refactor itself) I'll create a new PR that refactors every other model to avoid some clutter here.

Changes in config:

- Rename `serviceRequestId` for `serviceRequestImagingId`
- Rename `serviceRequestId` for `serviceRequestImagingDisplayId`
- New entry `serviceRequestLabId`
- New entry `serviceRequestLabDisplayId`


passcod (approved):
Mammoth effort! This all looks great. Left a bunch of notes, none needing addressing in _this_ PR, but for the next refactoring one.

nick-beyondessential (approved):

PR #3880, opened 2023-04-05
SAV-167: Show discharge summary on all facilities fix

tcodling:
### Changes

This card hides the encounter actions for encounters not on the current facility. This also had the side effect of hiding the discharge summary from other facilities too which should still be present. This fix adds a check that will show the encounter actions on any facility as long as it is complete


dannash100 (approved):
Yus!

PR #3889, opened 2023-04-10
[NASS-708] Create endpoint for exporting Reference data

renatokoyama:
### Changes
- Added new endpoint for exporting reference data

https://user-images.githubusercontent.com/17033058/231010181-df6a1a35-cbf3-4e6b-8dec-509bdee29a91.mov

### How to test
- Open desktop and point it to the sync server
- Login as admin
- Open Reference data
- Select export tab
- Select all the data you want to export
- Click to export
- It should create/download a file with all the reference data you selected
- Use this file to import into a new Sync server
- It should import all the data you exported previously.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


passcod (changes requested):
Good stuff but needs tests!

edmofro (changes requested):
Thanks @renatokoyama! I'm actually surprised there are so few exceptional cases, it's very nice to have the consistency if so but I'm worried we might have missed some. For example, with the existing importer, do we include the `mergedIntoId` of patients that have been merged? Aaaah...actually I have just cottoned on to the fact that it doesn't need to be handled explicitly because they are also soft deleted when they are merged, and we are going through the Sequelize find mechanism which excludes them automatically.

So if all exceptional cases are handled, my only comment is that this needs some tests. Potentially could reuse the test data (spreadsheets) that area already in the repo for the import unit tests

edmofro (approved):
Given how at risk of going stale this is (evidently more high priority things are preventing you from being able to dedicate yourself to it @renatokoyama), I would say that while @passcod's suggestions are great, this is now good enough to merge

Approved from me, thanks for picking it back up Renato!

passcod (approved):

PR #3901, opened 2023-04-12
EPI-166: Implement (a limited form of) _include support

passcod:
### Changes

- Add `search.mode` to searchset bundle entries
- Implement `_include` by post-process.
  - Figure out which Resources and which IDs thereof need fetching, then do so
  - Bunch of checks and verifications upfront (during `resolveIncludes`)
  - As FHIR says that failing to retrieve includes is non-fatal, gather errors as we do so and bundle them up as warnings, so there's at least some explanation for why things don't look normal if there _are_ errors
  - Why can't we do `_include` in the main query? Lol because of Sequelize of course. We'd need [custom joins](https://6349415a5218b826a3409349--sequelize-site.netlify.app/docs/v7/core-concepts/querying/).
  - Includes are based on search params, so we reuse those to resolve the includes (filtering on `type: REFERENCE`)
  - Only supports references with a `type` and `reference` field, not "loose" references.
- Refactors:
  - Split up the FHIR search handling (ie the query.js file) into multiple thematic files so it can be somewhat less overwhelming
  - Split up the OperationOutcome class from utils/fhir/errors.js for searchability and concern-separation
  - Update the status of multi-level ordering ie EPI-202
  - Add a Processing error and re-classifies some existing FHIR errors as extending Processing instead of FhirError directly
  - In tests, unrelated to this PR, check that the length of the `entry` array matches the reported `total` count (where appropriate)
  - Use a common middleware to set the FHIR version header, and another common middleware to do FHIR error handling, instead of repeating these things across all handlers.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
Generally looks good! I'm pretty scared of the complexity that exists in the fhir module but you've done a really good job mitigating it, really appreciate all the comments along the way. Just have a few little quibbles here and there.

mcccclean (approved):
LGTM!

PR #3907, opened 2023-04-12
NASS-682: Completed imaging request table sidebar

tcodling:
### Changes

We are adding a tab called 'Completed' which will show the same table as imaging requests except with a filter to only show completed imaging requests. This is the same table with a couple of conditional changes based on a prop passed into the view.

We also added an advanced search section that adds the fields: 'Area', 'Department' and 'Completed' for searching the table.

### Checklist

- [x] Code is finished
- [x] Tests written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Looks pretty good I really like the abstraction though it still seems a bit awkward? I left some notes around, mostly is just tidying up. Also I was wondering if you had a chance to reorder the fields by Flic's message?

NavarroEmilioLuis (approved):
Hey, I think most stuff is good to go.

As a general overview I think that these should be more self contained and that work for upcoming cards should be dealt later. Otherwise I'm happy with the changes and the not-changes!

NavarroEmilioLuis (approved):
Seems appropriate!

PR #3910, opened 2023-04-13
EPI-202: Enable _sort for nested arrays

passcod:
### Changes

This is not a full resolution of EPI-202. The behaviour as implemented is stable, but it has known caveats and spec violations, documented in the ticket.

Those are:
- too many results are returned
- if there are nulls in the sort, the nulls are returned first in arbitrary/random order

The second is easy to mitigate using Postgres 15's NULLS LAST mode. However as we use 12/14 in deployments at the moment, that mitigation is not put in place yet.

The first is (afaik) very hard to solve, as described in the ticket.

Enabling this (with known caveats) brings the materialised FHIR API to parity with the old one, such that the old could be deprecated and removed (ie mSupply could finally move to the new one).

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (approved):
I'm happy with this moving on - we already documented a lot of what happens/needs to happen.

What I would emphasize is being crystal clear this is the current outcome, as I think will differ from the old API (it didn't had duplicates, did it?). I'm not sure if it will have any sort of impact, my guess is it won't.

On the second caveat (random ordering): I actually think nulls first makes sense, do you think we could just add an extra ordering by, say, upstream ID so at least we know it's stable? - this suggestion would be to always add that ordering at the end on any search that specifies ordering

PR #3915, opened 2023-04-13
EPI-350: Make discharge note content automatic

NavarroEmilioLuis:
### Changes

The discharges were getting the default note, after testing, we realized we wanted to specify these were automated. See comment https://linear.app/bes/issue/EPI-350#comment-a8630e14


mcccclean (changes requested):
I think we should include the reason info in the `systemNote` as well - why not!

mcccclean (approved):

PR #3931, opened 2023-04-18
SAV-133: Vaccine table export fix

tcodling:
### Changes

Blank values were causing errors on the export spreadsheet for vaccinations table. Added in empty string values to prevent errors when no value present


NavarroEmilioLuis (commented):
Hey, this looks fine to be honest, I left one suggestion but that's really nitpicking and just my personal preference, happy for you leave it like its now.

One thing before moving on: there are some files with linting errors, is that getting resolved elsewhere?

NavarroEmilioLuis (approved):
Oh yeah, makes sense. Happy for this one to move on!

PR #3932, opened 2023-04-18
SAV-113: Fix broken status filter

tcodling:
### Changes

Before this card we would show all the "Given" and "Not Given" vaccines by default in the list. This card adds a checkbox that allows the user to toggle between showing the "Not Given" vaccines. This wasn't working as the checkbox option that altered the where clause was being overwritten by the default WHERE clause that filtered the statuses to display in the list. Now with this fix it shows just "Given" vaccines by default and the user can now switch between the two modes.

### DIFF
--- a/packages/desktop/app/components/ImmunisationsTable.js
+++ b/packages/desktop/app/components/ImmunisationsTable.js
@@ -28,11 +28,11 @@ const getFacility = record =>
   record.givenElsewhere ? record.givenBy : record.location?.facility?.name;
 
 const columns = [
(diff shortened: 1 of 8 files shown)


NavarroEmilioLuis (commented):
I'm a little curious here, to which route is this hooked up?

Seems to me that `optWhere` would have to be something like:

```
{ status: { [Op.in]: [VACCINE_STATUS.NOT_GIVEN] } }
```

which looks a bit odd, are we sending all that from the client? Or is the endpoint handling it and the client is only sending the status constant?

PR #3933, opened 2023-04-18
SAV-131: View vaccine modal fixes

tcodling:
### Changes

A few fixes based on testing
- New header text
- Fetch proper value for "Given by" field
- Display "Given elsewhere" in status when vaccine is givenElsewhere


chris-bes (approved):
LGTM!

PR #3948, opened 2023-04-20
SAV-142: Allow the modal to update on editing not given vax

tcodling:
### Changes

This card was about fixing the mobile vaccine modal to update with the new details after submitting the edit form. Previously it would use navigation to go back and show the previous details from before the edit. 

This was fixed for vaccines with status "Given" but not the "Not given" vaccines. This pr just extends the logic to work when changing a vaccine to "Not given" and vice versa.


mcccclean (approved):
LGTM! But with a general note for the future that we shouldn't be storing data in the route; it should go in a react context instead. That's out of scope of this fix though, just a refactor that will have to happen later.

PR #3951, opened 2023-04-20
SAV-131: Add ability to view vaccines with no location

tcodling:
### Changes

Error was being thrown when trying to view vaccine with no location.


chris-bes (approved):
LGTM!

PR #3952, opened 2023-04-20
EPI-377: Add fine-grained permissions for reports, db or otherwise

stretchkennedy:
### Changes

This PR:
- filters the list of reports shown to the end user by those they have permission to run
- prevents users running restricted reports directly
- prevents users creating report requests for restricted reports

It allows the following new permissions:
- DB-defined 
  ```json
  {
    "verb": "run",
    "noun": "ReportDefinition",
    "roleId": "nurse",
    "objectId": "blah"
  }
  ```
- Static
  ```json
  {
    "verb": "run",
    "noun": "StaticReport",
    "roleId": "nurse",
    "objectId": "incomplete-referrals"
  }
  ```

It doesn't remove the previous permissions system, which was read permissions for another model defined per-report e.g. can('read', 'Encounter'). However, it makes it easier to do so in future by centralising the check.

It also doesn't check permissions on the central server, instead choosing to trust the connected facility server, since this is what we generally do with everything else in the system and I don't think it would sensible to fix this with one-off pieces (but my logic here should be fact checked by the reviewer 😄).

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
Great stuff, really clean. Love the tidy-ups too. Just some linter errors and a few minor changes to pick up.

I think there's also some bits of this PR that really highlight how the test system has pretty weak points around common data and permissions testing - I reckon it's worth having a look for any quick wins but if it needs deeper thought / a heftier refactor then I'm happy to declare those ones out of scope and maybe they should be an agenda item in tech strategy instead.

mcccclean (approved):
LGTM! I think the test failure is an intermittent one that's living on dev :weary:

PR #3955, opened 2023-04-21
Use swc to build csca

passcod:
### Changes

While using CSCA to renew VDS, noticed it wasn't building due to some typescript issues, but worked perfectly fine with SWC, so switched it over.

### Checklist

- [x] Code is finished


dannash100 (approved):
Looks great!

PR #3956, opened 2023-04-23
EPI-164: General refactor

NavarroEmilioLuis:
### Changes

There are a bunch of TODOs out of PR #3876. This PR solves them. These are:
- Move some files to top level (not actual FHIR resources)
- index.js files should just be exports
- Apply further (various) refactoring :)
- Apply all refactoring from Service Request to other models (cleanup)

I reckon this is easier to look at by individual commits.

Didn't want to block the other cards further although this might need an extra commit depending on the outcome of https://linear.app/bes/issue/EPI-472/run-some-explain-queries-and-check-if-queries-match but it would be fairly small - creating a helper and only touching two different files.


passcod (approved):

mcccclean (changes requested):
I'm not sure about putting the model definitions in `ImagingStudy/model.js` instead of eg `FhirImagingStudy.js` - if the main argument against content in index.js files was that it makes things harder to ctrl+p for, we probably want a corresponding rule that a file should be named after the thing it exports... :thinking:

mcccclean (approved):

PR #3964, opened 2023-04-24
SAV-141: Ability to edit a vaccination record on desktop

georgekamar:
### Changes

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Looks good! Left a couple of annotations here and there - I think these epics are troublesome because it's harder to compare what's going on.

Overall, I think these components look really, really messy, the rendering logic is filled with conditions in several spots that make me wonder if that justifies changing them. I think there's also an improvement to handling the modals: instead of having three separate states and setters, there could just be one that specifies which modal is on (guaranteeing they can't be all open at once). Definitely out of scope but wanted to call it out now.

NavarroEmilioLuis (changes requested):
Looks good! Left a couple of annotations here and there - I think these epics are troublesome because it's harder to compare what's going on.

Overall, I think these components look really, really messy, the rendering logic is filled with conditions in several spots that make me wonder if that justifies changing them. I think there's also an improvement to handling the modals: instead of having three separate states and setters, there could just be one that specifies which modal is on (guaranteeing they can't be all open at once). Definitely out of scope but wanted to call it out now.

NavarroEmilioLuis (approved):
Beautifully done, thanks so much for going through these!

PR #3973, opened 2023-04-27
EPI-164: Add missing spots

NavarroEmilioLuis:
### Changes

Checking why the lab requests were not being materialised I noticed that:

- Model upstreams were missing (the ones that are new because of lab requests).
- Not possible to de-structure stuff created errors
- Currently the service requests table has a foreign key relation in upstream ID, this needs to be undone in order for us to save lab requests IDs there too.


mcccclean (changes requested):

mcccclean (approved):
Great stuff!

PR #3981, opened 2023-05-01
SAV-131: Vaccine view modal fixes

tcodling:
### Changes

This is a small bug fix for an issue that was throwing an error whenever you opened an elsewhere vaccine.

I have also added some logic to fetch the facility of the encounter for the vaccine if no location exists (if the user adds an other vaccine) as this was previously showing up with an empty value when viewing the vaccine modal


NavarroEmilioLuis (approved):
Fix looks good to me!

PR #4004, opened 2023-05-04
SAV-107: Covid certificate printing date fix

tcodling:
### Changes

There was another bug with the emailed covid vaccination certificates where the printing date was wrong as it was accidentally fetching from the metadata field. This has now been resolved by creating a new column on the certificate notification with a printing date that can be accessed when sending the certificate


dannash100 (approved):
🅿️ arty time

PR #4009, opened 2023-05-05
no-issue: unnecessary re-renders, data fetches causing flashing tables on lab request row click

dannash100:
_Note from Kennedy: we're tracking this in https://linear.app/bes/issue/EPI-498/lab-requests-table-causes-unnecessary-refetchingrefreshes now_

this was a pretty bad situ doing 5 or so re-renders and data fetches when clicking table - this caused some very noticeable flashes

I think ultimately the issue at root of this is the way we have a detachment between route and its associated data.
I.e you have to pre-prepare the data before pushing to the route. Where the ideal state would be that the route handles the loading of the content from params. 
This was something I looked into improving when doing routing changes but the scope was too broad at the time.

To fix the combination of these three actions was required
- Make the load functions useCallback to avoid them retriggering render
- Move the hooks to one level up and pass through as props
- Make the LabRequestTable a memo

**Before fix: Flashes 5 or so times**
https://user-images.githubusercontent.com/38335330/236359870-be016b65-2984-44e8-bee1-87eb1a444476.mov

**After fix: No flash**
https://user-images.githubusercontent.com/38335330/236359897-8ee5f631-6f6d-419f-bf90-129cb09f845b.mov

### Changes

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


tcodling (approved):
Not at complicated as I thought. gg 🥇

PR #4011, opened 2023-05-05
TAV-622: the shared situation

passcod:
### Changes

This looks massive, and it kinda is, but not as much as it could be. It's very hard to split up further as all of it is inter-related.

- finally fixes the shared situation, making it into a normal package that behaves correctly re installing and resolution
  - no more double-yarn-install!
  - no more `shared-src` (this rename is the bulk of the "files changed")
  - no more adding packages to sync-server and lan and desktop and meta just because one was added to shared
    - all now-redundant packages have been sorted out here, which has dropped a couple of now-unneeded dependencies
  - named `@tamanu/shared`, but...
    - aliases for node, webpack, and eslint so that `shared/...` still works for now
  - fixed all the cases of imports inside shared to `shared/...` and imports outside shared to `../../..../shared-src/...`

- switches all the server packages to SWC
  - not for mobile and desktop as these are a lot more complicated to do and work as-is
  - speeds up builds massively and tests in CI are a bit faster now
  - ⚠️ disallows some patterns of Jest spies, notably spying on `import * as foo`s, so some tests have to be reworked
  - ⚠️ because SWC works a lot better and is a lot more parallel, it can happen that when running a large test suite, your computer gets overwhelmed! On Linux, setting a high nice value (`nice -n10 yarn workspace lan test` for example) helps a lot.

- makes a new package `@tamanu/build-tooling` that contains all the babel/webpack/swc tooling and the common webpack config
  - makes it easy to upgrade these things in one place
  - removes some bloat/repetition from the package.jsons
  - mostly needed to make containerisation easier, but it's nice on its own too, which is why it's here.

#### shared exports

Due to making this a "proper package", the built files in shared live in `packages/shared/dist/...`. What that means is that every subpath import, like `import {} from 'shared/constants'` no longer works directly. There were few options:

- rewrite every single import to shared in the style `shared/dist/...`
- use [Package Exports](https://nodejs.org/api/packages.html#exports), a relatively recent Node/Webpack/etc feature that lets you define where subpath imports into a package point to.
  - this means to write a table of exports into shared's package.json
  - this also means that things no longer work automagically! if you add a file to shared that's meant to be used in an import directly, and it's not covered by the existing rules in the exports table, you'll need to also add a rule to cover it.

I picked the second option. Instead of rewriting tens of thousands of imports, I hammered out [a 20-line exports table](https://github.com/beyondessential/tamanu/blob/612f528ae66a6d918a72491fbb403be7b927896b/packages/shared/package.json#L5-L25) which covers all the paths that we currently import.

#### `shared` vs `@tamanu/shared` imports

The new name of the package is `@tamanu/shared`, but we refer to it as `shared` all the time. Fortunately, there's another recent feature to the rescue: [Package Imports (Aliases)](https://nodejs.org/api/packages.html#subpath-imports). That lets Node.js figure it out: Webpack and ESlint have their own aliasing configuration.

In the long run, though, we want to refactor-as-we-go all shared imports to use `@tamanu/shared` instead properly, so we can drop these workarounds. The desktop package can't deal with this workaround, so it has had its imports rewritten already.

### Checklist

- [x] Every shared import resolves
  - [x] Sync-server
  - [x] Meta-server
  - [x] LAN
  - [x] Shared (internally)
  - [x] Desktop
- [x] Enable SWC everywhere (except mobile and desktop)
  - [x] Shared (prior)
  - [x] Sync-server
  - [x] Meta-server
  - [x] LAN
- [x] Tests are adjusted
  - [x] Sync-server
  - [x] Meta-server
  - [x] LAN
  - [x] Shared
  - [x] Storybook
  - [x] [That one facility sync test](https://github.com/beyondessential/tamanu/pull/4011/commits/f6aafb748d7f38e8d62a4a321762ca9c5a3c44b9) that doesn't work
    - Consider: adding spies in the module itself enabled on demand? 
- [x] Lints are adjusted
  - [x] Sync-server
  - [x] Meta-server
  - [x] LAN
  - [x] Shared
- [x] CI/CD is adjusted
- [x] Redundant package requirements in dependencies from old shared pattern culled
- [x] Migrations still work
- [x] Production build for current deployment pattern is tested
  - [x] Desktop package
  - [x] Sync-server package
  - [x] LAN server package
  - [x] Meta deploy


mcccclean (commented):
Just leaving some comments as I see it's been receiving updates while I was reviewing so don't want to call it yet! But this is a heroic effort and has been such a long time coming, this rules

mcccclean (changes requested):
Huge effort, amazing stuff! LGTM except two comments that I reckon could use a link to a linear ticket & ahh there are some conflicts too, sorry I couldn't get to this sooner :pray:

mcccclean (approved):
:tada: 

(slightly rubberstamped as there's a LOT but I've scrolled down the whole thing at least :sweat_smile:)

edmofro (changes requested):

edmofro (approved):
Codeowner stamp for sync sections as flagged, thanks for this hefty and lifechanging piece of work @passcod 😀

PR #4021, opened 2023-05-08
SAV-198: Code Refactor

georgekamar:
### Changes

There seems to be a discrepancy (sometimes) between the number of patients in the tables and the numbers in the dashboard elements. This PR changes the way the facility filter is applied to the dashboard element API call to match the way it is applied in the patients table. It likely won't change anything but it'll be good for consistency in any case.

EDIT: 
This won't change anything but it's good for consistency, this is now a refactoring PR

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
Hmm, does this fix... actually work properly? :confused: The facility server should just remap the `:local` parameter to be the server's configured facilityId, which really should match the `useAuth().facility.id` in 100% of cases...!

The other change request (assuming that we can confirm this works) is to remove support for the `:local` parameter on the backend, now that nothing is using it.

mcccclean (approved):
Yeah, ah, yep this is much simpler now! Great little clean up, thanks :pray: (and yeah I we don't need to worry about affecting 1.26 as this is just going into dev)

PR #4026, opened 2023-05-09
EPI-451: Include missing triggers

NavarroEmilioLuis:
### Changes

This was caught pretty late. I *borrowed* the testing logic from a similar test cause I thought it made sense.

Anyway, this still needs tidying up, and it will depend on:
- Do we want to add versioning trigger to `fhir.encounters`?
- Is this going to be hotfix to staging? (Lab test panels work isn't there)


passcod (approved):
lgtm otherwise

PR #4027, opened 2023-05-09
EPI-413 Change reference language across board

biaoli0:
### Changes

- Add new setting in localisation
```
  "localisation": {
    "data": {
      "fields": {
        "clinician": {
          "shortLabel": "Clinician haha",
          "longLabel": "Clinician"
        }
      }
  }
```

- Add a `localised()` to manage the localisation for label 'clinician'. We can reuse it for additional localisation words in the future.
- Limitation: There are a couple of localisation in `share-src` that we can't manage. like Excel report exporting. I personally feel we can revisit it after settings came across to be database based. 

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Aah, I have to push back against the `localise` function here I'm really sorry! :pray: The task of switching the whole app to be localisable will be a pretty big one, and I think the details of how we go about it will need a bit more of a considered tech design.

The main issue I see with this approach is that it requires manually maintaining a separate list of localisable strings when we already have that in the form of the localisation config contents. We could automate the construction of that lookup table by reading the config, but it will fall over if we have two different terms that have the same English translation as each other, which we already do. 

So - yeah I'm sorry that it's a little more annoying, but until we have properly designed out a more automatic way of doing this we should just use the fully qualified localisation paths here, `general.clinician.label`. 

I'll also draw your attention to the `LocalisedField` component, which correctly localises the label for a field input, and to the `LocalisedText` component which is for localising text that appears outside of a form, which should be used in a lot of these cases.

mcccclean (approved):
LGTM! :+1:

I'll say that as we expand our localisation support, we likely won't be able to do manipulation of text (like `Supervising ${text}` and `${text} must be selected`), as we'll be using it for supporting multiple languages rather than just switching out pieces of terminology here and there. Eventually those strings will just have to be separate entries in the localisation db... but that doesn't have to happen now, and the work to switch them out will be the same whether we do it now or later, so I'm perfectly happy for that to happen down the track. Just wanted to explain my thinking about building out localisation features!

anilvkaran (approved):

PR #4037, opened 2023-05-10
SAV-107: Fix broken sort for vaccination certificate UVCI

tcodling:
### Changes

After changing to the new date-string format the sort function that is used to decide which vaccine the preview UCVI code is generated from was broken. I have extracted this out into a date util that can be used when we need to sort by date on the front end.


dannash100 (approved):
YUSS

PR #4040, opened 2023-05-11
SAV-159: Reverse where null dates show when sorting vaccine route

tcodling:
### Changes

Felicity wanted vaccines with a NULL date to act as if they are the "oldest" date when sorting. Currently they show up at the start when sorting new to old and at the end when sorting old to new and we want the reverse behaviour.


mcccclean (changes requested):
This was absolutely breaking my brain for ages, went and did something else for a sec and came back to it and thought of this change immediately. What do you reckon?

It also strikes me that this is a backend change that doesn't have any accompanying unit tests but I think with the time pressure on this module it's not critical.

mcccclean (approved):
Oh yep, LGTM - usually I try to avoid `let`s whenever possible but I agree with the linter here, a nested tern is worse.

image
awful creature :-1:

PR #4045, opened 2023-05-11
NASS-678: Select survey error on select clear icon button click

dannash100:
### Changes
Fix for ❌ ISSUE_1 NASS-678

Previously it was not possible to trigger change event with an empty id value
So have to account for this to allow clearing of the field.

**Fixes:**
image (35)

image (34)

### Checklist

- [x] Code is finished
- [x] Tests are written


tcodling (approved):
Legen

wait for it

dary

PR #4048, opened 2023-05-12
NASS-678: Bugfixes for clearing search fields

tcodling:
### Changes

After a round of tests there were a few edge cases that needed to be resolved:

- [x] Implement proper search fields for documents tab
- [x] Disable "Begin survey" button when survey category is cleared
- [x] After clearing a facility from dropdown for death workflow, it will try to insert a facility id that is an empty string
- [x] "X" doesn't actually clear search inputs for search tables. It will still store the old value that was used last time the user pressed "Search"
- [x] Clearing the area dropdown on the imaging table will "break" the search (always returns nothing afterwards)
- [x] Select dropdown seems to be shorter in many places than other inputs


dannash100 (approved):
Amazing work on this imho!

PR #4049, opened 2023-05-12
EPI-444: Add program survey result question type

NavarroEmilioLuis:
### Changes

Everything looked already in place, but I found the main issue to be that desktop was using program data element `id` as field names instead of `code` (like mobile).

This is supposed to be a "quick" fix - given this is aimed for 1.26 and has to be hotfixed. There is a lot of clutter with these components and I don't think now is the time to fix them. Further potential adjustments would be: cleaning up unused logic like extracting a `hidden` prop not used anywhere, actually re-using the helpers and/or getting rid of the `calculateSurveyResults` subcommand.

Also, I believe the simplest idea is to have desktop use `code` instead of `id`, not only it will be more efficient, it makes more sense because the visibility criteria uses `code` for it's logic. The problem being that changing it can potentially break a lot of stuff - it seemed pretty safe, but we're also out of time, I prefer to play it safer (the scope of this fix only touches one place, easily traceable).


biaoli0 (approved):
Code looks good! Just one question, I've seen mobile has the same function `getResultValue`, any reason why they don't share?

PR #4050, opened 2023-05-12
SAV-106-SAV-107: Don't manipulate nulls for certificates

tcodling:
### Changes

Another last min tweak for the vaccine certificates. After changing the null behaviour we made all vaccines with unknown date show on the top. We actually want them to be at the bottom like they were before.


chris-bes (approved):
Cool

PR #4059, opened 2023-05-15
NASS-717: dynamic select story improvements - mock suggester behavior

dannash100:
### Changes
Tiny change after doing dev review of ticket.
Stories felt a bit broken due to the way that suggester endpoint was mocked


tcodling (approved):
epic changes

PR #4060, opened 2023-05-15
EPI-486: Include medici report workaround

NavarroEmilioLuis:
### Changes

Pretty straightforward, just cherry picking the workaround included in v.1.25.3 to dev

Dev needed an extra change because there was a big refactor (that won't make it to staging) that renamed that import.


stretchkennedy (approved):
Same changes as reviewed previously, LGTM 👍

PR #4072, opened 2023-05-15
NASS-682: Only require results when filter is applied

tcodling:
### Changes

We currently were only including imaging requests with a corresponding result in the new completed table. This means that the completed table was excluding imaging requests without a saved result. This should be impossible now with the creation workflow but there are records like this in prod so we want to make sure they still show up in the completed table. 

I have added in a line so that it will show all completed imaging requests (including those with no result) unless a result filter is applied and then it will only return results with that date


dannash100 (approved):
yayer amazing!

PR #4076, opened 2023-05-15
SAV-168: Fix flaky lab test

tcodling:
### Changes

There was a lab test that was failing sometimes due to the status being assigned randomly. This meant sometimes it was assigned as an error or cancelled lab which means it wouldn't be fetched by the query which filters these out automatically. (at least i think this was the cause. It happens quite rarely and this makes sense to me as the cause)


NavarroEmilioLuis (approved):
Amazing... thanks for the changes!

PR #4079, opened 2023-05-16
NASS-752: advanced imaging search

tcodling:
### Changes

We want to add an advanced search to the basic imaging request table with the fields Area, Requested By and Department. We also want to exclude completed requests as they are now shown in a new table

Many of the changes in this card are around updating the way that the tables filter by status. When making the completed table in another card I set this up so that each table could filter by a single status. When I got to this card I realised that we actually need to filter by two statuses so I actually ended up making more work for myself 🤦 Now instead of giving the route a status and doing all the logic based on that, we now have a seperate memoryKey used to track the different search parameter states and then an array of statuses that we want to be shown in the table which all the logic is based off.

Screenshot:

### Checklist - This is my first 1.27 card so have added a couple of (at least I think) necessary checkboxes since it needs to be tested before merge

- [x] Code is finished
- [x] Tests are written
- [x] Passed code review
- [x] Deployed to test branch
- [x] Passed testing on test branch

_Upon merging:_


biaoli0 (commented):
Hey @tcodling I would like to discuss a different approach on handling different status. Love to know what you think.

biaoli0 (approved):
Thanks @tcodling. Keeping the code pattern looks good to me. Yeah using params could introduce a new state and it could be hard to track.

PR #4095, opened 2023-05-19
EPI-467: Add configurable validation to display ID

NavarroEmilioLuis:
### Changes

Okay, card requirements changed after the initial draft. Given that it was also very messy I decided to roll back a couple of commits and tackle it from a different angle.

The gist of this work is that we want to be able to validate the display ID on edit, comparing it to a regex that should be configurable. Right now, we only want to validate the display ID field if it's being edited, so this is more close to that logic and also more contained!


biaoli0 (approved):
LGTM! Love the new angle to handle this. Just got a minor suggestion, can pre approve.

nick-beyondessential (approved):

PR #4101, opened 2023-05-22
SAV-106: Vax table rendered multiple times

tcodling:
### Changes

When visiting the vaccines tab for a patient, the table will re-render twice.

With help from Tom we discovered that this table re-renders every time due to the fact that useEffect uses an object for a dependency. When we instead compare the stringified object as a useEffect dependency, it works as you would expect and only re-renders when the fetch options change.

We felt this was a suitable time to ignore the linter as the dependency was causing the problem but open to alternatives

Previous behaviour

https://github.com/beyondessential/tamanu/assets/58054247/aed8bb7f-2a28-44c8-9779-3c26cfadc581


chris-bes (approved):
Nice catch!

PR #4102, opened 2023-05-22
SAV-100: Stop allowing the user to scroll only the header

tcodling:
### Changes

We discovered a bug with the mobile vaccination table header where the user was able to scroll the header independently from the rest of the table. This is due to the slightly roundabout logic that we needed to do to allow the header to act sticky and also scroll horizontally.

The fix I did was to simply add the prop that disables the ability to scroll the header row. The scroll still works as expected when the table is scrolled horizontally


dannash100 (approved):
Discussed that this doesn't seem to break the stickyHeaderIndicies so look lit

PR #4104, opened 2023-05-22
NASS-678: Autocomplete form state + search fix

tcodling:
### Changes

We discovered a couple of issues with state not tracking properly with the new clear button for autocomplete fields. I discovered this was a result of me overcomplicating things and calling to multiple of the built in methods causing weird behaviour. When I just use the single built in method properly to change the form/input state everything works smoothly as I would expect :)


chris-bes (changes requested):
Just 1 more thing

chris-bes (changes requested):
Just 1 more thing

chris-bes (approved):

PR #4105, opened 2023-05-22
EPI-492: Add ability to configure vitals

NavarroEmilioLuis:
### Changes

Vitals validation can now be grouped by age. Though it obviously needed a new format, it shouldn't be a breaking change so still needs to support without age grouping.


mcccclean (changes requested):
Good first pass at this - just have some suggestions around simplifying the architecture a bit.

mcccclean (changes requested):

mcccclean (approved):

PR #4109, opened 2023-05-23
TAN-2140: Move crudHelpers to shared

Alastair-L:
### Changes

Before using crudHelpers on the sync server, they need to be moved to shared. I've split this step out to make the subsequent diff cleaner.

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (commented):
First, thanks for splitting it up. Though I'm not 100% sure we want to have crud helpers in the central server... My reasoning here is that central server endpoints should be more robust and thus we probably want to force it to be careful

There's also the weird fix of having the `models` attached directly to the context instead of through the `store`, which I think it's always been confusing + the weird permission thing that I can't find now? (I think both are on the follow-up PR)

mcccclean (approved):

PR #4111, opened 2023-05-23
SAV-131: Disable header "x" button for vaccine modals

tcodling:
### Changes

We are moving away from having an x button in the corner of all of our modals. I made a new prop (disableHeaderCloseIcon) for disabling this when doing the vaccination work but Ethan did the same thing with a more suitably named prop (cornerExitButton) on dev which was overwritten by the feature branch merge. 

This fix essentially removes all of the x buttons from the vaccination modals while using the new prop created by Ethan instead. This should hopefully stop any merge problems :)


NavarroEmilioLuis (approved):
LGTM, guessing this needs to go to staging, too?

PR #4117, opened 2023-05-24
EPI-490 Add confirming modal for discharged form

biaoli0:
### Changes

- Add a confirming modal for form component
- Add confirming modal for discharged form 
### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Hey, I noticed that you had a back and forth with Felicity, very good at keeping an eye on those small details. I have a couple of comments here and there but a really big change request if possible - I want to take a step back here, I'm not sure this is the right approach:

I feel like the form component should just handle itself and forget about the confirmation modal because it's already doing a lot. Also it seemed from the video that this solution adds a modal on top of a modal which I think should never happen.

Maybe we already have something to handle this situation with `PaginatedForm`, even though this one is not paginated at all. 🤔 Still, might be worth a try before a new implementation. I'd suggest trying a simpler approach reusing something or creating a separate component for this.

Happy to hear your thoughts too if you disagree, it looks like it needed quite some work to figure it out - which I think is fine but I think definitely not necessary for a 2 pointer!

NavarroEmilioLuis (approved):
Sorry it took me this long to completely check.

This looks good to me - it does have that hacky bit but to be honest it seems OK because of the scope. I think that going forward if we want this kind of approach elsewhere we need to seriously consider abstracting it into its own thing (which is definitely NOT a 2 point task!).

PR #4124, opened 2023-05-24
SAV-100: Header scroll behaviour tweak

tcodling:
### Changes

Klaus discovered some more weird behaviour with how the vaccine table scrolls on mobile where you can swipe between tables on the header. In order to resolve this I removed the ability to swipe between tables (which was never really useful for this table in the first place, in fact I think it was probably annoying). This means it now works in a much more predictable way although not as ideal as I would like due to technical limitations due to the layering of the scrollviews for this table (ideally the whole table will scroll when the header is scrolled but I could not get that to work in a reasonable amount of time).

Before:

https://github.com/beyondessential/tamanu/assets/58054247/0a9cd93c-2043-4d29-9cce-fdfd357e0db3

After

https://github.com/beyondessential/tamanu/assets/58054247/a9a68778-665c-4f6c-afd0-152bc569449d

### DIFF
--- a/packages/mobile/App/ui/components/Table/index.tsx
+++ b/packages/mobile/App/ui/components/Table/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { StyledView, RowView } from '/styled/common';
+import { StyledView, RowView, StyledScrollView } from '/styled/common';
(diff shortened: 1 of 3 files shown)


dannash100 (approved):
Love it

PR #4126, opened 2023-05-25
NASS-678: Add missing search components + select sizing

tcodling:
### Changes

Some of our search fields in the lesser used search tabs didn't have a proper search component implemented. There were also some select components that didn't have the proper size prop set so this PR just adds those in


dannash100 (approved):

dannash100 (approved):
YESSIR

PR #4128, opened 2023-05-25
SAV-132: Populate location and department in vaccine forms for mobile

chris-bes:
### Changes
- Populate location and department in vaccine forms for mobile

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/mobile/App/models/Encounter.ts
+++ b/packages/mobile/App/models/Encounter.ts
@@ -148,26 +148,32 @@ export class Encounter extends BaseModel implements IEncounter {
     await Patient.markForSync(this.patient);
   }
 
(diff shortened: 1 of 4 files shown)


georgekamar (approved):
Hey @chris-bes , I just have one question about this: why isn't locationGroupId also included in this since it is tied to locationId ? I would think it would be derived from the location obtained but since it's just an ID not sure how it can be derived.

Im approving this to not block anything in case I am just missing context, since other than that it looks good to me. If Im not then I think we should add the locationGroupId (the area) to this.

P.S: I did not comment everywhere where locationGroupId is missing to not be too repetitive so if it needs to be added don't count on my comments only 🙏🏼

PR #4139, opened 2023-05-26
NASS-733: Staging fix - checkboxes

NavarroEmilioLuis:
### Changes

Problem was caused by upgrading library Formik to v2, which changed the way checkboxes work internally. After a long investigation, I believe this is the correct approach to solving it.

The main reasoning being that the `value` **prop** on HTML inputs with type checkbox only control what gets sent over. However we're managing state, and what gets sent is actually the state of the field, which is a different value. Using it as query params will also work because we only check that the key exists, we never care if it's actually `"checked"` - `"true"` reads just as right to me.

A lot of background info found on the card if you're interested.

PS: had a bit of back and forth with McLean, so pretty much all commits co-authored by as well! Thanks again 🙏

### DIFF
--- a/packages/desktop/app/components/Field/CheckField.js
+++ b/packages/desktop/app/components/Field/CheckField.js
@@ -7,6 +7,11 @@ import FormControl from '@material-ui/core/FormControl';
 import FormHelperText from '@material-ui/core/FormHelperText';
 import { Colors } from '../../constants';
 
+/* 
+  Note that the Checkbox value prop only controls what gets sent,
+  not the checkbox state. It's also worth noting that usually forms
+  will send the state value, not the prop value.
+*/
 const CheckControl = React.memo(({ value, ...props }) => (
   <Checkbox
     icon={<i className="far fa-square" />}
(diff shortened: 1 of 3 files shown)


tcodling (commented):
LGTM! I think we actually want it to go into dev first and then get cherry picked into staging though.

Also the build was failing on what I think was a unrelated test so I set it to restart

mcccclean (changes requested):
This LGTM but duplicating this fix across multiple controls is a bit awkward to me - we still have a situation where there's a weird "trick" to implementing a checkbox that a dev needs to know if they want to do it.

Really, the `value="true"` bit should be only in CheckField.js, and the instances of Checkbox in DisplayIdField and DischargeForm should be replaced with instances of CheckField (and omit the checked prop - this means all of our input controls can keep using `value` for their actual value rather than these two mysteriously using `checked`).

Also I think `value="true"` should be moved to _after_ `{...props}` in CheckField - considering it breaks if it's set to anything else, we don't want to let people override it.

(note that these aren't problems introduced in this PR - if there's time pressure this work can easily become a maintenance task for later)

mcccclean (approved):

PR #4141, opened 2023-05-26
TAV-623: Build containers in GitHub Actions

passcod:
### Changes

Builds container images for central and facility on every commit to dev and for non-draft pull requests. Images are tagged with the git commit sha (~immutable) and with the branch name (~mutable tag to the latest built image for the branch).

Building those images all the time (and not just on deploy) makes them part of CI in a way, so that if a change breaks the container build, it can be noticed and corrected early. It also makes it easier to deploy in a way that's more separate from Github (given that testers tend to avoid dealing with it). However, we "save" on some builds by skipping draft PRs. Undrafting a PR triggers a build though, so there's no need to push a commit after undrafting to trigger one.

ARM64 images can be built with this setup (commented out), but in GHA this is very very slow (>5x) at the moment as it's doing it through QEMU. Native ARM64 Linux GHA runners are on GitHub's roadmap for this year.

Building the meta server is also possible but currently commented out as it's not used for now.

Images are pushed to the private AWS ECR registry for tamanu, and authenticate to AWS via GitHub's OIDC token, thus requiring no AWS secrets to be stored here.

### Checklist

- [x] Code is finished

_Upon merging:_

### DIFF
--- a/.github/workflows/cd-containers.yml
+++ b/.github/workflows/cd-containers.yml
@@ -44,7 +44,7 @@ jobs:
           - name: meta
             path: meta-server
 
-    name: Build ${{ matrix.package.name }} container for ${{ matrix.platform }}
+    name: Build ${{ matrix.package.name }} container for ${{ matrix.platform.arch }}
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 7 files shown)


mcccclean (approved):
LGTM. 

I know I have a habit of accidentally opening draft PRs in non-draft mode and then drafting them, will that be fine / just slightly wasteful?

PR #4147, opened 2023-05-29
NASS-678: Select state fix for forms and searchbar

tcodling:
### Changes

Another weird interaction has been discovered during testing today related to the clearing of select fields. When pressing the "x" button on the select component to clear, we need to set a value for the field state to reset to.  We discovered that if we set this to null, search fields won't work properly as they try to search for an empty string in the searchParamaters. However if we set this to undefined, search will work properly but we no longer have the option to edit existing records through forms to null as undefined takes it completely out of the edit object sent to the backend.

In order to solve this I needed to set an undefined value when clearing a select field on a search bar and null when clearing a select field on the form. I achieved this by adding an initialValue to the CustomisableSearchBar component `isSearchForm: true` which is checked on the component level and will change the field clear value accordingly

My my main concern for this fix is around the setting of initialValues with my new custom isSearchForm value but I couldn't figure out any other way of setting a value on the searchbar level that can flow down and be checked on the field level. Very open to ideas! :)

### DIFF
--- a/packages/desktop/app/components/Field/Field.js
+++ b/packages/desktop/app/components/Field/Field.js
@@ -6,9 +6,19 @@ import { TextField } from './TextField';
 import { FORM_STATUSES } from '../../constants';
 
 export const Field = formikConnect(
-  ({ formik: { errors, status }, name, component = TextField, onChange, helperText, ...props }) => {
+  ({
+    formik: {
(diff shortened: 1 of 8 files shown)


chris-bes (approved):
LGTM!

PR #4148, opened 2023-05-29
Fix conflicts for staging v1.26.0 -> Master 

mcccclean:
Fixing conflicts to get staging-1-26-0 ready to merge into master. Mostly the conflicts were just the version number (which is all "keep ours" so doesn't appear in this PR), but there's some FHIR stuff as well.


NavarroEmilioLuis (approved):
LGTM, really weird that it had a conflict (don't recall those lines having anything else changed??) but everything looks in place here

PR #4152, opened 2023-05-30
NASS-774: Labs search fixes

tcodling:
### Changes

A few bug fixes related to the labs search bar discovered during regression testing.

- [x] Fix clearing state for autocomplete fields on search bars
- [x] Use id comparisons filters for lab fields changed to select/autocomplete
- [x] Change DynamicSelectField to autocomplete (this is in the codebase already but is actually being implemented properly in another card)


stretchkennedy (changes requested):
Mostly LGTM - two questions that might lead to changes, one optional nitpick

stretchkennedy (approved):

PR #4154, opened 2023-05-30
NASS-682: unwanted sorting persistence across similar imaging request routes

dannash100:
### Changes
Due to the way the routes where set up using the same underlying file in the root of route but dynamically changing status prop this was causing the table to not properly reinitialise when changing tabs and cause sticky behaviour when changing between all and completed tab ( detail: initialSort with DataFetchingTable was never changing beyond first chosen route)

Discussed solution with Tom and he had great idea to reduce not dry side while preserving the fix
So I applied to labs aswel

Also removed some overly pre-cautious option chaining from last pr

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


mcccclean (changes requested):
Oh yeah hm, I was going to complain that it's kind of DRY to have those components so similar but it does make the routes file a lot cleaner so yeah I think it's an improvement.

I do wonder if it might trip someone up when they go to make some modification to the table, now they'll have to do it in two places - what do you think of just exporting the `Completed` variant from the `ImagingRequestListingView` file as well, so that when an edit needs to be made it's obvious to the maintainer that there's two bits?

mcccclean (changes requested):
One last fix! Almost there :pray:

mcccclean (approved):

PR #4156, opened 2023-05-30
NASS-682: Wrong number of rows per page sorting by NHN

tcodling:
### Changes

In order to get sorting to work for imaging requests specifically for the NHN column, we added a value of subQuery: true to the findAll method used on the get imaging route. This actually had some weird side effects around pagination so we needed to remove this for now. After some discussion with Michael on linear we decided to not make the NHN sortable as it was not a particularly useful sort column and was causing technical problems not worth trying to solve right now given the release timing.


dannash100 (approved):
With Michaels suggestion I approve! I wonder if we make ticket for maintenance to resolve this eager loading sequelize sort issue for nhn or if we just assume that no one will need to sort by NHN - the use cases are hard to imagine - but im not entirely across such things

PR #4167, opened 2023-06-01
SAV-274: Stop from setting to null when clearing appointment autocomplete

tcodling:
### Changes

The appointment calendar has some custom logic for searching as the input changes so we had a suggester error pop up here. I solved this by setting to an empty string when the value it is being changed to is null to make it an empty string. This allows for the proper suggester logic to run as an empty input without throwing errors.

### DIFF
--- a/packages/desktop/app/components/Field/AutocompleteField.js
+++ b/packages/desktop/app/components/Field/AutocompleteField.js
@@ -234,7 +234,7 @@ export class AutocompleteInput extends Component {
   };
 
   handleClearValue = () => {
-    const { onChange, onClear, name, required, form } = this.props;
+    const { onChange, name, required, form } = this.props;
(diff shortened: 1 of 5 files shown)


dannash100 (commented):
Discussed a few other options but I think this is the best.
they included
1. add required prop to AutocompleteInput - but this field is only _required_ because of the way the field is being handled (controlled with state) with filterValue. - So its kind of random to then use required key here? But this would fix it also
2. use onClear to set state value to '' - but this kinda feels wrong because a clear is a change event ( I wonder if there is any cases that we would need this ) - might be worth a look over

Check this file out which uses the same as above but different solution (number 2) - this is the only place using onClear so we could remove all together?
tamanu/packages/desktop/app/components/Field/LocationField.js

dannash100 (approved):
Love it

PR #4168, opened 2023-06-01
EPI-517: Fix materialiser?

NavarroEmilioLuis:
### Changes

Ok so after lots of head scratches, I realized the FHIR worker was simply dying out, always thinking it was still being processed when it wasn't. I don't really know the cause of it or if we can do anything about it (seems that we expect this to happen, and its caused by the database transaction).

I believe we don't have enough data so I want to push some logs that might shed some light - I guess preferably using honeycomb to have a couple of days of logs.

The first commit is the actual "fix", which is simply ensuring the process gets restarted. Unfortunately it still fails quite a lot and it will take a while to go through 8 mill records, so better get started.

### DIFF
--- a/packages/shared-src/src/models/fhir/Job.js
+++ b/packages/shared-src/src/models/fhir/Job.js
@@ -106,13 +106,7 @@ export class FhirJob extends Model {
         );
       } catch (err) {
         // 40001 is the Postgres error code for serialization failure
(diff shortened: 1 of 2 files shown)


passcod (commented):
🤔

passcod (approved):

PR #4174, opened 2023-06-02
NASS-682: Fix incorrect pagination on ImagingRequest table

mcccclean:
### Changes

NOTE: This PR also includes a little bit of a rearrange of the existing tests to make it a bit cleaner to write the new ones. I haven't touched the internal logic of the other tests, the only new ones are in the `pagination` at the bottom.

The query for ImagingRequests was joining the ImagingResults relation in a strange way - the result would include a row for each ImagingResult and still obey the LIMIT 10, and then Sequelize would filter out the duplicates (reformatting them as an associated array on the final ImagingRequest object), resulting in fewer results per page than we'd expect.

The fix was to replace the included relation in the querybuilder with an included _attribute_ using a raw SQL subquery.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/ImagingRequestsTable.js
+++ b/packages/desktop/app/components/ImagingRequestsTable.js
@@ -28,9 +28,7 @@ const getPatientName = ({ encounter }) => <PatientNameDisplay patient={encounter
 const getPatientDisplayId = ({ encounter }) => encounter.patient.displayId;
 const getStatus = ({ status }) => <StatusDisplay status={status} />;
 const getDate = ({ requestedDate }) => <DateDisplay date={requestedDate} timeOnlyTooltip />;
-const getCompletedDate = ({ results }) => (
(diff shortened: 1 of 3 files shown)


chris-bes (changes requested):
I just have 1 question about the removed filter, otherwise LGTM!

dannash100 (approved):
@chris-bes This was a hectic fix thanks for all involved 🥳 ! Feels a little bit like pushing sequelize to do what it doesnt want to do in this endpoint - so probably really good candidate for rewrite in SQL
Do you think there is any major potential performance implications of double subquery in this endpoint now?

dannash100 (changes requested):
Oh actully relevant test failure re ```"message": "(0 , _sequelize.literal) is not a function"```

dannash100 (approved):
So sick! love it 💯

chris-bes (approved):

PR #4178, opened 2023-06-03
Containerisation fixes

passcod:
### Changes

Various fixes from finishing up the containers epic:

- Fix workflow when running on a branch (instead of on a PR)
  - Also simplify the normalisation/tag generation logic
  - Add a bunch of OCI-spec labels to the image (doesn't fix anything, it's just nice)
  - Fix docker caching to not conflict between central and facility
- Provisioning didn't tolerate missing sections
- Default facility name/code when not provided
- Allow arbitrary fields in user provision (used to set displayNames)
- Allow absolute pathing for reference data files

### DIFF
--- a/.github/workflows/cd-containers.yml
+++ b/.github/workflows/cd-containers.yml
@@ -28,9 +28,10 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - linux/amd64
-          # - linux/arm64
-          # it works, but it's just too slow
+          - arch: linux/amd64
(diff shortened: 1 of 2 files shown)


georgekamar (commented):
There's just two things I didnt understand:
- The removal on lines 71-83 of the workflow
- The cache-from/to scopes

Other than that it all looks good to me except for one thing which seems off

georgekamar (approved):

PR #4182, opened 2023-06-05
EPI-474: Set lastUpdated to time of materialisation

NavarroEmilioLuis:
### Changes

It was decided that it was more correct to update this field on every rematerialisation.


edmofro (commented):
Will leave the full review to @passcod but I have one question

passcod (approved):

PR #4190, opened 2023-06-07
EPI-474: Failing tests cleanup

NavarroEmilioLuis:
### Changes

Because of the change needed to the core work of the card, we don't have a way of knowing the exact timestamp of the bundles. The timestamps from the resources can be determined by reading their field, which is not great, but what's most important is the format.

I only realized this was happening now because sometimes the tests run so fast that the `lastUpdated` field is exactly the same as the previous `updatedAt` 😆


edmofro (approved):
lgtm!

PR #4193, opened 2023-06-07
No issue: Fix intermittent LAN test

NavarroEmilioLuis:
### Changes

~~It seems this test doesn't fail locally (what?).~~ Doesn't fail as much! So I'll be pushing some debug logs here for CI.

Finally got it to error locally somehow, this was the log:
```
$ NODE_ENV=test jest patientlocations
  console.log
    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@

      at Object.log (__tests__/apiv1/PatientLocations.test.js:38:13)
          at runMicrotasks (<anonymous>)

  console.log
    8 3

      at Object.log (__tests__/apiv1/PatientLocations.test.js:39:13)
          at runMicrotasks (<anonymous>)
```

As I suspected, there are more locations than we expect, so that just means we didn't clean properly.

I assume the reason it happens is that we probably create some locations randomly on the test setup, if these end up with a `maxOccupancy` of 1 by chance, then these tests get screwed up!

Scratch that, we always seed locations with that max occupancy, the thing is that they might grab the config facility ID, so that's the issue

### DIFF
--- a/packages/lan/__tests__/apiv1/PatientLocations.test.js
+++ b/packages/lan/__tests__/apiv1/PatientLocations.test.js
@@ -27,7 +27,6 @@ describe('PatientLocations', () => {
     models = ctx.models;
     app = await baseApp.asRole('practitioner');
     patient = await models.Patient.create(await createDummyPatient(models));
-    const locationTest = await models.Location.count({ where: { maxOccupancy: 1 } });
     locations = await models.Location.bulkCreate([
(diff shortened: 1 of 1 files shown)


dannash100 (approved):
Amazing its been quite bad atm ay the intermittent failure situ

PR #4194, opened 2023-06-07
TAN-2220: syncAllLabRequests shouldn't resync encounters that are already synced

edmofro:
### Changes

Currently, the `buildSyncFilter` for `Encounter` will sync an encounter attached to a lab request every time the lab request changes, even if that encounter is already in sync locally. 

This is causing a really bad bug in Aspen's environment, when the following happens:
- Create a lab request on encounter x to record some result back from the lab
- Sync kicks in, snapshotting that lab request to be pushed to the central server
- Discharge encounter x, recording it locally just after the push has started so it's not pushed yet to the central server
- Push finishes
- Pull starts, and detects there's a new lab request for encounter x on the central server, so pulls down the central server copy of encounter x _with no end_date_
- The encounter no longer appears as discharged on the facility server

While this PR does not account for every case where the encounter might already be in sync on the facility server (e.g. when it has already synced with an earlier lab request), it does fix the issue at hand in Ba Hospital where the encounters being reverted are all for patients that are marked for sync.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/shared-src/src/models/Encounter.js
+++ b/packages/shared-src/src/models/Encounter.js
@@ -225,6 +225,7 @@ export class Encounter extends Model {
           INNER JOIN lab_requests lr ON lr.encounter_id = e.id
           WHERE e.updated_at_sync_tick > :since
           OR lr.updated_at_sync_tick > :since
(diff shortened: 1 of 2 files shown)


chris-bes (approved):
LGTM!

PR #4207, opened 2023-06-08
EPI-513: Create new database/model for vital logs

NavarroEmilioLuis:
### Changes

I think most fields and choices are straightforward, the ones that probably need a bit of clarification are:

- Storing both previous and next values
  - The solution to determine both values based on a chain of records adds quite some complexity.
  - That same solution is not sync proofed since we would have to assume what value they edited.
  - This is a closer representation of what's being registered: a log.

- These can be null
  - Easier to explain, previous value could be non-existent, new value could be trying to nullify something.


stretchkennedy (approved):
LGTM 🎉

I think the double-storage of previous/next is one of those optimisation rabbit holes that probably aren't worth it, given that vitals aren't likely to change much.

PR #4208, opened 2023-06-09
SAV-279: Make field required

tcodling:
### Changes

We found another spot where the `required` prop was missing which causes weird behaviour with the new "x" clear field. This is essentially an extension of this [previous PR](https://github.com/beyondessential/tamanu/pull/4188/files) going through and adding required to fields as needed.


dannash100 (approved):
yes yes 🙏

PR #4215, opened 2023-06-11
Refuse merge if the merge-paused label is on

passcod:
### Changes

This adds a status check that fails if the label is on.

If we add this to the required checks it would actively stop those PRs from being merged.
(Incidentally it would make it possible to enable auto merge on merge-paused PRs.)

### Checklist

- [x] Code is finished


edmofro (approved):
cool!

PR #4216, opened 2023-06-12
NOTAM-60: Test migrations on Postgres 15

passcod:
### Changes

As it says on the tin.

(Incidentally I've been running 15 locally since it came out (oct/dec last year) so it's most likely not a big deal.)

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
I mean, it will run on all 3, right? No biggie

PR #4230, opened 2023-06-13
EPI-381: Program surveys visibility criteria on desktop not working

stretchkennedy:
### Changes

There are two related fixes here: 
1. Program surveys proper, where we pass `allComponents` down and use it to check visibility + perform calculations
2. Death and discharge forms which build on `PaginatedForm` and also use visibility criteria, where we map over all children of children to get a list of every React component then pass all the (React) components to `getVisibleQuestions`, where they're turned into quasi-`SurveyScreenComponent` objects

I'm really unsure about the design and flow of 2. I feel that grabbing data from the props of your children's children is complicated intergenerational coupling between React components, and is hard to think about and work on. However, I also feel fixing it should be out of scope for this card, and should also involve consultation with devs who've got more context on that area of code than I do.

### Checklist

- [x] Code is finished

_Upon merging:_

- [x] Update the changelog (look for the draft release [here](https://github.com/beyondessential/tamanu/releases, or create a new one if it doesn't exist following [the Slab doc](https://beyond-essential.slab.com/posts/merging-a-tamanu-ticket-okxp8s4s#h0y52-creating-a-draft-release)
- [x] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed


mcccclean (changes requested):
Marking my above comment as a change request so it shows up in the right filters

mcccclean (approved):
Oh yep, yeah you're quite right that the whole setup of it is pretty tangled (and I think we're very much on the same page about reading descendents' props); this PR is a good step forward in clearing that up. Great candidate for a maintenance task and yeah I think it'll be made much easier by this work. :+1::heavy_check_mark:

PR #4231, opened 2023-06-13
EPI-488: Make vitals editable on desktop

NavarroEmilioLuis:
### Changes

- Complete modal/form UI
- Include missing fields
- Add delete entry button


stretchkennedy (dismissed):
Mostly LGTM, one bit of discussion around naming (which isn't really to do with the PR, so I'm happy to stamp this if you want to make it out of scope 😄)

stretchkennedy (approved):

PR #4236, opened 2023-06-15
MDEV-153: Add currentFacilityId replacement parameter for reports

mcccclean:
### Changes

Adds a `:currentFacilityId` to the replacements scope for db driven reports to make running queries against the current facility a little more convenient.

(Also man it's a good thing I wrote tests - almost didn't bother! The facility server test setup wasn't setting LocalSystemFact and as we move to db config we're gonna want that!)

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


edmofro (commented):
The code to provide the parameter looks good, and thanks for the tests and fixup there. 

There's one thing I'm not sure on though. When we import a report, there is a validation step that makes sure all substitutions in a query have a corresponding entry in the report JSON. For start and end date, this isn't required as they can be provided by the system. Similarly with `currentFacilityId`, we shouldn't validate that there's a field providing that param, and if anything we should validate that it's _not_ included as an explicit field. Does this PR handle that implicitly, or is it missing?

edmofro (changes requested):
The one date thing to work out, otherwise all looking good

edmofro (approved):
lgtm!

PR #4237, opened 2023-06-15
EPI-526: Create lan endpoints

NavarroEmilioLuis:
### Changes

Was +164 -16 line changes before the unit tests!

This was quite tricky. Originally I planned on creating regular endpoints, however, after a closer look on how the vitals work I realized that:
- GET endpoint is not needed. We have to fetch the history just as we are retrieving the answers because the table needs to display whether they changed or not.
- POST endpoint is not needed. We will not be creating anything new, the log is just us keeping track of the survey response answer update.
- PUT endpoint was needed. This one is fairly straightforward but it does quite a lot of work. I also blocked from blindly receiving any type of survey response answer, instead only accepts answers from vitals surveys that are not calculated questions.


biaoli0 (commented):
Code looks great overall, great work! 
I just have a few general questions/suggestions:
1.  On the log, I am having a thought to have a survey response answer log. I feel the vital answer is the same as survey response answer, it could potentially get used by other survey but maybe not.
2.  The create date in a survey response is missing out and it causes complexity in vital. We always need to find the date from the survey response answer. If it is required by all other surveys, add a new column would be great to ease some pains. (out of scope)

biaoli0 (approved):
The new endpoint look great!

PR #4247, opened 2023-06-19
SAV-291: Created Imaging Request did not show up in Encounter's Imaging request tab, but appear again later

chris-bes:
### Changes
- Only apply search values from context when the IR table is not for Encounter

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


edmofro (approved):
There's so much going on in this file that switches behaviour depending on whether it's inside an encounter or not, that I think it should be two separate components. There could be a common exported const that has the columns and accessors. 

Happy with this as a quick fix, but could you follow up with a bit more of a refactor? If you don't have time now, add it as a maintenance item in a maintenance sprint

PR #4257, opened 2023-06-20
EPI-491: Editable vitals reason for change field

NavarroEmilioLuis:
### Changes

This field was already partially drafted, I'm just trying to make diffs smaller/easier to review. I don't think the field needs its own component because it is very small plus very specific.

- Field options are configurable (but have a default).
- Field required is false by default, but should have feature flag to make it required.

Config changes:
- New localisation entry `vitalEditReasons` that should be a list of options (with a default).
- New feature flag entry `mandatoryVitalEditReason` that specifies whether the field is required or not.


biaoli0 (approved):
LGTM!

PR #4261, opened 2023-06-20
TAN-2224: Add ability to admit to hospital directly from a clinic

Alastair-L:
### Changes

Card: https://linear.app/bes/issue/TAN-2224/add-ability-to-admit-to-hospital-directly-from-a-clinic-encounter

- Add new option to dropdown button for Encounter Actions

This seems a bit too easy, is there something I'm not thinking of? If not, congratulations to previous devs for some very easy-to-change code 🎉

### Checklist

- [x] Code is finished
- [x] UI Screenshots added to the Linear issue\

_Upon merging:_


NavarroEmilioLuis (commented):
The simplicity makes me question that, too. I think for everyone's sake we should check:

- Can't we just switch the 'Admit to hospital' position all the time? If not, that's fine.
- This isn't really equivalent because the previous workflow would also create a discharge, and I believe this simply changes the encounter type and creates a note.
- Does this mean we should actually be revisiting the logic of an "encounter progression moving forward"?

Hoping all responses are OK and this is good to go, pretty clean, nice work!

NavarroEmilioLuis (approved):
Happy to follow-up later, this ticket is not meant to solve that puzzle. However I'll say that it probably makes sense now cause you have the context, but give it a couple of months and it's pretty easy to miss this out - you'll have to trace the commits and maybe get a sense if you stumble with this PR. Which is fine for this, I just wanted to have confirmation it's actually the backend logic that we want and if we could compromise with the selectable options order 👍

PR #4270, opened 2023-06-21
Fix permission/ability helpers calling style

passcod:
### Changes

Use explicit models argument to allow late binding rather than relying on imports in shared which probably doesn't quite work as well anymore.

### Checklist

- [x] Code is finished
- [x] Tests are adjusted


NavarroEmilioLuis (approved):
Kinda weird this only causes problems in EB? Anyway, this looks fine and it's probably for the better anyway 👍

mcccclean (changes requested):
One tiny change

mcccclean (approved):

PR #4280, opened 2023-06-22
No issue: Fix various fhir processes

NavarroEmilioLuis:
### Changes

The recent shared situation change was a bit of a gamechanger, previously the way it worked, we always had the context the models get after being initialized. Now, anything that imports models directly inside shared will just be the plain objects so things aren't going so well!


stretchkennedy (approved):
LGTM 👍

On a deeper note, I'm left wondering:
1. How did this ever work? We're calling models prior to initialisation all over the place
2. What actually broke it? I'm not 100% sure on this either
3. Is there some way to make it work that way again and save all this effort?

PR #4281, opened 2023-06-23
NASS-583: patient lab results table fixes consolidate

tcodling:
### Changes

 - Should remember the last selected category/panel when navigating away from results tab
 - Align result column values with headers
 - Remove pagination logic from table. (should just show as a list with no options)
 - Wrong department showing on lab request view
 - ~~Date column header should be 2 lines~~ (this appears to be fixed already)
 - ~~Export currently breaks the styling of the whole app until reset~~ (being addressed in another PR)


mcccclean (changes requested):
Couple changes, but looking pretty good!

mcccclean (changes requested):
Almost there!

mcccclean (approved):

PR #4299, opened 2023-06-27
Fixes for release workflows

passcod:
### Changes

- Use a PAT instead of the GHA token in more places where permission bypass is needed (ie to push to protected branches)

### Checklist

- [x] Code is finished


stretchkennedy (approved):
LGTM 🎉 

I'm assuming the PAT's got the right permissions and has been set, etc.

PR #4300, opened 2023-06-27
Update server list aspen

nick-beyondessential:
- update host address on meta-server list for aspen clone and demo (moved to new account)
- fix https on fiji clone


NavarroEmilioLuis (approved):
LGTM

PR #4305, opened 2023-06-27
TAV-660: Build desktop apps in container

passcod:
### Changes

- Package all variants at once:
  - Linux (AppImage and DEB)
  - Mac (tar only, can't make dmg on Linux)
  - Windows:
    - `system`: normal NSIS-based .exe that installs to Program Files
    - `appdata`: NSIS-based .exe that installs to AppData
    - `msi`: MSI installer (without auto-update)
    - `unelevated`: NSIS-based .exe that installs to Program Files without UAC prompt
- Do all that in a container, so it's easy to run/reproduce in GHA and locally
- GHA Desktop build is now on Linux, so we avoid the macOS tax (10x price!)
- Split the CD workflows, which also lets us build the desktop client all the time, not just for auto-deploys
- Doesn't upload artifacts to Github anymore, as they were getting too large, and leaves them in S3 but fixes the folder structure.
  - Posts a comment on the PR with links when done, and even indicates when there's a build ongoing 
image

Idea for production/release builds is that we'll package everything on release, and then have a separate workflow/script responsible for putting the right files in the right place for a particular deployment. That lets us control precisely when the desktop update is made available during a deployment upgrade, or provide the files before the upgrade even starts in the case of MSI distributions.

That part will be done in another PR because this one is already Quite A Thing.

Also yes Hopefully™ we Soon™ won't even _have_ desktop builds anymore but Until Then We Might As Well Be Comfortable.

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/branch-info.yml
+++ b/.github/workflows/branch-info.yml
@@ -1,0 +1,31 @@
+name: Branch info
+
+on:
+  workflow_call:
+    outputs:
+      branch:
+        value: ${{ jobs.run.outputs.branch }}
+      actual_ref:
+        value: ${{ github.event.pull_request.head.ref || github.ref }}
+
+jobs:
+  run:
+    name: Branch name
+    runs-on: ubuntu-latest
+    steps:
+      - name: Normalise branch/ref name
+        uses: actions/github-script@v6
+        id: ref
+        env:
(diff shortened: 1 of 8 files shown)


stretchkennedy (approved):
LGTM, other than one optional nitpick about the regex

I didn't realise the client build was on osx 😂

PR #4309, opened 2023-06-28
EPI-520: Fix ID card DOB display

NavarroEmilioLuis:
### Changes

Given that DOB are date strings, we cannot rely on JS to parse those as we want because per spec, they are considered in UTC instead of local time, which is a bit messy.

There are other places were this happens but I'm creating a new linear card for tracking that, cause it affects more places.


tcodling (approved):
lgtm :)

PR #4310, opened 2023-06-28
EPI-541: Fix usage of date helper functions

NavarroEmilioLuis:
### Changes

The point of this is to have a controlled API. I originally didn't want to export any of the helper functions but tried three different designs that all felt a bit awkward:
- Creating more methods for `DateDisplay`: No flexibility plus maybe a bit WET.
- Making the second argument a string, then mapping that string to the helper function: Again no flexibility, forces to know what string to use (cause using a constant to overcome this makes things pretty much the same, you have to export something other than the component).
- Refactor the component to be purely presentational and abstracting the raw formatted value into its own function, that the component would expose itself: doesn't cover all types of formats currently being used, least flexibility among all, super verbose API and not straightforward logic.

I think that this approach gives you the best dev experience. The only thing you need to know is that you always use `DateDisplay` or `DateDisplay.rawFormat` and that's it, if you need another format, just pass as argument 👍


stretchkennedy (approved):
Two optional suggestions; otherwise, LGTM 🎉

stretchkennedy (approved):
LGTM (one suggested comment to explain why the error exists) 👍

PR #4317, opened 2023-06-29
Replace tailscale action with official one

passcod:
### Changes

Cleanup now that the features we wanted were merged upstream.

### Checklist

- [x] Code is finished
- [x] Testing: starting an auto-deploy. If it works, this ticket works too!


edmofro (approved):
lgtm

PR #4321, opened 2023-06-30
NASS-583: Remove category column when sorting by it

tcodling:
### Changes

This is pretty scary looking but is almost all formatting. Have highlighted my changes in my two comments :)

Here we just add a conditional check to not show the category column when sorting by category as they all just show the same value so is not needed

_Upon merging:_


stretchkennedy (approved):
LGTM! Github's hide whitespace feature was very useful 😄

PR #4322, opened 2023-06-30
Fix the CD Down workflow

passcod:
### Changes

In a recent PR I made a `branch-info` shared workflow which provides some branch information, such as a normalised name for the CD auto deploys, and version information for the CD client builder.

The normalised branch name is obtained directly from the github information, and the version information is obtained by checking out the branch and reading the package.json.

However, what I failed to notice is that when a PR is merged or closed, its branch gets deleted, and the version part of the workflow fails because it can't check out the branch anymore. Thus, the CD Down workflow fails and auto deploys are no longer deleted when a PR is closed or merged.

~~This PR defaults the version workflow to checkout the `main` branch if it can't checkout the actual branch. This is a bit weird and may result in wrong version info, but only for workflows running on a deleted branch, which at the moment is only CD Down, and CD Down doesn't use the version info.~~

This PR defaults the version to the empty string if it can't checkout the branch.

---

Also brings in a drive-by fix for overzealous concurrency control for building the desktop client. That build should not interfere with the CD Up/Down operations, nor vice versa.

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/branch-info.yml
+++ b/.github/workflows/branch-info.yml
@@ -39,7 +39,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.pull_request.head.ref || github.ref }}
+          ref: ${{ github.event.pull_request.head.ref || github.ref || 'main' }}
       - uses: actions/github-script@v6
         id: version
         with:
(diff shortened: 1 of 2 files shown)


stretchkennedy (approved):
Stamped.

I've got some misgivings about whether this is correct behaviour if a PR was closed rather than merged, but as you note, it doesn't matter when we're removing the CD environment.

PR #4354, opened 2023-07-05
TAMOC-22: Lab notes test

mcccclean:
### Changes

A bug was introduced to release/1.27 by a bad merge, so there's no actual logic fix to bring back to main. However I added a test to catch the misbehaviour introduced by the bug, and that should be merged back into main.

### Checklist

- [x] Code is finished
- [x] Tests are written


dannash100 (approved):
So sorry this was my bad 😖

PR #4364, opened 2023-07-06
EPI-545: Prevent users sending login requests by hammering button

stretchkennedy:
### Changes

Because dispatch(login(…)) isn't awaited on the login screen, users can hammer the login button multiple times and fire off lots of requests. There's also no feedback when they click the button, meaning they're more likely to hammer it, either out of frustration or confusion.

This is not great, because /login  is a very expensive route with no rate limiting and a synchronous call to bcrypt, and facility servers are single-process, so a user spamming the button enough times can lock up the entire facility server for every other user.

### Checklist

- [x] Code is finished

_Upon merging:_


dannash100 (approved):
Beautiful stuff, I agree - this is very likely a cause of issue, expecially if the endpoint is unexpectedly slow or any number of other reasons to hammer a button 😆

mcccclean (approved):

PR #4369, opened 2023-07-07
Fix selected date issue in different timezone

biaoli0:
### Changes

[Issue 2 in this comment](https://linear.app/bes/issue/EPI-494#comment-41f4e3df), which is that in different timezone, if you select '2023/07/06' as a date string, when it is parsed into Date it could become '2023/07/05 02:00 GMT -4:00' and its 'startOfDay' is '07/05 00:00:00' but not '07/06 00:00:00'.

The fix is to use .parse() from 'date-fns' as it has options to handle it.  

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/Charts/components/DateTimeSelector.js
+++ b/packages/desktop/app/components/Charts/components/DateTimeSelector.js
@@ -1,7 +1,7 @@
 import React, { useEffect, useCallback, useState } from 'react';
 import styled from 'styled-components';
 import { debounce } from 'lodash';
-import { addDays, format, startOfDay } from 'date-fns';
+import { addDays, format, parse, startOfDay } from 'date-fns';
 
 import { DateInput as DateInputComponent, SelectInput as SelectInputComponent } from '../../Field';
 import { Y_AXIS_WIDTH } from '../constants';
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (changes requested):
I'm having a bit of a hard time with the overall logic, mainly the need to control the `DateInput` value - then, having the need to convert it back and forth from a date object to a date string.

As a general rule of thumb, we should always be using date strings or date time strings where possible, only converting to/from JS dates as little as possible. Here the issue is caused because the Date constructor parses the date strings as UTC instead of the expected local time. For fixing this is easier (and tested) to use `parseISO`, which should make things work provided no extra conversions are done elsewhere.

Probably using `parse` has a similar (or even the same) behavior but it's a bit clunky since we specify a format there, plus fill in the gaps with a `new Date()` call.

NavarroEmilioLuis (approved):
Looks good. I had a chat with Biao and there is not much we can do to avoid parsing the date string again since the underlying `DateInput` breaks if that doesn't happen - unfortunately we don't use it anywhere else directly.

This will be looked into a bit further but pre-approving cause it would just be a minor change if testing goes well.

PR #4370, opened 2023-07-07
NASS-637: Default lab request form type determined before field available

dannash100:
### Changes

This gets around an issue where its not always safe to default form select type to panels, or individual (and in future superset) as the selection is dynamic based on a few factors
- Move the lab request form type logic to own file 
- skeletonized loading state see video (note load times exaggerated in local-dev)
- As the value is required field you won't be able to click through until it has loaded and populated default option.
- I began building this into Radio logic - but currently we determine the design of the radio by weither its options have description which aren't available at this point. A full solution for loading state skeletons for radio is out of scope for this 1.27 ticket in my opinion

Skeleton loading state for this radio set
https://github.com/beyondessential/tamanu/assets/38335330/73e7fa35-d064-43c7-b18c-d73812dac8f1

### Checklist
- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/forms/LabRequestForm/LabRequestFormScreen1.js
+++ b/packages/desktop/app/forms/LabRequestForm/LabRequestFormScreen1.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { useQuery } from '@tanstack/react-query';
 import * as yup from 'yup';
 import { LAB_REQUEST_FORM_TYPES } from '@tamanu/shared/constants/labs';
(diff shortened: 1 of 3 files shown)


stretchkennedy (changes requested):
Few nits, otherwise looks good 👍

stretchkennedy (changes requested):
One minor thing but otherwise looking good

stretchkennedy (approved):
🎉

PR #4374, opened 2023-07-07
EPI-488: Last bits

NavarroEmilioLuis:
### Changes

These are fixes trying to resolve some differences noted at testing. Most stuff is explained per commit but the overall biggest change is that now all entries are supposed to have a log, not just further edits.


biaoli0 (changes requested):
Nice work, I got a few questions here and there, plus the migration should be updated to a subcommand.

biaoli0 (approved):
Thanks for moving the isVitalSurvey for me :D Looks great to me overall.

nick-beyondessential (approved):

PR #4375, opened 2023-07-07
NASS-751: Planned move patient encounter info pane incorrectly displaying

dannash100:
### Changes

- Bring header outside of nth-child logic of determining the divider that only applies when there are two elements in row.
- Remove context and replace with selectors from components `toString` or some magic im not sure see [styled-components: Advanced Usage](https://styled-components.com/docs/advanced#referring-to-other-components) Thank u heaps @tcaiger 

Sadly our chrome version (electrons one) doesn't quite support `nth-child(odd of .class)` yet or else I could have avoided the header prop by excluding the header component from the odd counting. Discussed with tom c that this was cleaner without the ability to use that exact syntax right now.

Screenshot 2023-07-07 at 4 39 28 PM

### Checklist
- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/InfoCard.js
+++ b/packages/desktop/app/components/InfoCard.js
@@ -80,9 +80,9 @@ export const InfoCardItem = ({ label, value, ...props }) => (
   </CardCell>
 );
 
-export const InfoCard = ({ HeaderComponent = null, children, elevated, inlineValues }) => (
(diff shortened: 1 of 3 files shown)


mcccclean (changes requested):
Lil fix :pray:

mcccclean (approved):
LGTM

PR #4380, opened 2023-07-09
NASS-810: Vitals table export missing button and spreadsheet cell fix

dannash100:
### Changes
Vitals table export button disapeared due to logic determining the hiding of footer when no results.
- Remove default of 0 from count as not always appropriate to table without pagination
  - Another option here would be to just pass in data.length i guess 🤔 but kinda weird 
- Capatilize result cells with css and not lodash
- Fix prop name on exportName (it was not being passed through)
- Make a version of cell that is designed for export and doesn't render tooltips and do range validation logic
  - This was done also because the output export had loads of `-` in empty cells which could cause trouble i feel.
  
 Its back :)
Screenshot 2023-07-10 at 11 28 09 AM

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/FormattedTableCell.js
+++ b/packages/desktop/app/components/FormattedTableCell.js
@@ -1,4 +1,4 @@
-import { capitalize, isNumber } from 'lodash';
+import { isNumber } from 'lodash';
 import React, { useMemo } from 'react';
 import styled from 'styled-components';
 import { Colors } from '../constants';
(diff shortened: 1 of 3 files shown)


NavarroEmilioLuis (changes requested):
I think this all looks good, but can we just allow exports on vitals table and then get this more proper stuff pointed at 1.28 instead? This increases the amount of stuff that needs testing/can potentially get ignored.

Also CSS capitalize is not equivalent to Lodash's capitalize. The first will be applied on every word while the second one is applied on just the first letter of the string.

mcccclean (changes requested):

PR #4396, opened 2023-07-12
EPI-552: Create feature flag that disables editing vitals

NavarroEmilioLuis:
### Changes

This feature flag is needed for us to consolidate the initial vital log creation, not letting users edit vitals until we switch it on


biaoli0 (changes requested):
Look good to me overall! I am just being picky on the unit tests.

biaoli0 (approved):
LGTM!

PR #4408, opened 2023-07-13
EPI-488: Fix blank AVPU and same value submission

NavarroEmilioLuis:
### Changes

Shees what a diff. The only thing left to do here is to write up those unit tests


stretchkennedy (changes requested):
One edge case bug that I'd like to get fixed unless it's a ton of work, one very minor nit, otherwise looks good

stretchkennedy (approved):

PR #4416, opened 2023-07-17
Add safeguard for invalid inputs for publish workflow

passcod:
### Changes

I don't see any way that there's anything _wrong_ with the scripts, and I don't see any way for it to exhibit [the behaviour observed](https://github.com/beyondessential/tamanu/actions/runs/5551360025) under normal usage, so it must have accidentally been called with a `version` input of `release/`, probably as the caller was trying to search for the correct `release/` branch.

This patch doesn't change any behaviour as there's none to, but it adds a check that the version input is either one of the special values `major`, `minor`, `patch`, or an explicit x.y.z version.

It also manually bumps the version to what it should have been (1.28.1).

### Checklist

- [x] Code is finished

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tamanu",
-  "version": "release/",
+  "version": "1.28.1",
   "description": "This repo contains all the packages for Tamanu",
   "main": "index.js",
   "homepage": "https://github.com/beyondessential/tamanu.git#readme",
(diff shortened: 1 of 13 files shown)


NavarroEmilioLuis (commented):
Hmmmm but when we saw this, it actually happened on `release/1.27` (you can see that's still the version in that branch). This happened twice so it's definitely something that is caused by the script (well not THE script, but A script/action whatever) unless we failed the same twice, which is plausible but not as likely?

NavarroEmilioLuis (approved):
On the other hand, will approve so we unblock `release/1.28` branch because we will need to do this bump anyway, plus the extra safety might give us more diagnostics

PR #4419, opened 2023-07-18
EPI-559: Fix delete entry button position

NavarroEmilioLuis:
### Changes

A bit messy to fixate to the top given the hardcoded offset but 🤷 adding an empty label on the top seems hackier, plus it would need to have some text and set to `visibility: hidden`.

Pictures in Linear!


stretchkennedy (approved):
LGTM 👍 

It'd _maybe_ be nicer HTML to put the box inside a div with the field or account for empty labels in some way, but I think it'd make the actual components worse for little benefit

PR #4421, opened 2023-07-18
EPI-560 Add scrollable view to Table

biaoli0:
### Changes
Ignore the white space please, only add `<ScrollView>` at line 45 and it's close tag.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/mobile/App/ui/components/Table/index.tsx
+++ b/packages/mobile/App/ui/components/Table/index.tsx
@@ -34,41 +34,39 @@ interface TableProps {
 }
 
 export const Table = ({
-         Title = null,
-         rows,
-         columns,
-         cells,
-         tableHeader = null,
-         onPressItem,
-         scrollHandler,
-       }: TableProps): JSX.Element => (
-         <ScrollView>
-           <RowView>
-             <StyledView>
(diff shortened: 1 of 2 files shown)


NavarroEmilioLuis (changes requested):
At first I was a bit skeptical about modifying the table component here, it seems this would not break the vaccine table though I'm unsure if it's because it already displays well and the `ScrollView` added here is never actually used there.

Regardless if that's true or not, this approach fails because of the hacky display of the vitals table that try to fill in the screen. If you check this out with a patient that has no vitals recorded or just one/two (so there are no enough columns to cover the screen horizontally) you'll see that it only scrolls the populated columns.

I would rather play this change safer and change the [`StyledView` here](https://github.com/beyondessential/tamanu/blob/main/packages/mobile/App/ui/components/VitalsTable/index.tsx#L95) for a `ScrollView` - or maybe even `StyledScrollView` (will need to change the closing HTML tag and import it, too).

NavarroEmilioLuis (approved):
Yeah it was very weird, thanks for grabbing this in such short notice

PR #4422, opened 2023-07-19
Update ECR Push AWS role for CD Up

passcod:
### Changes

I added a new role but that shouldn't have broken the old one... let's try it anyway.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
❓❔🤔❔❓

PR #4427, opened 2023-07-20
NOTAM-66: Final installer setup for Aspen

passcod:
### Changes

- Remove experimental NSIS and Zip builds
- Remove Linux deb build (AppImage is fine on all distros)
- Only build MSI in Github Actions, leave it out of package.json so it gets ignored in Codeship
- Only build to release bucket on actual releases (tags), not on everything in the release/ branch
- Allow selecting which bucket to upload to for manual dispatches
- Add separate build specifically for Aspen's options:
  - Versioned desktop shortcut
  - Installs into a different path in Program Files
  - Different product ID so it doesn't conflict with the NSIS version

### Checklist

- [x] Code is finished


edmofro (commented):
Couple of questions to clarify my understanding of minor points, but overall looks good. 

I wonder though whether we're better to always produce an Aspen-shaped build alongside the rest, rather than having to run a separate release workflow?

edmofro (approved):
Love it, thanks :-)

PR #4429, opened 2023-07-21
TAV-724: Hospital Simulator

passcod:
### Concept

This simulator's metaphor is that we're playing the board game called Hospital. ~~No, I'm only a little insane.~~ (We can keep the structure and rename all of this stuff if we don't care about the theme, too.)

- there are well-defined **Activities** that players can do
- **Players** have a "class", which defines what activities they do
- Players have a **routine**, which is a set of activities they always do, an **inbox**, which is a set of activities they're given by other players to do, and an outbox, which is where they put activities to send to other players during their turn
- A **Game** has a win condition or we can call it a **stop condition**
- At setup, a game allocates N players of each class
- Every round:
  1. All players play their turn simultaneously
  2. The game collects their outboxes
  3. The game distributes activities to inboxes
  4. The stop condition is checked

Activities are defined by calling the (internal) Tamanu APIs. To provide structure, they have three phases:
1. (optional) Gather, where information about the action and/or the world is collected
2. (mandatory) Act, the core action
3. (optional) Call, which adds to the player's outbox

This is all engineered to make adding to the simulation as straightforward as possible. The framework should remain fairly static, and then:
- developers add more Activities and set them to be done by classes of Players
- users define a Game which uses a set of players and defines a stop condition, then play a run of that game

Because the simulation runs against Tamanu APIs and doesn't use the database directly (so far), it can be run against any random (dedicated) Tamanu cluster without needing special access or any credentials beyond a normal Tamanu admin.

This PR defines the framework and:
- 2 player classes (Nurse and Receptionist)
- 3 activities
  - intake (creating patients)
  - admit patient (to hospital or clinic)
  - fill survey (used for entering vitals)

In local tests, with a single server each of central and facility, I ran games performing up to 10000 API requests per minute, which is pretty neat. Oh yeah, this whole thing pulls double duty as a way to generate data *and* stress-test the servers!

Next phase after this is: https://linear.app/bes/issue/TAV-765/add-vaccination-activities-to-hospital-simulator

### Other changes

- Adds a new admin [user management route](https://github.com/beyondessential/tamanu/pull/4429/files#diff-25924815cc90770a4144259f824d22b8c39ee027a348beadba832391f571421e)

### Checklist

- [x] Code is finished


mcccclean (changes requested):
This rules! A lot of fun to review & it's going to be so useful.

I've left a couple notes on line-by-line things but I've also got some architecture level notes that should probably be addressed first.

I think there's a bit of a tangle happening here that I'm struggling to get my head around (maybe I've misread it?) But it feels like the relationship between an Activity and a Player is a bit unclear, where it's hard to tell which of the pair is driving the operation. First the Players populate an outbox of calls, and then those calls are redistributed by the Game to... other Players, who then enact it?

It seems that there's two levels of decisionmaking that have to happen - at a coordinated faction level (all Nurses follow behaviour X, all Doctors follow behaviour Y) and at an agent level (one agent performing one action), and Player is kind of doing both duties. To use the boardgame analogy: I'm not sure whether a Player represents the human standing at the table or the pawn on the board - fine when it's a Monopoly but I think the hospital is more of a chess.

I'd recommend making the distinction explicit, so that the Player class is solely responsible for queuing up a bunch of actions and assigning them to one/many Pawns (or Agents? if we want to go more videogame-AI-ish?) which are then responsible for actually carrying them out. 

A Player's relationship to User is also a bit of a source of confusion for me, though honestly I think that mostly goes away with a Pawn/Player split; it'd be fine to have a mix of (Pawns with unique Users) and (Pawns who share their User with other Pawns of the same Player) depending on the requirements of the particular actions being taken.

PR #4430, opened 2023-07-21
EPI-574: Fine tune vitals

NavarroEmilioLuis:
### Changes

Five commits, five fixes:
- Change the copy in the table legend
- Change copy on no data graph view
- The tooltip on blood pressure only showed results for SBP, with this change it now shows both places.
- Fix intermittent vital table error. This happened because by the time the table was rendering there was no `visualisationConfigs` to match. This value will get updated accordingly once the state settles in.
- The lower boundary on arrow vectors went beyond the max lower range. This is just clamping that value to be a little bit better.


biaoli0 (changes requested):
Got a suggestion on reusing some logic for getting the measure data. Others are looking good!

biaoli0 (approved):

PR #4439, opened 2023-07-24
NASS-797: Remove exact displayID search from lab search bar

tcodling:
### Changes

We have a component called `DisplayIdSearchField` that was used for optionally enforcing a full string comparison when searching for NHN. We are phasing this field out and there are two left that aren't even functional (labs sidebar and covid patients sidebar). This card is just replacing the component (which isn't working on the form right now) with a basic string search field + another one on the covid patient register. 

I have also removed the staticValues logic from the few tables it is present as these are leftovers from places where the `DisplayIdSearchField` was previously and it currently does nothing (no logic on backend to handle the static values that I can find).

Am also wondering if maybe I should just delete the whole component `DisplayIdSearchField` since it isn't used anymore but not sure if that would be jumping the gun. Would be interested to hear what reviewer thinks 🤔 

Update: After my review i did end up deleting `DisplayIdSearchField` and updating all fields where this was used to use the same "Search NHN" format for the placeholder text

### Checklist

- [x] Code is finished

_Upon merging:_


edmofro (approved):
Definitely think you should remove the component. It will always exist in git history if we want to dredge it up!

All lgtm, with one question that I'm sure is fine :-)

edmofro (approved):

edmofro (approved):
Nice!

edmofro (approved):

PR #4443, opened 2023-07-24
EPI-536 Add age to patient summary screen on desktop

biaoli0:
### Issues:
1. [Issue here](https://linear.app/bes/issue/EPI-536#comment-8864f9a1) are Timezone related, which is that in GMT-0600, if birth date is '2023/07/06' as a date string, it will parse into Date and become '2023/07/05 18:00 GMT -6:00'.
2. Set system timezone to GMT-6, unit tests will capture these issues, however cannot dynamically set timezone in an easy way in Jest, because node always looks at system environments. Can run with 'yarn TZ=GMT-6 NODE_ENV=test jest' but it will set TZ globally. Will chat about it on dev chat.

### Changes:
1. Use parseDate() instead of new Date()

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/shared/__tests__/utils/date.test.js
+++ b/packages/shared/__tests__/utils/date.test.js
@@ -1,8 +1,9 @@
+import { format } from 'date-fns';
 import { getDisplayAge } from '../../src/utils/date';
 
 describe('date', () => {
   describe('getDisplayAge', () => {
-    const now = new Date('2023-07-11');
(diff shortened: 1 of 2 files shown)


tcodling (approved):
LGTM! Like i said just watch out for `new Date()` in the future :)

PR #4444, opened 2023-07-24
SAV-326: Add 'useConfig' and 'useFeatureFlag' hooks to desktop

georgekamar:
### Changes

- Add `useConfig('key')` hook to replace `useLocalisation().getLocalisation('key')`
- Add `useFeatureFlag('key')` hook to replace `useLocalisation().getLocalisation('features.key')`
- Replaced all instances of 2nd point with `useFeatureFlag`
- Replaced some instances of `getLocalisation` with `useConfig`
- Refactor `getLocalisation` function
- Add `const [val1, val2, val3, etc.] = getLocalisation([key1, key2, key3, etc.])` functionality, also implemented in `useFeatureFlag` and `useConfig`

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/components/Field/DisplayIdField.js
+++ b/packages/desktop/app/components/Field/DisplayIdField.js
@@ -1,15 +1,16 @@
 import React from 'react';
 import { useFormikContext } from 'formik';
 
-import { useLocalisation } from '../../contexts/Localisation';
+import { useConfig } from '../../contexts/Localisation';
 import { TextField } from './TextField';
 import { LocalisedField } from './LocalisedField';
 
 export const DisplayIdField = ({ name = 'displayId', required }) => {
   const { initialValues } = useFormikContext();
-  const { getLocalisation } = useLocalisation();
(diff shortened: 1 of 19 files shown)


chris-bes (changes requested):
Hey @georgekamar , thanks for the amazing effort! But unfortunately I think this is probably not what we want. The ultimate goal is to have a unified way to get feature flag and config. I think from the tech design, we agreed that for feature flag and config, we will start migrate them to be in Settings table instead (that includes the configurable note types that you were working on before). And for the front end, it will fetch it from a generic settings endpoint.

I remember McLean has attached the tech design notes in one of the tickets (can't remember which one). But here it is: https://beyond-essential.slab.com/posts/consistently-accessing-localisation-values-6ddg20jb

He was also working on https://linear.app/bes/issue/SAV-325/extend-setting-with-more-dev-friendly-functions before away for leave. I think maybe it's worth to park this until he's back coz he might have already got some WIP work for the backend at least

PR #4447, opened 2023-07-24
NASS-737: Subcommands to populate db-defined config default settings

dannash100:
### Changes 

**UPDATE**: on paring with Edwin we decided to change angel a bit and handle defaults outside of database, and join that logic with a larger schema for settings, broken into global defaults, facility defaults and central defaults. This gets around an issue with facility config and central config name clashes - but different internal schemas. I.e see sync key. 

This also gets around needlessly storing what are very much unchanging (without codechange) data across each facility in the db. 

Another reason for change was the difficultly it presented when adding new settings - basically requiring a migration to seed this new config and default

With these considerations I was able to move back to a single migration that uses default schema to determine which local settings to migration, while skipping sensitive secrets and credentials - as well as technical things that require restart inherently.
Important considerations when presenting this config through admin panel

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


passcod (commented):
Is this using the same config file as the config module? I'd think we want to have a split config for defaults, as there'll always be a need for an actual config file for like db creds and such.

This is probably another PR/card but also there should be some kind of validation schema?

Also there's a few existing settings eg fhir and integrations that should be migrated to this split default/value format.

PR #4448, opened 2023-07-25
EPI-579: Create VitalLog in mobile

NavarroEmilioLuis:
### Changes

The missing piece of the puzzle. This lets us create initial vital log entry when submitting vitals in mobile


biaoli0 (commented):
Do we need an initial log entry for all existing vital answers?

biaoli0 (approved):
LGTM!

PR #4476, opened 2023-07-28
EPI-584: Fully remove queries on logout, don't just invalidate them

stretchkennedy:
### Changes

Resets queries rather than invalidating them on logout, meaning that the underlying cache results are fully cleared and not just marked stale.

In EPI-584 there are two things going on: SAV-280, which we don't care about; and discharges being shown to users who shouldn't be able to see them, which we _do_ care about quite a lot. I believe this is happening because we invalidate queries on logout, which _leaves them in the cache_ while they refetch...but if they fail to refetch, probably just shows them.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/contexts/Auth.js
+++ b/packages/desktop/app/contexts/Auth.js
@@ -22,7 +22,7 @@ export const useAuth = () => {
     })),
     onLogout: () => {
       dispatch(logout());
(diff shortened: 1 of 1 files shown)


tcaiger (approved):
👍

PR #4477, opened 2023-07-28
TAMOC-35: Clarify and fix version compatibilty checks

passcod:
### Changes

- Parse the version and reformat it so that, in general:
  - MIN_CLIENT_VERSION is always x.y.0
  - MAX_CLIENT_VERSION is always x.y.999
    - the patch version of MAX_CLIENT_VERSION doesn't matter, as we always allow higher patch versions to connect. nevertheless, i think having it as .999 makes the intent clearer.
- Setup a simple way to add an override if one is ever needed.
- Improve the comments to make all this clear.

### Checklist

- [x] Code is finished


edmofro (approved):
lgtm

PR #4479, opened 2023-07-28
Fix initial vital subcommand

NavarroEmilioLuis:
### Changes

Self explanatory


stretchkennedy (approved):
LGTM

Had some offline discussion about getting rid of batching, but even with batching this is an improvement

stretchkennedy (approved):

PR #4487, opened 2023-07-31
EPI-586: Fix initial vital logs subcommand

NavarroEmilioLuis:
### Changes

Just the draft. Local testing looks fine, doesn't generate duplicates but will test it out further and will also see if it impacts the performance we had.

This is EPI-586.


dannash100 (approved):
Looks good 😇

PR #4489, opened 2023-08-01
Allow more than 255 chars for content

Alastair-L:
### Changes

I completely got swapped around on what TEXT and STRING meant. This fixes that.

Can I confirm that it's ok to edit the migration directly because the epic branch hasn't been merged to main yet?

### Checklist

- [x] Code is finished

### DIFF
--- a/packages/shared/src/migrations/1679271355192-addPatientLetterTemplateTable.js
+++ b/packages/shared/src/migrations/1679271355192-addPatientLetterTemplateTable.js
@@ -34,7 +34,7 @@ export async function up(query) {
       type: Sequelize.TEXT,
     },
     visibility_status: {
-      type: Sequelize.STRING,
+      type: Sequelize.TEXT,
       defaultValue: VISIBILITY_STATUSES.CURRENT,
     },
     created_by_id: {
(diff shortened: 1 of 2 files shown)


NavarroEmilioLuis (changes requested):
Hey so this looks good. My only suggestion (not a hard one though but I'll try to make a case for it) is not replacing `visibilityStatus` to string because:

- Apparently using `STRING` [varchar(255)] or `TEXT` doesn't matter to postgres behind the scenes, see https://stackoverflow.com/a/4849030
- This is a column shared by many models, while I know it's not unanimous, the majority seem to favor `TEXT`.
- There is not 2 without 3.

Feel free to push back if you're not convinced. Also yes you're right, given the work is on an epic branch this would be the only acceptable way of editing a migration file. However be mindful that:
- If the epic branch is already deployed this could lead to problems there.
- In your local setup there is also room for troubles because of this (think about not having issues now, then for some reason in 4 months you get weird behavior because it never got applied).

So make sure you check both! There is also a way of circumventing this issue by migrating down then up one step or adding the changes directly through SQL at both instances. Best of luck on both!

NavarroEmilioLuis (approved):
Beautiful!

PR #4490, opened 2023-08-01
Add ability to print patient letters

Alastair-L:
### Changes

I have no idea when this got lost 😬. Unfortunately I wasn't able to test the actual print functionality as it's difficult to pipe through wsl.

### Checklist

- [x] Code is finished

image


NavarroEmilioLuis (approved):
Looks fine because we only have two document sources anyway. I'm not sure if that will grow eventually but it's fine to leave like this for now.

General suggestion though, when working on epics, try including the card name somewhere so PRs are forever linked in Linear but also to be able to get more info possibly!

PR #4505, opened 2023-08-03
Update meta-dev server list

passcod:
### Changes

Rollup of changes from meta prod's list


edmofro (changes requested):
There's a linting error saying it wants one of the longer entries to be across multiple lines. I actually prefer the consistency in a file like this, but we'd better keep the linter happy

edmofro (approved):

PR #4509, opened 2023-08-04
NASS-831: Make sure important sync considerations can't be forgotten

edmofro:
### Changes

When introducing a new model, there are a couple of mistakes that are really easy to make with the current code:
1. Forgetting to add a buildSyncFilter function, so that all records try to sync to every facility even if they should be linked to patients that are marked for sync
2. Adding a buildSyncFilter that is not patient-specific (e.g. for settings), but forgetting to add that to the list of exceptions in getPatientLinkedModels

The changes in this PR means it is virtually impossible for a dev to make either of those mistakes:
- Split out `buildSyncFilter` into `buildPatientSyncFilter` for models that are linked to patients and sync based on which patients are marked for sync in the facility, and `buildSyncFilter` for other models that have sync filtering per facility (at the moment that's just `PatientFacility` and `Setting`). This means that `getPatientLinkedModels` does not need to have any exceptions built into it - it's just any model that has `buildPatientSyncFilter` defined
- Every model that syncs either `PULL_FROM_CENTRAL` or `BIDIRECTIONAL` must implement one of those two sync filter functions, or it will throw an error during snapshotting on central

All of the other changes in here are just renaming things or adding a null returning `buildSyncFilter` on all models that sync to facilities and didn't already have one. 

### Checklist

- [x] Code is finished
- [x] Tests are written (prior tests should catch regressions 🙌)

_Upon merging:_


chris-bes (commented):
Code LGTM! I only have 1 suggestion. There's also some conflicts

chris-bes (approved):
LGTM!

PR #4513, opened 2023-08-04
TAN-2137: Remove template validation

tcodling:
### Changes

Missed some validation tweaks on my last pr. Template needed was still being validated even without the required star and  the name didn't have a required star even though it was required for the form to submit properly. We also made the body required per comment from Megan


chris-bes (approved):
LGTM!

PR #4525, opened 2023-08-07
EPI-595: Refactor code around updatedAtByField

NavarroEmilioLuis:
### Changes

This refactor happened in #4503 but the core work in that PR was scrapped. This is just trying to save the effort - should be straightforward. Finding out a common place for the function and updating minor things here and there.


edmofro (commented):
Super nice refactor, it all looks clean and useful to me! Sorry to say though, I'd prefer not to merge it, as we don't need it right now and with sync there is always a risk that things go wrong because it's easy to have blind spots. 

@chris-bes noting this refactor for future use if we need to add `updatedAtByField` to any other models

PR #4527, opened 2023-08-07
EPI-583: Update patient merge logic

NavarroEmilioLuis:
### Changes

We want to keep everything from the selected patient but if a field does not exist, then we should look into the unwanted record and copy it over.


edmofro (changes requested):
So nice to have intuitive clean logic through all of this, thanks!

Main pushback would on whether there are more tests we can add 🙏

edmofro (approved):
lgtm, thanks for the efforts on this one @NavarroEmilioLuis!

PR #4530, opened 2023-08-08
NASS-737 NASS-735: Migrate local.json config to settings table

dannash100:
### Changes

**CI issue** ✅ 
Having this issue on the step `CI / Test migrations server=lan` when NODE_ENV is `test` in facility test migration step it selects `balwyn` as facilityId  but doesn't exist in context of test setup - `'Key (facility_id)=(balwyn) is not present in table "facilities".',`
- I have have put in a case to return early in case of facility context with 'test' env and no facility
- Underlying issue is that multiple places in the tests expect facility from test.json to not exist and it was turning into a bit of a high touch change to address.
- A fix might be to either change the ENV of test migration step to somthing like `test-migration` to return early on?
- Will address when swapping out config usage in test env where appropriate

**Test failures from snapshots** 🕐 
Due to the test.json content being different in auto deploy env, the setting subcommand snapshot-based tests are breaking due to the migrated settings adding ones not present in the commited test.json.
Potentially we should address this when updating subcommand with scope considerations

**Please check out:**

The bulk of actual logic is in [packages/shared/src/migrations/1691371570736-populateSettingsFromLocalConfig.js](https://github.com/beyondessential/tamanu/pull/4530/files#diff-bb9b90e5bbb8cca09a614992a86cc298e8f799f1e634015de515e59da866625d)
A large amount of diff is the restructure of config values like creating scoped default files.

**Migration handles:**
- Migration of local.json config to settings table
- This is determined by new scoped defaults located in shared `settings/facility.js`, `settings/central.js` and `settings/global.js`
  - Going foward from epic this is where new config is defined
  - These new default files are used to determine which config goes where and which config keys (like db) should be left
  behind (credentials, db connection config etc)
- Only central server migration creates global scoped settings
- Restructures localisation
  - Notably localisation is added in `global` scope
  - Moves `localisation.data` to just `localisation` - as well as things like merges `localisation.timeZone` and `localisation.countryTimeZone`
  - Move a handful of keys out of localisation (these where historically added to be available across contexts) examples include `country` and `features`

**State of config after migration:**
- Config files are left as is for now while we continue development on epic
- Existing db defined config settings will not be effected by the up and down of this migration
- Proposed new state of config files (things remaining in node-config) have been added to this pr in a subfolder ready for replacement when further into code changes related to this epic
- Credentials are not added to db
- Settings table now contains all relevant scoped entries from local.json
Screenshot 2023-08-14 at 1 21 49 PM

**Next steps:**
- The actual fallback to this default values will be handled by settings utility which handle any caching and scoped access of settings
- Settings default files to become schemas with validation, information and defaults (mabye something to do with sensitivity of keys re permissions) for each possible setting
along these lines.

```
'deletionAction': {
  defaultValue: 'RENAME',
  typehint: 'choice',
  choices: ['RENAME', 'DEFAULT', 'NONE'],
  validation: yup.string().oneOf(['RENAME', 'DEFAULT', 'NONE']),
  description: 'Strategy to use to handle old patients after a merge',
  longDescription: `
- RENAME sets the patient's firstName to 'Deleted' and lastName to 'Patient'
- DESTROY sets `deleted_at` to a timestamp
- NONE is a no-op and does nothing
  `,
}
```

### Checklist

- [x] Code is finished
- [n/a migration] Tests are written

_Upon merging:_


passcod (changes requested):
generally good a few notes only

passcod (approved):
LGTM

PR #4531, opened 2023-08-08
TAMOC-44: Fix loading indicator obscuring modal

mcccclean:
### Changes

The loading indicator z-index was set too high and was rendering over the top of the modal. I can't tell why it was set at all? But I'm going in with a light touch and have just brought it down a little bit.

### Checklist

- [x] Code is finished

### DIFF
--- a/packages/desktop/app/components/DropdownButton.js
+++ b/packages/desktop/app/components/DropdownButton.js
@@ -95,7 +95,7 @@ const MenuButton = styled(Button)`
 
 const Popper = styled(MuiPopper)`
   margin-top: 2px;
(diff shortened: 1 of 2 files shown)


biaoli0 (approved):
@mcccclean Thanks for catching that! Worth to check if `LoadingIndicator` in modal will show up, e.g. in `BirthNotificationCertificateModal` as they are sharing the same z-index (modal has 1300 z-index in my memory). Theoretically it should work perfectly because `LoadingIndicator` is rendered afterward.

biaoli0 (changes requested):
Request changes for potentially hiding a loading spinner on modal.

biaoli0 (approved):

PR #4532, opened 2023-08-08
TAMOC-42: Add a min width to the programs pane

mcccclean:
### Changes

I couldn't figure out why the modal was showing up as a normal width previously - I wonder if it used to have some additional content that was removed? But we probably never want it to show up narrower than about 500px either way, so I'm just including that as a new min-width to try and keep a light touch.

### Checklist

- [x] Code is finished

_Upon merging:_


edmofro (approved):
A shame we don't know the root cause, but this seems fine

PR #4538, opened 2023-08-09
Iterate packages in dependency order for build-shared

passcod:
### Changes

- Iterate packages in dependency order for build-shared
- Define sharedness of a package programmatically instead of statically (ie "has something that depends on it")
- And therefore make a top-level `build` script that also resolves the dependency order to avoid issues (it's the same script)
- Run build-shared in container build instead of just the shared package's build, so it (will) works properly with multiple "shared" packages.

### Checklist

- [x] Code is finished

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,8 @@ RUN apk add --no-cache \
     jq \
     make \
     python3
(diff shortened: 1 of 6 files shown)


edmofro (approved):
This is such a common thing I'm surprised there's not better inbuilt support for it.

PR #4541, opened 2023-08-09
Adjust common configs for existing and new packages

passcod:
### Changes

Eclectic set of fixes extracted from the big PR.

- Jest coverage from ts and js
- move-dts script in build-tooling to wrangle tsc putting types not where it should
- Update typescript configuration for version 5.5+
- Remove includes from common tsconfig as they were getting misinterpreted (resolves relative to the common file, which is nonsensical here)
- Split build and lint steps for more fine-grained runs in development
- Use tsbuildinfo file for incremental typescript builds
- And add `yarn clean` command to get rid of these + builds for when clean builds are wanted

### Checklist

- [x] Code is finished


dannash100 (approved):
Looks good @passcod!

PR #4543, opened 2023-08-09
Split out Tamanu API client

passcod:
### Changes

Create new package `@tamanu/api-client` which contains a client for the Tamanu (internal) API, for use within other Tamanu packages. This is the client defined in desktop, though at a latter point it might be interesting to join to that the client defined in facility/lan for sync.

~~Includes a CLI API client:~~

image

### Checklist

- [x] Code is finished
- [x] CI passes (dependent on upstream PR)


mcccclean (changes requested):
This is really neat! Just one change request but will yield in the face of much pushback on it.

mcccclean (approved):

PR #4550, opened 2023-08-10
Use a single lint job in CI

passcod:
### Changes

Make the lint-all script able to continue on errors (with exit(1) at the end if any package failed) so we can lint all packages in a single job run in CI, instead of spinning up N jobs for N packages.

This should also prevent CI from passing if lint fails, which it wasn't doing previously in some cases.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Looks good!

PS: though it seems package `csca` is missing something, maybe?

PR #4553, opened 2023-08-10
Provisioner: support programs

passcod:
### Changes

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
This looks good to me. I have a comment about the file itself:

I noticed that we repeat the pattern of checking the errors before continuing in at least two places which makes me think about the overall logic - this is using like an abort early behavior, which is fine but it makes me wonder if there is any error along the line, do we care about rolling back the changes?

PR #4554, opened 2023-08-10
TAN-2276: Remove scrollbar on document printing

tcodling:
### Changes

Currently on windows when printing the new patient letters a scrollbar will appear in the actual PDF. On further investigation we discovered that this is because the printable prop was added to the document preview modal but the document preview logic isn't setup for printing so we into problems.

To fix this I added a custom print handler which puts the PDF data into an iframe which is then printed. 

Because this was a windows specific error I actually let Klaus test and pass this on autodeploy before code review as we weren't sure if we were even going down the right path and I couldn't test on my Mac.


NavarroEmilioLuis (commented):
Well you already mentioned this works. I believe it's a valid implementation to use an iframe for this but this makes me wonder - why can't we use `PrintPage`?

NavarroEmilioLuis (approved):
Sounds good to me - happy to move on

PR #4556, opened 2023-08-11
Only warn on job grab fails if it's a large amount

passcod:
### Changes

This makes job grab failures a debug log most of the time. It's annoying that this fails on first go but as most of the time it succeeds on second go it's more annoying to litter logs with a warn.

If more than 5 retries warns will be emitted (new behaviour), and if it reaches the limit it will error (was already the case).

### Checklist

- [x] Code is finished


stretchkennedy (approved):
🎉

PR #4562, opened 2023-08-14
Fix for desktop build

passcod:
### Changes

Core bug was garbage being prepended to `yarn -s workspaces info` which made the process not complete successfully, as well as the new package-aware build-shared not being run for desktop properly.

### Checklist

- [x] Code is finished
- [x] Testing: download the builds attached to the PR and test they work


edmofro (approved):
lgtm!

PR #4564, opened 2023-08-14
[NASS-818] Export and Import reference data in chunks

renatokoyama:
https://linear.app/bes/issue/NASS-818/importexport-with-chunking

### Changes

- Created a new endpoint that allows the upload/download using chunks of data
- This is needed due to a validation we have on `nginx` for larger files 

## Export

https://github.com/beyondessential/tamanu/assets/17033058/1d0859f9-4706-4c52-bc0d-30f620514326

## Import

https://github.com/beyondessential/tamanu/assets/17033058/2546e145-d0c3-4c63-87b8-3e9d3f2b4ee6

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/api/TamanuApi.js
+++ b/packages/desktop/app/api/TamanuApi.js
@@ -305,6 +305,34 @@ export class TamanuApi {
     });
   }
 
+  async uploadFileInChunks(endpoint, filePath, options) {
+    const fileStats = await promises.stat(filePath);
+    const fileSize = fileStats.size;
+
+    const chunkSize = 5 * 1024 * 1024; // 5MB
+
+    let start = 0;
+    let end = Math.min(chunkSize, fileSize);
+    const fileContent = await promises.readFile(filePath);
+    while (start < fileSize) {
+      const chunk = fileContent.slice(start, end);
+      const headers = {
+        'content-type': 'application/octet-stream',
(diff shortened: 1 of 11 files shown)


mcccclean (commented):
<edited: accidental submit>

mcccclean (changes requested):
~I've got a few architectural-level change requests for this PR.~

~First, I'm not quite sure what problem it's solving; the only issue mentioned in the Linear ticket is the nginx limit, which I think we can simply adjust to a higher limit rather than building this all out.~

~Second, I think if we decide that we need to chunk our uploads and downloads (ie if we want better display of transfer progress?) I think it would be much cleaner to do that through the FileReader stream API rather than splitting uploads across multiple independent requests.~

Edit: spoke with the tech strategy team and got a better picture of the context behind the decision to chunk stuff, ignore the above, I don't think streaming is the right answer for this.  :pray:

mcccclean (changes requested):
Ok, I think we should pause this one for now. I think that we're probably better off using a third-party solution to get resumable uploads rather than trying to roll our own, but it probably could use a bit of investigation as to which library to use.

The top candidate seems to be TUS, which has a [server](https://github.com/tus/tus-node-server) and [client side](https://github.com/tus/tus-js-client) package for JS that seem pretty close to what we're after. What do you think?

PR #4569, opened 2023-08-15
NASS-850: Clean up lab panels table names

dannash100:
### Changes
- We are going to add supersets and became apparent that panel tables names are a bit too specific and long.
- This pr drops the "test" from them - i.e from `lab_test_panels` to `lab_panels`
- Add migration for mobile (these data types are not used on mobile yet but need to be synced)
- Supersets will follow this naming and just be `lab_supersets`
- Without dropping sync triggers - we end up with two triggers with different names when the post migration steps create new ones for the new table names.
- P key index doesn't get renamed like the foreign key ones automatically so I did that too so its a full rename on every level.

Mobile doing full sync:
Screenshot 2023-08-17 at 10 18 58 AM

**TODO:**
- Will need to update any reports in the reports repo too if they exist referencing panel yet.
- Update reference data sheets, and also s3 reference data for auto-deploy for lab panels
- Update documentation 

### Checklist
- [x] Code is finished
- [x] Tests are written

_Upon merging:_


edmofro (changes requested):
All change requests are super minor, you've done an amazing job of being super exhaustive here. Of course super easy to miss something for both of us, so this will need solid testing.

Glad we checked in with Maui on how much of a burden this places on them, I should have thought of that before you'd put in all the effort but fortunately Juliana is happy! Good to be considering the wider team more as historically I've been focused on just the dev stuff and once that's done it's invisible work.

edmofro (approved):

PR #4580, opened 2023-08-16
TAV-738: Release logic improvements

passcod:
### Changes

- TAN-2265: Find release by name and tag, rather than by exact tag only, and check both `1.2.3` and `v1.2.3` variations, so we don't miss publishing an existing release draft.
- Swap the "make next draft" (which never fails) and the "publish release" (which might) steps so we don't skip the former if the latter fails.
- TAV-738: If we're publishing a hotfix on a branch that is not the latest published branch, we shouldn't mark it as latest.

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/publish-release-version.yml
+++ b/.github/workflows/publish-release-version.yml
@@ -53,15 +53,6 @@ jobs:
 
     # do release management after the critical bits above as we care less if those fail
 
-    - name: Publish the draft release if one exists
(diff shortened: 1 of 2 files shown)


edmofro (changes requested):
All makes perfect sense, just a few nits. Happy for you to push back on any of them.

edmofro (approved):

PR #4591, opened 2023-08-17
NASS-854: Add unique constraint on report definition name

dannash100:
### Changes
The majority of this migration based on the [location code uniqueness migration ](https://github.com/beyondessential/tamanu/blob/main/packages/shared/src/migrations/1679627805922-uniqueLocationCodes.js)that @mcccclean wrote.
I did notice in running on my local the down migration did not remove the unique constraint when using `updateColumn` with unique: false on the down migration, which is interesting. 

Dropping constraint directly did however work to do that.

- Add migration to ensure that report definition names are unique as this is an assumption in the logic for db-defined-reports
- I believe adding index inside the indexes parameter in model is only used in `sequelize.sync()` to setup the db, so not necessary - but maybe best to leave that in I thought.

### Checklist

- [x] Code is finished
- [n/a] Tests are written
- [] Testing notes added to the Linear issue

_Upon merging:_


edmofro (approved):
lgtm, we'll have to add a note to the manual release steps to look for duplicate report names and rename any as part of whatever release this goes into

PR #4597, opened 2023-08-18
TAV-735: Upload builds to release and write client download list

passcod:
### Changes

When a tag is created, we build desktop clients. Currently, those are silently uploaded to S3. With this PR:
- Two variants (Windows exe and msi) are also uploaded to the github release
- The full list of public S3 URLs that you get on PRs is copied to the release description

The files uploaded to the github release are renamed to a permanent name like `TamanuSetup.x64.exe`; that means that we have a permalink to the latest release on github:

<https://github.com/beyondessential/tamanu/releases/latest/download/TamanuSetup.x64.exe>

### Checklist

- [x] Code is finished
- [x] Has been tested

### DIFF
--- a/.github/workflows/cd-client.yml
+++ b/.github/workflows/cd-client.yml
@@ -161,13 +161,13 @@ jobs:
           rm -rf release
 
       - name: Use release bucket
-        if: inputs.bucket == 'release' || startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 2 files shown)


edmofro (approved):
Great change and good to get that fix in before today's release workflow. I made a few comments but no required changes, so merge when ready

PR #4600, opened 2023-08-20
SAV-358: Include facility id in start session logs

edmofro:
### Changes

When connecting to central server to debug sync issues, it's a common frustration that the facility that is connecting is only available once it gets into the pull phase. This PR sends through the `facilityId` when it first starts a sync session, for logging in both the `debug_info` column and the honeycomb logs

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/lan/app/sync/CentralServerConnection.js
+++ b/packages/lan/app/sync/CentralServerConnection.js
@@ -211,7 +211,10 @@ export class CentralServerConnection {
   }
 
   async startSyncSession() {
-    const { sessionId } = await this.fetch('sync', { method: 'POST' });
+    const { sessionId } = await this.fetch('sync', {
+      method: 'POST',
+      body: { facilityId: config.serverFacilityId },
+    });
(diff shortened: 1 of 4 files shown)


mcccclean (changes requested):
Oh yeah this is useful, and I don't think it's too risky (like, if there's anything broken about it we'll find out well before it hits any production branch).

The only downside I see is that the `facilityId` is not particularly useful if it hits the logs without a corresponding `deviceId`, just because mobile devices often share a facilityId with a facility server, and without the deviceId (which takes the form of "mobile-$shortid" or "facility-$shortid" respectively) it's impossible to tell them apart from the server logs.

edmofro (commented):
I think that's covered? See inline comments

mcccclean (approved):
Oh huh, yep, deviceId all present! I... wonder what I was looking at :thinking:

PR #4601, opened 2023-08-20
TAV-735: more release fixes

passcod:
### Changes

- Replace GraphQL create commit API with good old git CLI commands
- That splits a single function in three but ah well
- `make_latest` in the API is a boolean [AS A STRING WTAF](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release)
- address minor comments from previous PR

**1.31 cut release workflow has been run with this code, no need to re-run**

### Checklist

- [x] Code is finished


edmofro (approved):
LGTM

PR #4613, opened 2023-08-22
EPI-601: Create FHIR practitioner resource

NavarroEmilioLuis:
### Changes

Just adding a new FHIR resource!


mcccclean (approved):
LGTM.

This is a little bit of a rubber stamp as I'm not super familiar with the FHIR materialiser setup, so I wouldn't notice any issues at the architecture level. But I'm assuming that this is just a pretty routine operation and it's just duplicating existing patterns so that'd be fine?

PR #4615, opened 2023-08-23
NASS-855: Add db_role to report_definitions column migration

tcodling:
### Changes

~~This card adds a new column to the `report_definitions` table that contains a string defining what hardcoded DB reporting user should be used to run the report (either `tamanu_dataset_reporting` or `tamanu_raw_reporting`. This is part of a bigger project that is adding different permissions regarding which schema to access when creating db defined reports.~~

~~All pre-existing DB defined reports are set to `tamanu_raw_reporting` when the migration is run.~~

After some discussion we decided to make these users/roles configurable. Now this PR covers the migration adding the new column and filling the db reporting role of already created reports with the "raw" reporting role name defined in the config.

I have also setup a new key in the DB section of config for sync and lan called "reportCredientials" that contains a "raw" and "dataset" postgres user credentials.

```
"reportCredentials": {
      "raw": {
        "username": "",
        "password": ""
      },
     "dataset": {
        "username": "",
        "password": ""
      }
    }
```

### Checklist

- [x] Code is finished

_Upon merging:_


stretchkennedy (changes requested):
One potential gotcha with denormalising the db role name like this (but LMK if I've missed something, there's probably a lot of design context I have no idea about 😅 )

stretchkennedy (dismissed):
LGTM 👍 

See comment for one caveat, but if Edwin's cool with it then all good

stretchkennedy (changes requested):
Nearly all LGTM! One gotcha about validating the value of `dbRole` against the values of the constants, other than that it's great 😄

stretchkennedy (approved):
LGTM :shipit:

anilvkaran (approved):

PR #4618, opened 2023-08-25
TAN-2299: Notes UX/UI updates

NavarroEmilioLuis:
### Changes

Please check the card item by item. It's almost 1 commit per update except two quick refactors/lint.

There was one item I had to redo: increase the table size - long story short, it seems we really want to preserve the scroll within the table body, that way being able to attach the lazy loading handler properly. Otherwise it would have to live in the actual table component and I think we probably do want to reuse that anyway.

### DIFF
--- a/packages/desktop/app/components/NoteChangeLog.js
+++ b/packages/desktop/app/components/NoteChangeLog.js
@@ -19,7 +19,7 @@ const StyledListItemText = styled(ListItemText)`
     width: 100%;
   }
 `;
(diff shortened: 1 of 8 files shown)


biaoli0 (approved):
@NavarroEmilioLuis Great work to me! Can't tell any flaws. Good use of the forwardRef().
In some usage of `lazyLoading`, I prefer to use the name you came up with, `isNoteTable` rather than `lazyLoading`, but it is up to you to change it.

PR #4619, opened 2023-08-25
NASS-856: New permission for db schema editing field when creating reports

tcodling:
### Changes

In NASS-855 we added a new column to the report definitions table called `db_schema` that references the postgres DB role which determines the relevant schema to access when running reports. We need to add a new field to the report creator that lets them choose one of these roles if they have the appropriate permission.

This card covers:
- Adding the field to report creation to save which schema the report should be run on (`reporting` or `raw` as a string in the `db_schema` column)
- Adding a display field with the DB role to the display section of the "Edit Report" tab
- Adding a permission check to the report creator that only shows the DB role field when the user has the relevant `write ReportDbSchema` permissions
- Bonus: Front end sorting wasn't implemented for the reports list that this was branched off. I completed that in this PR

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


stretchkennedy (changes requested):
Couple of nits, but the comparator is the only big gotcha

stretchkennedy (approved):
LGTM! 🎉 

Really great to see the sort order bug get fixed here, RAYG is wonderful

PR #4620, opened 2023-08-25
EPI-615 Use the right key for onBehalfOfId

biaoli0:
### Changes
onBehalfOfId does not get passed along which create this issue.


stretchkennedy (approved):
LGTM 👍

PR #4626, opened 2023-08-28
TAMOC-54: Fix broken builds due to @tamanu/api-client not building

stretchkennedy:
### Changes

1. Github Actions runs the root Dockerfile as part of the CD Client/Build desktop client job
2. The Dockerfile builds shared by removing everything except shared
3. The _do-with-all-packages.mjs script determines whether a package is shared by whether it's depended on by other workspaces packages
4. Because 2. removed everything except shared, @tamanu/api-client is not dependent on by anything, and is not built
5. For some reason this doesn't break the build and instead the client is built with a missing package

### Checklist

- [x] Code is finished

_Upon merging:_

### DIFF
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -26,7 +26,11 @@ common() {
 
 build_shared() {
   yarn clean
-  yarn build-shared
+  # as `yarn build-shared` removes anything that's not a dependency of another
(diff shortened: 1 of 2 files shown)


passcod (approved):
very strange! i wonder if we'd better skip doing the separate shared build step, and build the desired package at once

PR #4644, opened 2023-08-30
TAN-2274: Hide record death button on mobile

tcodling:
### Changes

It looks as though the death workflow for mobile was partially created but never actually implemented quite a long time ago (~3 years). The front end forms have been completely created from what I can tell however on submit it simply logs the values submitted.
image

This card is just hiding that button in order to reduce confusion for the end user until it is actually implemented. Although as mentioned in my comment I am unsure if it is better to comment the code out temporarily or just delete for now knowing we can dredge from git history if we need to as much of the work still looks valuable

### Checklist

- [x] Code is finished

_Upon merging:_


renatokoyama (approved):

renatokoyama (approved):
LGTM

PR #4647, opened 2023-08-31
EPI-452 Add prefilter criteria before materialisation

biaoli0:
### Changes:
FHIR has heavily integrated with raw query so I think `pre-filter criteria` should be also a query. It has a similar concept as `getQueryToFindUpstreamIds()` to generate a Sequelize query, then converted to raw query. 

#### It is used in the following areas as also mentioned in the ticket:
1. sub command fhir refresh -> doRefresh()
2. FhirMissingResources schedule task
3. ALL_FROM_UPSTREAM job worker

#### Current filtering logic: 
Filter out `encounters` have `encounter_type = 'surveyResponse'` 

#### Refactoring in the future: 
Not in this PR, but I really want to refactor these functions above to be an abstract class to capsulise the steps they both share:
1. fetch upstream ids
2. filter upstream ids
3. run: call FHIR job worker or call Resource.materialiseFromUpstream(upstreamId)
4. delete those FHIR resource doesn't meet pre-filter criteria. (TODO)

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Incredible work!

I left a couple of comments, I think this is good enough to continue but want to make the most out of the time we have and try to improve it a bit so we don't get caught off-guard. LMK what you think!

NavarroEmilioLuis (approved):
Looks great. Not mandatory but one final suggestion would be to change the name of `fakeResourcesOfFhirServiceRequest`. Given we have `fakeResourcesOfFhirServiceRequestWithLabRequest` is weird that the other one doesn't mention it's with imaging request so probably `fakeResourcesOfFhirServiceRequestWithImagingRequest`?

PR #4650, opened 2023-08-31
EPI-599: Add encounterType history to the medici report

NavarroEmilioLuis:
### Changes

Just updating the shape of the report to include all encounter types instead of the latest one


biaoli0 (commented):
Looking good! Just a question before approval.

biaoli0 (approved):

PR #4652, opened 2023-09-01
EPI-620: Update labs printout

NavarroEmilioLuis:
### Changes

Pictures in linear. Changes are meant to be very contained for this update though I wonder if we should make all `TableCellTag`s not wrappable (would only affect imaging requests statuses aside from the labs)


biaoli0 (approved):
LGTM!

PR #4659, opened 2023-09-03
SAV-83: Add gap between persisted batches when pulling

chris-bes:
### Changes
- Attempted to replicate the lag locally:
  - (Local) marking 250k of encounters for sync (with all patients have been marked for sync, so all of the encounters should be pulled down)
  - Then I navigated to pages that involve encounters + perform any actions that update/create encounters
  - Result: No lag observed

- Attempted to replicate the lag in Aspen clone:
  - Marking 2.2 million of vital_logs in Aspen clone for sync (around 700k pulled down to Lautoka clone, and 1.5 millions pulled down to Ba clone)
  - Then I navigated to pages that involve vitals + perform any actions that update/create vitals
  - Result: No lag observed

However, even if it was replicatable, I think that it is still likely because the connection pool got exhausted. So I think with the combination of `persistedCacheBatchSize` and `delayTimeBetweenPersistedCacheBatchInMilliseconds` configuration that are adjustable, we should able to get some gap time to free up and allow active usage to be able to happen.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


edmofro (changes requested):
I imagine you're writing up your investigation and findings about this now, but diving in with an early review just to suggest a couple of name changes. This is a nice simple fix, thanks @chris-bes!

edmofro (approved):
Thanks for clarifying all the thinking. I think this makes sense as a tool in the toolbox, even if we can't prove it fixes the original issue. Lgtm!

nick-beyondessential (approved):

PR #4661, opened 2023-09-03
NASS-817: Differentiate between suggester endpoints and those support /all in report params

dannash100:
### Changes

- SuggesterSelect fields use `/all` suggester endpoint and therefore don't work for every suggester causing an error from mismatch from the suggester select in report form parameters and report generation form.

### Checklist

- [x] Code is finished
- [n/a] Tests are written

_Upon merging:_

### DIFF
--- a/packages/constants/src/suggesters.ts
+++ b/packages/constants/src/suggesters.ts
@@ -1,15 +1,19 @@
 import { REFERENCE_TYPE_VALUES } from './importable';
 
-export const SUGGESTER_ENDPOINTS = [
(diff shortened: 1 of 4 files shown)


edmofro (commented):
Yes nice! 

There would be an optional extension to apply the same validation mechanism that @renatokoyama did for general suggesters to validate that the list of those supporting `/all` is kept in sync, but not necessary for the release

I just have one question so will hold off approving until I understand the extra change

edmofro (approved):

PR #4665, opened 2023-09-04
NASS-689: Remove session expiry message when opening Tamanu after logout

tcodling:
### Changes

Currently if you log out of and then close Tamanu, it will show a "Session expired: Please log in again" error to the user when they reopen the app which is confusing. 

This is because the token is not cleared when logging out so it will restore the session when re-opening, however it won't be valid so the user is kicked out when Tamanu makes a request to the patients table and gets rejected.

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM

PR #4666, opened 2023-09-04
NASS-817: Validate suggestions considering possible variations of /:id and /all endpoints

dannash100:
### Changes

**Update**
- changing angle and just making sure all the routes have matching all endpoints
- only two outstanding ones which don't make sense to have an all suggester right now `patientLabTestCategories`,
`patientLabTestPanelTypes` due to there dynamic base query of patient id.
 
Addresses something we found we offering lists of suggesters in report generation parameter form:
- One suggester has only an `/all` route (had to change things a bit from initial look for this one)
- Many suggesters have no `/all` route

This extends Renato's work and checks that
- There is no suggester in the supporting_all constant list that doesn't have an `/all` route
- That there is no suggester in supporting_id constant list that doesn't have an `/:id` route
- This means that if it is in both lists it will check for `/all` and `/:id` route.

It also checks the other way around - like that there are no routes with no constant entry as per Renato's work

### Checklist
- [x] Code is finished
- [n/a] Tests are written

_Upon merging:_


mcccclean (changes requested):
Code-wise this LGTM.

However - I might be missing some context for this but, is this approach kind of papering over an architectural flaw?

The goal of the suggester API is so that frontend selection-type components don't have to know anything at all about the backend, as long as we code our autocompletes and dropdowns etc to know the suggester API generally, and make the backend treat all suggestions the same, then we can plug anything into anything and not have to worry about it.

So if a suggester is missing the `/:id` route, any selector component pointing to it will break when added to an edit-existing-record form, as components use that route to query the server for what it should display for its initial value. And a backend suggester missing the `/all` route won't be able to support Select components.

So, yeah, just want to gently push back against this approach - is there a reason we don't just implement `/:id` and `/all` endpoints for the suggesters that are currently missing them? I feel like rather than codifying these varying degrees of suggester support, we should just say if something is going to implement part of the suggester API, it should implement the whole suggester API.

(unless there's a good reason! like I said, might be missing some context here. also, the work of implementing `/all`s may well be a maintentance task rather than a rerequisite for this PR)

mcccclean (approved):
@dannash100 Oh yep, I like that approach, good thinking.

PR #4676, opened 2023-09-06
NASS-739 using db config fetcher

renatokoyama:
### Changes

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (commented):
(Responding to some questions from a Slack DM):

I think this is a good start. The SettingsReader utility looks correct to me, and I think it's OK that some methods will have to switch from sync to async to be able to read config according to the new system - we knew that would have to happen when planning this feature, I think it's unavoidable.

(I suppose an alternative would be to give the SettingsReader a way to read + caches all of the available settings in advance? but I think that's adding more complication than just making some functions async).

There's one change that I would suggest though, which might make things read a little bit simpler. Rather than updating the functions to accept the settings parameter, update them so that they're accepting a context object with the models + settings reader. 

To use an example:

```javascript
// current state
function myUtility(models, parameter1, parameter2) {
  const mySetting = config.mymodule.setting;
  // etc etc
}

// approach used in this PR
function myUtility(models, parameter1, parameter2, mySetting) {
  // etc etc
}

// application context approach
async function myUtility({ models, settings }, parameter1, parameter2) {
  const mySetting = await settings.get('mymodule.setting');
  // etc etc
}
```

The central server module already actually uses a formal ApplicationContext object, which I think will be extended to become a utility in @shared and include settings, permissions etc as well as models, but that will probably be waiting for a maintenance task. For now, I think it will be enough to just keep that pattern in mind: a single object that contains all the links to the other application systems necessary to perform a given task.

PR #4680, opened 2023-09-08
TAV-784: Replace cls-hooked with built-in AsyncLocalStorage

passcod:
### Changes

`cls-hooked` hasn't been updated since 2017 and has known performance issues. [AsyncLocalStorage](https://nodejs.org/docs/latest-v16.x/api/async_context.html) is built on the same async-hooks API underneath but is an optimised and memory-safe implementation maintained directly by the Node.js team.

This is a tiny shim which implements only the parts that sequelize uses.

### Checklist

- [x] Code is finished
- [x] Tests pass

### DIFF
--- a/packages/lan/package.json
+++ b/packages/lan/package.json
@@ -46,7 +46,6 @@
     "case": "^1.6.3",
     "chalk": "^5.0.1",
     "chance": "^1.1.8",
-    "cls-hooked": "^4.2.2",
     "commander": "^9.0.0",
     "compression": "^1.7.2",
     "config": "^3.0.1",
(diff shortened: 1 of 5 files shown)


mcccclean (approved):
Yessss this is GREAT, cls-hooked has caused us a few headaches before so it's fantastic to not be tied to it any more. Couple suggestions but nothing critical.

mcccclean (approved):

stretchkennedy (approved):
Yeah, my slack message was in relation to hooks running outside the closure, this looks fine 👍 

From memory Ethan had a different fix for it that didn't involve messing with cls-hooked

PR #4688, opened 2023-09-11
EPI-384: Allow surveys to omit data from PatientData fields

stretchkennedy:
### Changes

Makes surveys configurable to omit data from a PatientData field from the final result.

I've removed the submitted questions in the UI, based on what was in the card and from our refinement. This means calculations (in e.g. Result or CalculatedQuestion) can't use them. I've also only added the `omitFromResult` flag for `PatientData` imports, although this isn't actually validated yet and will thus allow `omitFromResult` for all question types (see https://linear.app/bes/issue/NASS-826/enable-validate-survey-question-config-on-import).

Haven't written tests because it's primarily in the UI, which we don't have a good way of testing now.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (dismissed):
Looks good overall but there are some strings that do not match constants and it's probably going to brake or was broken?

mcccclean (approved):
LGTM! A couple cosmetic suggestions, doesn't need re-review.

PR #4719, opened 2023-09-18
NASS-789: Support commas in options for multi-select component

tcodling:
### Changes

Currently we can't select options in a multi-select field if they have a comma in them. Because the logic for the multiselect field is based around joining and splitting by commas, this causes problems when one of the options contains a comma and makes it un-selectable.

The solution we decided on after a while to this was to use JSON arrays for the logic surrounding multi-select fields. Instead of constantly joining and splitting strings by `,` all over the app, we now save as a JSON valid array string in the DB and then parse when to be used/displayed on the front end. This removes a bunch of weirdness around splitting logic and also now allows commas to be used as options which was never possible before.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


chris-bes (changes requested):
LGTM! Just 1 or 2 concerns I have. Really excited for this change!

chris-bes (approved):
LGTM!

chris-bes (approved):
LGTM!

PR #4724, opened 2023-09-21
NASS-863: Add permission check for lab results table

tcodling:
### Changes

Currently on the lab results table on patient view we don't indicate when the user doesn't have permission to view the results table. Instead we see a "no data" message which could be confusing. In this card we added a new `read` `LabTestResult` permission to be checked when viewing this table and also improved the error message for clarity.

Before
image

After
image

### Checklist

- [x] Code is finished

_Upon merging:_


Alastair-L (approved):
LGTM! One small nit

PR #4725, opened 2023-09-22
TAMOC-65: Make facility server independent

NavarroEmilioLuis:
### Changes

A whole bunch of info on the card. Issue is that facility server cannot start if it fails to connect to the central server. This is obviously bad as it should stand on its own.

~This approach will let it connect through but will have to wait out around 10 seconds where we timeout our request (could be less depending on config, but at the same time, could be more!). Still, this is an improvement in that theoretically we will know how long will a server be offline or not.~ No longer true - fixating this to 2 seconds wait, so total of 4 because we will make two requests.

~We could pierce the `timeout` argument and make it smaller only whenever no token has been set.~ I'm actually applying this, I don't think we should let to config how long is a facility server unavailable, given this step makes Tamanu unusable.

Note that this update will only affect on start, where `this.token` is `null`, whenever we get an invalid token we will try to connect again with another call to `connect()`, so those parameters won't be used anywhere. Hopefully.

### DIFF
--- a/packages/lan/app/subCommands/serve.js
+++ b/packages/lan/app/subCommands/serve.js
@@ -35,7 +35,6 @@ async function serve({ skipMigrationCheck }) {
   await performDatabaseIntegrityChecks(context);
 
   context.centralServer = new CentralServerConnection(context);
-  context.centralServer.connect(); // preemptively connect central server to speed up sync
   context.syncManager = new FacilitySyncManager(context);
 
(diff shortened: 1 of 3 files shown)


mcccclean (changes requested):
Nice work, this has been an annoying issue for ages.

mcccclean (approved):

chris-bes (approved):
LGTM

PR #4728, opened 2023-09-24
Add corepack support

passcod:
### Changes

- Add the [`packageManager`](https://nodejs.org/docs/latest-v16.x/api/packages.html#packagemanager) field to package.json
- Add instructions to the readme

Essentially with [corepack](https://nodejs.org/docs/latest-v16.x/api/corepack.html) there's no need to `npm i -g yarn` and also no need to keep yarn updated globally, and corepack is distributed with nodejs, so there's really nothing to do except:

```bash
$ corepack enable
```

once.

### Checklist

- [x] Code is finished
- [x] On merge: amend onboarding

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ $ pyenv global 2.7.18
 Put eval "$(pyenv init --path)" in ~/.zprofile (or ~/.bash_profile or ~/.zshrc)
 ```
 
+Enable corepack (once):
+
+```bash
+$ corepack enable
+```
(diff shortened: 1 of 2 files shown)


stretchkennedy (approved):
LGTM 👍 

Probably needs a message in #tamanu-dev once it's merged giving migration instructions (`npm uninstall -g yarn && corepack enable`?), but I'll leave that up to you

PR #4730, opened 2023-09-24
Allow configuring deviceId in config

passcod:
### Changes

Make it possible to specify a deviceId in the facility server config. This is used by the auto-deploys.

### Checklist

- [x] Code is finished

_Upon merging:_

- [ ] Update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly) as needed

### DIFF
--- a/packages/facility-server/app/sync/initDeviceId.js
+++ b/packages/facility-server/app/sync/initDeviceId.js
@@ -1,11 +1,16 @@
+import config from 'config';
 import shortid from 'shortid';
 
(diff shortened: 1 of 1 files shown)


edmofro (approved):
I realise this is still in draft, but fwiw lgtm. Approving so it no longer requires codeowner

NavarroEmilioLuis (approved):
LGTM. Though this means we can never change it once we start... Though I guess we'll do that through the database + config?

PR #4731, opened 2023-09-25
EPI-571: Build server zips

NavarroEmilioLuis:
#### This is a historical description first:

This is a continuation of the draft #4418 which was already very close to finish in a way. Aside from the card work, it introduces upgrading v2 actions to v3, which is great.

Alright, I believe this is important context of what I've created so far: in the S3 folder (bes-tamanu-release-servers/test/1.34.0) you'll find 4 different builds (I'll address them with the `test-<n>` they have:

1- Linux build: this one is just the first proper output I was able to generate (the process here seems a bit redundant but might be useful?) this one doesn't work on Windows but works right off the bat on Linux.
2- Windows build in .zip: doesn't work right off the bat but could work after unzipping and creating 2 symlinks.
3- Windows build in .tar: works.
4- Windows build in .tar.gz: works.

Summary of builds above:
- All folders uncompressed are around 209mb size, disk size varies.
- The first 2 take forever to decompress on Windows (30+ minutes with the extraction option provided by the GUI - Powershell commands are not that much better).
- Last 2 decompress around 3 minutes on Powershell with `tar -xvf <path_to_tar_file>`
- Gzip file is the most optimal size (50mb) then comes zip (70mb) and tar sucks (250mb).

#### Now with the actual content of the PR:

There are some outputs in S3 (mentioned above) - I kept what I believe is the best version (the last one). It fits the criteria of "just works" on Windows* but I'm unsure if this work is sufficient. I'm also not sure if it's too much of a workaround.

The way it works:
- Build the Linux image and push to ECR (I think that's almost unchanged from the draft).
- Grab that in ubuntu-latest
- Extract the /app folder
- Remove all node modules here cause it's just faster
- Prepare the production folder
- Zip and upload as an artifact
- Grab that artifact in windows-latest
- Unzip and install dependencies
- Upload to S3

Problems I found in the way:
- Was never able to create a Windows image. Even if I were to succeed, was never able to do `yarn build` on Windows successfully for some reason.
- Windows-latest (even on bash) does not run `zip`

Doubts/things to keep an eye:
- *Only tested on my computer, probably needs to be tested in a clone.
- The length of GHA usage and costs - I ignore everything but seems good with the current setup?
- Was never able to figure out what `Signs the zip` references?
- The folder structure differs from what I've seen: it has some files, node_modules and packages folders, inside packages we have constants, shared and lan. To work the app you need to run inside `lan` but it's not like we can ditch everything else?
- There is a warning `Warning: Your docker password is not masked.` so I'm unsure if I just fucked us up and we need to change that ASAP.
  - https://github.com/beyondessential/tamanu-new-branching/actions/runs/6268184659/job/17022669643
  - Honestly I don't think so given it's a warning
  - I've seen a couple of threads that specify it's normal with that action


passcod (approved):
The code looks good, I think we definitely need more testing before we merge though. If you could rig something to produce 1.33 builds I'll talk to Taveuni about trying this build to deploy Demo 2 today (tuesday), but no worries if you don't manage to in time.

passcod (changes requested):

passcod (changes requested):

passcod (approved):

PR #4735, opened 2023-09-26
EPI-639: Mobile program/referral submissions error

NavarroEmilioLuis:
### Changes

Alright so I did a bit of cleanup in file `SurveyForm/index` that is unrelated to this issue.

Regarding the actual issue, the error was happening [here](https://github.com/beyondessential/tamanu/blob/main/packages/mobile/App/ui/components/Forms/SurveyForm/helpers.ts#L57) where `components` was `null`.

This happened because the `areComponentsLoaded` variable was never updated when the backend effect was trying to actually fetch the components. Given this is async, although unlikely, I thought it was also a good idea to require that `components` exists before moving on, based on the fact that no components means nothing to display on the survey, and unless we get an error, it means we're still fetching the info and somehow the loading variables failed to cover for us.


biaoli0 (approved):
What an odd, I've encountered the same thing yesterday. Thanks for fixing this!

PR #4748, opened 2023-09-29
NASS-857 db reporting user connections

dannash100:
### Changes
Creation of reporting user instances i.e `raw` and `reporting`
Report is then run in the context of either one of these connections based on db_schema

#### A few things to consider
- Having multiple connections
  - I made it so there is a seperate pool config for these reporting instances 🤔
  - Creates a distinction between the main connection and the child report ones - altho this could also be abstracted in another ways: there are a handful of actions in init db like migrations, test recreate db all that which doesn't need to be considered for the reporting connections.
 
- Test context for reports schema and reporting users
  - In this PR I am only initialising the mock schema and roles in the JEST_WORKER_ID (shard? type thing) for the ReportSchema tests to save creating these connections for every test suite.
  - I wonder if there is a better way to organise the creation of this mock schema and roles. I almost thought like a migration that skips if not `proccess.env.NODE_ENV='test'` but I thought that would be weird
  *update* on second thought I think I feel ohk about it being kinda part of demoData

- Currently this is only implemented for reporting instances on facility, but a few small changes for central would get it going

### Checklist
- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


stretchkennedy (changes requested):
Looking good so far! I think we can remove some of the duplication and maybe refactor the way context is created on the facility server, but the rest is minor stuff around naming etc.

stretchkennedy (approved):
3 nits left, otherwise LGTM 👍 
They're reasonably minor and I'm happy for you to merge without fixing them, but I thought you might possibly have missed them

PR #4764, opened 2023-10-05
NASS-883: Translated text replacements

tcodling:
### Changes

This card handles the replacement logic for the new `TranslatedText.js` component. If the component is given a replacements prop which will come in the form of an object with keys for each replacement. It will filter through the string (could be either the translation or fallback value) and then replace each of the colon marked words (`:value`) with the matching key value from the replacements object. This handles JSX. If there are no matching replacements in the replacement object, we just return the original colon value?

Example of component with replacements:
```
<TranslatedText 
  stringId="encounter.changingType" 
  fallback="Changing encounter from :from to :to"
  replacements={{
    from: <b><TranslatedText stringId={encounter.encounterType} /></b>,
    to: <b><TranslatedText stringId={value.encounterType} /></b>,
  }}
/>
```
Will resolve to something like "Changing encounter from **Hospital** to **Clinic**" (as you can see this makes use of the `<b>` tag used)

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Looking good! Couple suggestions

mcccclean (approved):
LGTM, just one tiny note that doesn't need a rereview.

PR #4766, opened 2023-10-05
SAV-440: Refactor sync functions out of centralServerConnection

mcccclean:
### Changes

While starting work on SAV-440 I noticed that there were a few functions in CentralServerConnection (which is meant to manage sending general http requests to the central server, managing parsing JSON, auth, headers etc) that were extremely sync-specific and which probably should live in FacilityServerManager instead (which is meant to manage sync).

This is just moving some functions, it represents 0% new or altered functionality. I figured I'd do it in a separate PR so that the actual functionality changes for SAV-440 don't get mixed up with this work.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

No followup merge actions, this is purely refactoring work.

### DIFF
--- a/packages/lan/app/sync/CentralServerConnection.js
+++ b/packages/lan/app/sync/CentralServerConnection.js
@@ -59,6 +59,7 @@ export class CentralServerConnection {
       retryAuth = true,
       awaitConnection = true,
       backoff,
+      query,
       ...otherParams
     } = params;
 
@@ -78,7 +79,8 @@ export class CentralServerConnection {
       }
     }
 
-    const url = `${this.host}/${API_VERSION}/${endpoint}`;
+    const queryString = query ? `?${objectToQueryString(query)}` : '';
+    const url = `${this.host}/${API_VERSION}/${endpoint}${queryString}`;
     log.debug(`[sync] ${method} ${url}`);
 
     return callWithBackoff(async () => {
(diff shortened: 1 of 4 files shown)


stretchkennedy (approved):
One optional and debateable refactor, otherwise LGTM 👍

edmofro (commented):
I'm not completely sold that the `CentralSyncManager` is meant to just manage sending general requests...In fact from the very origins of this file, back [when it was named `Source`](https://github.com/beyondessential/tamanu/blob/d7311026e7b21bac58750d4a27515b0f4e0d9d5f/packages/lan/app/sync/Source.js) and [subsequently `WebRemote`](https://github.com/beyondessential/tamanu/blob/861343e74fe96bb4557a0d1025ab36c7b2da4991/packages/lan/app/sync/WebRemote.js), it had the idea that it would manage the "connecting to the server" bit of sync-specific logic. Even though I didn't have anything to do with that original architecture, it makes some sense to me, because sync is really the main (and other than login, maybe the only?) reason that the facility server connects to the central server. I also like that separation from the `FacilitySyncManager` class, because it creates a different object that it can inject into the utility functions other than itself, which keeps the concepts clean in my head. I can see an argument for `CentralServerConnection` evolving to the point where sync isn't referenced if the facility server starts to do more than just syncing records, but at this point I still think that's the primary reason it's connecting the central server so it's fine to have it in the same class. 

If you feel strongly about them being separated in this way, I'm actually fine with it now that I've done my spiel, but please make the equivalent change in mobile and then I'll stamp this as codeowner :+1:

PR #4793, opened 2023-10-12
Re-land API package

passcod:
Reverts beyondessential/tamanu#4636, which reverted #4543

This should now be possible, as EPI-571 has landed


mcccclean (changes requested):
Haven't reviewed in detail (I figure it's the same as the previous version, so it's already been reviewed & approved) but it seems like there's some syntax error or new incompatibility? The tests are failing all over the place, haven't inspected in detail but I think it's some typescript thing?

mcccclean (changes requested):
LGTM aside from the TODO!

(won't be around to re-review, if the todo's a non-issue then someone else feel free to :heavy_check_mark:)

mcccclean (approved):

PR #4794, opened 2023-10-12
EPI-635: Make vitals graph range configurable by age

NavarroEmilioLuis:
### Changes

Vitals graph range needs to be configurable by age, this happens at a central place in desktop so it's pretty contained. Here we also add some validation around it.


biaoli0 (changes requested):
Hey @NavarroEmilioLuis , just a few minor things that I requested to change, overall are looking great!

biaoli0 (approved):

PR #4817, opened 2023-10-17
TAV-831: don't bundle-build servers

passcod:
### Changes

This PR removes webpack from the servers (lan, sync, meta). Code is compiled with swc, but no longer bundled into a single file with webpack.

Good:
- Way faster to build (orders of magnitude, from ~10 seconds to ~100ms)
- Way faster to start (2-4×, from 3-8 seconds to 1-2 seconds)
- Stack traces in production have useful file locations (no more Error on app.bundle.js:1:281794)

Meh:
- Slight change in that `node dist/app.bundle.js ...` doesn't work anymore, and has to be `node dist ...` instead
- No longer "obfuscating" sources in built servers
- Very slightly more files output (dwarfed by the node_modules)

Incidentally this PR also rewrites all remaining `import 'shared...'` to `import '@tamanu/shared...'` and removes the `shared` to `@tamanu/shared` alias. These were supported by webpack, but aren't by Node 16. Only about 700 odd such imports, not too bad!

### Checklist

- [x] Code is finished

_Upon merging:_


stretchkennedy (changes requested):
One gotcha with import order in the facility `index.js`, couple of optional nits/questions, otherwise LGTM

That's a lot of files! 😄

stretchkennedy (approved):
🎉

chris-bes (approved):
Had a scan through the sync part and the changes are only for updating imports, so LGTM!

PR #4821, opened 2023-10-18
Don't separately build shared in Docker

passcod:
### Changes

Speed things up ever so slightly by only doing the build once and only installing dependencies twice.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
Oh awesome! You've eliminated a whole intermediate container 🎉

PR #4834, opened 2023-10-20
Fix server package folder from release-X.Y.Z to release-vX.Y.Z

passcod:
### Changes

Noticed Taveuni were renaming that folder to insert the v on OTS.

### Checklist

- [x] Code is finished


stretchkennedy (approved):
:shipit:

PR #4836, opened 2023-10-22
SAV-299: Build importer for historical survey responses

georgekamar:
### Changes

Created an importer for historical survey responses

Two questions here: (file is `constants/src/surveys.js`)
image

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (commented):
Leaving this as a comment rather than a request changes as I still have some of this PR to review, but I wanted to leave the notes I've made so far so that you wouldn't get blocked.

This is great work by the way! Really clean code - I've obviously left a fair few change requests, but it's because the logic was really easy to follow, so it was simple to see where things needed to work differently. :+1: :+1:

mcccclean (changes requested):
Did some further review and left some more notes on the ticket - sorry for the delays :pray:

mcccclean (approved):
Logically this looks alright. I was going to mess around with the import function to avoid the `let answer` (in favour of each branch just returning its answer) but in worrying that I'd break the logic I noticed that this PR doesn't have any tests. I'm gonna approve without them because we've got other urgent stuff on but just a big red flag that this functionality should probably be considered pretty brittle.

Edit: Lot of :x:'s :grimacing: - brittle indeed

PR #4842, opened 2023-10-24
EPI-658: Read only field should be disabled

NavarroEmilioLuis:
### Changes

Pretty self explanatory. This field is used by question types: calculated question, user data and patient data. All of these should be disabled except patient data when there is a specific config for it (and that's already working). This also resolves EPI-656!


biaoli0 (approved):
LGTM!

PR #4843, opened 2023-10-26
EPI-565 Apply soft deletion to encounters

biaoli0:
### Changes

1. Please ignore the messy commit history
2. Each of the clinical models add a new deletionStatus column:
```
  encounters
  document_metadata
  referrals
  notes
  encounter_medications
  invoices
  vitals
  procedures
  lab_requests
  imaging_requests
  survey_responses
```
3. Before an encounter is deleted, destroy all records in these models that referring to it. (like destroy cascade)
4. Add `deletedAt is null` to queries that used to get and count encounters.  
5. Endpoint to delete an encounter
6. Add a check for all encounter crud not to update a soft deleted record  

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


passcod (changes requested):
LGTM in general but a handful of small+style changes

passcod (approved):

edmofro (commented):
@NavarroEmilioLuis I do have some confusion about the approach we've taken here. We seem to have different strategies for desktop and mobile.
- Desktop/facility server, we have fixed all direct SQL that was missing the `deleted_at` filter
- Mobile, we have introduced a new `deletion_status` column, and are filtering on that
Why aren't we filtering on the same thing in both places?

PR #4844, opened 2023-10-26
NASS-919: Report creator permission fixes

tcodling:
### Changes

Klaus found a couple of small bugs with the permissions that were introduced after we introduced the different default reporting schemas based on whether reporting is `enabled: true` or `false`. These two bugs appeared when reports were set to `enabled: false` and without permission to edit dbSchema field

1. List of reports is only showing "reporting" type reports whereas they should see all reports (as per current behaviour) if reporting isn't enabled - **Have added filtering on the endpoint based on config and permission**
2. "Can be run on/dbSchema" dropdown doesn't appear ever as it is dependent on default value which is set to `null` when reports is disabled and no permission - **Now also show when reporting schema is disabled and set to `null`**

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


dannash100 (approved):
This is better imho handeling on backend nice one and thank yoU!

anilvkaran (approved):

PR #4846, opened 2023-10-27
NASS-907: Mobile translation model and migration

dannash100:
### Changes
This is into [feature/nass-891-translation-backend](https://github.com/beyondessential/tamanu/tree/feature/nass-891-translation-backend) which is in review process but I figure that this is seperate enough to put up pr agianst that one now. Then change base branch to epic.

- Adds migration and model for translated strings
- Add composite key logic for typeorm following the `PatientFacility` logic which also uses a composite pkey

Mobile sqlite database after syncing translation
Screenshot 2023-10-27 at 12 31 05 PMs

### Checklist
- [x] Code is finished
- [n/a] Tests are written

_Upon merging:_


tcodling (approved):
LGTM! Only thing I would wonder is if we need to add some sort of caching or localstorage for translations on mobile similar to how we do desktop? Although I would imagine this would be covered in the context card coming up. Worth thinking about however :)

PR #4870, opened 2023-11-06
TAV-866: Implement a test for the facility server working offline

rail-rain:
Open questions:

- [x] What commands should be in `ci.yml`. What should be in `test-facility-offline.sh` and other script files.
- [x] Is the output verbose enough or too verbose?

TODO:

- [x] Make the other test work again
  - It doesn't run because `setup-posgres-one.sh` cannot find the socket file for Postgre while `test-facility-offline.sh` can.
- [ ] Rebase commits into one

### Changes

- "Add test to Tamanu for facility server resilience to network loss"

### Checklist

- [x] Code is finished

_Upon merging:_


passcod (changes requested):
A good start! :) Does seem to have broken the other CI jobs though, so something to check over

passcod (approved):

PR #4880, opened 2023-11-08
Don't cancel any cd-down flows

passcod:
### Changes

Some CD Down workflows are getting cancelled due to overzealous concurrency control. This might be a bit too heavy-handed in return, but as a starting point let's remove the concurrency control and see how that goes.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Ahh hopefully this does the trick

PR #4887, opened 2023-11-09
SAV-493: Add transaction within cert notification task

stretchkennedy:
### Changes

In Nauru, multiple PatientCommunication records are getting created for one CertificateNotification, even when SAV-492 doesn't apply. The most likely cause, from looking through the code, is that creating a PatientCommunication and marking the CertificateNotification as processed isn't atomic, so if the system is interrupted in the middle it'll create two emails.

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (approved):
Looks good!

PR #4894, opened 2023-11-09
EPI-516: Add new report default date range parameter

NavarroEmilioLuis:
### Changes

Reports are getting a new default, last 24 hours! Hardcoded reports should not be affected though


tcodling (approved):
LGTM! :))

PR #4911, opened 2023-11-14
NASS-878: Fix circumstance edit bug

tcodling:
### Changes

Circumstance field was previously throwing a validation error when you tried to edit an existing vaccine record without changing the prepopulated circumstance multiselect field. This was because the initial values are not correctly populated to the formik values object when opening the edit vaccine form. Only when interacting with it does the onchange format the values correctly. Here I just needed to stringify the value when initialised so that we don't need to interact with the field in order to turn it into a string.

Maybe there is room to clean up some of this logic of stringifying and parsing multiselect values as they travel around but I feel it is out of scope of this card to change a core field as late in release.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/desktop/app/forms/VaccineForm.js
+++ b/packages/desktop/app/forms/VaccineForm.js
@@ -156,6 +156,7 @@ export const VaccineForm = ({
             }
           : {
               ...currentVaccineRecordValues,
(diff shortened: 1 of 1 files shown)


biaoli0 (approved):
Just get one suggestion, can preapprove.

PR #4924, opened 2023-11-16
Fix the docker build: include lock and exclude unused packages

passcod:
### Changes

Discovered while reviewing the builds during upgrades and with takayuki that:
- central builds included the lan server, along with csca etc tooling
- the yarn.lock was absent

This may have been introduced last month in #4821.

### Checklist

- [x] Code is finished


rail-rain (approved):
I personally don't know if I have a good understanding in this, but this looks good to me.

PR #4943, opened 2023-11-20
Fix mobile build

passcod:
### Changes

Mostly this is including the `.nvmrc` files to get appcenter to install the right Node version (and also should work for local devs!). Plus some other missed version bits and updating the yarn.lock just in case.

### Checklist

- [x] Code is finished

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/iron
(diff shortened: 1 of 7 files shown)


biaoli0 (approved):
LGTM! I had a good time revisiting the shell command.

PR #4952, opened 2023-11-21
Rename all JSX files to the jsx extension

passcod:
### Changes

This is required for most modern build tooling (following on from [Evan You's declaration in 2021](https://twitter.com/youyuxi/status/1362050255009816577?post_id=1362051769531052039)) and so required for the Tamanu Web work, but done _today_ so it doesn't suffer from inane amounts of staleness.


edmofro (approved):
stamp

PR #4953, opened 2023-11-22
TAMOC-103: Improve performance of fetching notes for an encounter (HOTFIX)

mcccclean:
### Changes

Previously the query was doing a `ROW_NUMBER() OVER (PARTITION BY ...)` over the entire Notes table to determine which notes were the latest. This was logically sound but slowed to a crawl when running on 100k's of notes.

This updates the query logic to first filter for the current `recordId`'s notes (reliably reducing the data set to a matter of 100s) and determining the current notes based on that subset. Then it does the rest of the filtering using the returned IDs for that query. It should be much faster.

Most of this PR is tests, as it turns out coverage was pretty sparse and I didn't want to break anything with the logic change.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/lan/__tests__/apiv1/Note.test.js
+++ b/packages/lan/__tests__/apiv1/Note.test.js
@@ -3,9 +3,13 @@ import {
   createDummyEncounter,
   randomReferenceId,
 } from '@tamanu/shared/demoData/patients';
-import { NOTE_RECORD_TYPES, NOTE_TYPES } from '@tamanu/constants';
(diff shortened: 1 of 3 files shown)


chris-bes (approved):
You're my hero! I love it this is really smart and simple that I couldn't think of 😅. So glad it doesn't require any data migration as well. I've also tested the raw query in Aspen clone and it's only 3 millisecs!
Screen Shot 2023-11-23 at 10 11 32 am

PR #4972, opened 2023-11-27
[NASS-739] adding config fallback

renatokoyama:
### Changes

Added a fallback for reading config when reading settings.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Looking good, just a couple quick notes.

- how does this fit with https://github.com/beyondessential/tamanu/pull/4973/files? is one of these a duplicate?
- it's a bit of a shame that we're adding a dependency on `config` to the shared module, when the ultimate goal is to reduce our usage of it! I wonder if we could use the dependency injection pattern instead to avoid having to do this?

```javascript
// in sync server or facility server initialisation
import config from 'config';

...
app.use(buildSettingsReader(config));
```

PR #5023, opened 2023-12-07
TAMOC-120: FHIR queue running slow

NavarroEmilioLuis:
### Changes

Okay so after a long debug, dropping this index really frees up the worker query. Still, I added here something that will be beneficial for logs.


mcccclean (changes requested):

mcccclean (approved):
Happy with leaving that change as-is, nice work :+1:

Edit: I swear I saw CI pass on this :thinking:? The failure seems to be an intermittent patient name sorting one. Just re-running the tests now.

PR #5032, opened 2023-12-11
SAV-522: Fix flaky test (rename test patients to something less conflictable)

mcccclean:
### Changes

This PR fixes an intermittently failing test, where randomly generated patients were showing up in the results of the patient search tests and throwing off its expectations. I tried to investigate how those patients were making it in but it was VERY flaky and time-consuming, so I went with the simpler approach of renaming the patients such that the random generator never butts into them.

It also adds a dev dependency on a shared module that my yarn was complaining about - possibly something new since the upgrade to node 20? But it doesn't affect any functionality.

### Checklist

- [x] Code is finished
- [x] Tests are written


stretchkennedy (approved):
Code LGTM and I love the comment 👍 

I'd recommend not merging until the linter errors get fixed. Also guessing that the package.json change might conflict with main?

PR #5042, opened 2023-12-13
Tamanu Web: fix Jest

passcod:
### Changes

- Upgrade to Storybook 7 so it works with Vite


edmofro (commented):
Still in draft and has a couple of linting things, so not approving yet, but pretty much all lgtu (us, reviewing with Daniel during donut time)

dannash100 (dismissed):

PR #5045, opened 2023-12-13
Tamanu Web: Disable codeship

passcod:
### Changes

Disables codeship, but leaves a tiny stub so it remains green — we want codeship to still run on this repo until the metaserver no longer relies on it.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
One question that is non-blocker, but would be good to check if we can get rid of it entirely anyway?

Otherwise looks good!

PR #5048, opened 2023-12-13
Tamanu Web: Rework package scripts

passcod:
### Changes

- Remove obsolete scripts (from codeship era)
- Add `clean:deps` that removes all node_modules


edmofro (changes requested):
Just one question

edmofro (approved):

PR #5053, opened 2023-12-13
Tamanu Web: Clobber electron code

passcod:
### Changes

The "original" tamanu web work, taking an axe to a whole lot of Electron-specific code in the desktop app.


edmofro (commented):
Hey not going to review this atm as I suspect the .jsx renames coming in is a git history issue? 
One question from a glance at the first file though - are we getting rid of storybook?

edmofro (changes requested):
One extra thing needs commenting out, I believe

edmofro (approved):

PR #5054, opened 2023-12-13
Tamanu Web: Delete discovery

passcod:
### Changes

- Delete the discovery code.
- API assumes it's available at the `/api` route.
- Remove all server selection UI.


edmofro (approved):
One non-blocking question, approved!

PR #5057, opened 2023-12-14
Tamanu Web: no more babel nor webpack

passcod:
### Changes

Completely remove babel and webpack except for mobile.


edmofro (approved):
Another non-blocking question

PR #5058, opened 2023-12-14
Allow passing a target to build-all script

passcod:
### Changes

Makes it easy to build one particular package and all its dependencies.


edmofro (approved):
Nice idea

PR #5060, opened 2023-12-14
Tamanu Web: Introduce Vite

passcod:
### Changes

- Add Vite
- Configure it for Tamanu Web
- Build shared packages twice (to both ESM and CJS), to support Vite's requirements and Node's/servers' needs
- A few outstanding electron-related clobbers as informed by Vite complaining about them


edmofro (changes requested):
A few minor things I don't fully understand, so just some clarity on that and this will be good to go

edmofro (approved):

PR #5062, opened 2023-12-14
Unify editorconfigs

passcod:
### Changes

One to rule them all.


dannash100 (commented):
Just to make sure I understand here - the indent size for .sh files would change from 2 to 4 in case of `/scripts/.editorconfig`

dannash100 (approved):
Looks great just tiny question for me really

PR #5068, opened 2023-12-15
NASS-1024: Force single tab on Web

flynncs:
### Changes

Prevents user from opening multiple tabs of Tamanu Web at once.
- Utilises localStorage 

Upon opening an additional tab, the user will be displayed the following screen:

image

[Link to investigation](https://beyond-essential.slab.com/posts/nass-1024-forced-single-tab-l8jq2nk9)

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Nice work with this! Logically this all looks fine, but I think it has a problem where the tab detection logic is split up across App.jsx and index.js in a way that someone coming across this later will likely find quite confusing.

I'd suggest making a new file, `singleTab.js`, which exports functions for those other files to read. Then you'd have...

```javascript
// index.js
import { initSingleTabDetection } from './singleTab';

window.onload = () => {
  initialiseSingleTabDetection();
};

// App.jsx
import { isPrimaryTab } from './singleTab';

if (!isPrimaryTab()) {
  return <MultiTabErrorScreen />;
}
```

That way, when someone comes across it later, it's super clear what it's doing, and it's obvious where to go if more context is needed.

Also, I think the `MultiTabView` should be renamed to `MultiTabErrorScreen` - MultiTabView sounds like it's just a different way of displaying Tamanu across multiple tabs. Also, according to the same principle as above, I think that file could live quite happily in a `singleTab` directory with the actual logic, so that all the singleTab-related stuff is contained in one place. (I know there's an existing pattern of putting lots of views in `/views` but it's VERY crowded in there these days)

mcccclean (changes requested):
Looking great, much cleaner! But I identified a tricky bug with the `localStorage` approach; I have a different suggestion.

PR #5070, opened 2023-12-15
TAV-717: Tamanu Web Production Build

passcod:
### Changes

- Do a vite producton build and pack it to a zip
- Pre-compress all textish assets to gzip, brotli, and zstd
- Include the error pages (placeholder'd for now)
- Sign the zip
- Upload to artifact with retention: 14 days (lifetime of release)
- Upload to S3 for releases, RCs, and main


edmofro (approved):
A few questions but only for my own understanding

PR #5074, opened 2023-12-17
MDEV-215: New report default range past 18 years

NavarroEmilioLuis:
### Changes

A new upcoming DB report needs this new range to make as default


tcodling (approved):
LGTM! Does make me wonder though if maybe we should allow more flexibility in defining a default time range in the report eventually though rather than us having to do a code change every time. I can see potential for all sorts of defaults people might want (1-2 weeks, last year, last 3 months, etc.)

PR #5086, opened 2023-12-19
TAV-913: Tamanu Web to Tester Envs

passcod:
### Changes

- Some cleanups lost in merges or splitups
- Build docker container that serves frontend
- Aggressively precompress files
- Get rid of Sass
- Move all runtime dependencies out of devDeps
- Move vite out of devDeps
- Get rid of old webpack dependencies
- In coordination with https://github.com/beyondessential/ops/pull/44
- Disable vite manual chunking as that was causing a bug in the production build where an object ('shams') was spuriously undefined


edmofro (approved):
lgtm

PR #5090, opened 2023-12-20
TAN-2448: Program registry sync

NavarroEmilioLuis:
### Changes

Okay so I think this is it. After thinking about it for a while, it seems to me that this one should be a patient sync filter and not just a sync filter - the core reason is that we do want to apply the normal patient filter plus added logic, which I think is best represented this way, plus, we already get the `sessionConfig` for free and the sync filter API is not asynchronous.

Also, I've spent quite some time looking at this to see if we could throw in some unit testing but I think that given the urgency, plus the fact that testing sync seems artificial to me, I decided to leave it out!

I'm not completely sure about the setting name, but it seemed appropriate 😅


edmofro (approved):
All looks fantastic, and very happy with the names you've given the setting (and everything else)

PR #5093, opened 2023-12-21
TAN-2295: Fix registry search filters

mcccclean:
### Changes

The core issue was a simpleTextFilter trying to do an uppercase transform on patient.sex, which is currently an enum rather than text. Fixes in this PR:

- transform `sex` to lowercase in JS rather than transforming patient.sex to uppercase in SQL
- fix a performance issue where it was sequencing over the whole db before filtering rather than after (same bug as the notes one!)
- add some (non-exhaustive) tests against patient filtering
- (unrelated) add an explanatory comment to PatientProgramRegistrationConditions

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### DIFF
--- a/packages/lan/__tests__/apiv1/ProgramRegistry.test.js
+++ b/packages/lan/__tests__/apiv1/ProgramRegistry.test.js
@@ -226,5 +226,50 @@ describe('ProgramRegistry', () => {
         },
       ]);
     });
+
+    describe('Filtering', () => {
+      const patientFilters = [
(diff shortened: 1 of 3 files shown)


NavarroEmilioLuis (approved):
I suspect that all mobile tests are breaking atm for some reason, got that on other branches as well (unsure if just program registry branches but def not from this PR).

I think this looks good, just needs to remove a console.log and lint some files 👍 I also find weird the need of the transform (the query also changed, and it's used on patient filters!) but maybe this one is different somehow, and honestly I think it's ok anyway

PR #5100, opened 2023-12-21
No issue: more aggressive yarn cache in CI

passcod:
### Changes

- Cache the content of node_modules in an attempt to cut down on recompute
- Add an explicit check for the yarn.locks being out of date

#### Before

image

image

#### After

image

image

### Checklist

- [x] Code is finished

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,22 @@ jobs:
         with:
           node-version: 20.x
           cache: yarn
+      - uses: actions/cache/restore@v3
+        with:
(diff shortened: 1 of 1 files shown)


mcccclean (approved):
LGTM and what a rad result!

I notice the CI config has become quite verbose, which I feel like might turn into pain later? I wonder if there's some way to get some... modularity or abstraction or something? (uh... in a way that is net less complicated :grimacing:, like a script that generates CI yml files is probably pretty fraught). Not a huge deal though and definitely doesn't impede this particular PR at all, just a thought.

PR #5111, opened 2024-01-08
Merge release/2.0 to main

passcod:
Includes fixes for web frontend related to `process` errors.

### DIFF
--- a/packages/web/app/components/Field/Form.jsx
+++ b/packages/web/app/components/Field/Form.jsx
@@ -1,3 +1,5 @@
+/* global NODE_ENV */
+
 import React, { useEffect } from 'react';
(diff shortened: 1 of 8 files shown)


mcccclean (approved):

mcccclean (changes requested):
Oh wait, the use of NODE_ENV is breaking the linter apparently. Probably that needs a linter config fix to allow it as an undeclared global? 

Also while I'm leaving suggestions - if we do have to add some declaration to each file that reads NODE_ENV, maybe it'd be nicer to keep the that access to be inside a single `isDevelopment()` util function that everywhere else just imports? But I don't feel strongly about that.

mcccclean (approved):

PR #5126, opened 2024-01-09
Fix 1.38 server-packager for Node 20

passcod:
Fixes the packaging workflows.

Then we bump the version of the next release to 1.38.2 (skipping .1) so that we don't overwrite the 1.38.1 servers (which are really the 1.38.0 servers) if/when we need to hotfix something.


NavarroEmilioLuis (approved):
Looks good. Looked up why use `20.x` and seems like a very good idea!

PR #5128, opened 2024-01-09
TAN-2456: Program registry: make deleted registrations not block reregistration

Alastair-L:
### Changes

- The logic was completely incorrect on the first pass, this fixes it.
- Specifically it was looking at all registrations for a blocking one, not just the most recent one

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Hey I'm not sure that this would be the best way, so my ask is that maybe let's revisit that SQL? Let me know what you think. Also there is a duped case in another file, worth checking that out

NavarroEmilioLuis (approved):
LGTM!

PR #5132, opened 2024-01-10
Add hint about needing to remove the rewrite: line

passcod:
### Changes

In tester (and later, production/demo/clone) deployments, the rewrite from `/api/` to `/v1/` is done in the deployment routing itself, so doing it in vite breaks the config. In development, there's no such routing layer, so it's necessary.


NavarroEmilioLuis (approved):
This looks good but maybe copy change would be better?

PR #5138, opened 2024-01-11
Fix: pad data not immediately updating

tcodling:
### Changes

We were having slightly strange behaviour where on the release/2.0 branch the pad data was not updating immediately on save. You actually had to move away from the screen and back again in order to show the new data. When testing on my local, it seemed that it would work about 50% of the time. 

On some further investigation and testing it seems like this was caused by a race condition where the queries weren't being invalidated before the patient reload was triggered. Still not exactly sure why we had slightly different behaviour on the release branch vs the local deployment though 🤔 

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


tcaiger (approved):
Nice one!

PR #5144, opened 2024-01-11
Fix/pdf printing storybook

tcaiger:
### Changes
- Fix storybook build
- Setup Storybook for Print PDFS

### Checklist

- [x] Code is finished
- [ ] Tests are written

Screenshots
image


dannash100 (changes requested):
I think we should follow the [linked](https://storybook.js.org/docs/builders/vite) guide and override vite settings with the storybook specific changes

PR #5146, opened 2024-01-11
EPI-675: Add new reference data type

NavarroEmilioLuis:
### Changes

Just adds new reference data type "contact relationship". Should be importable/exportable. Validation: id, code and name should be unique.


Alastair-L (approved):
LGTM!

PR #5147, opened 2024-01-11
TAN-2287: Fix how survey answer names are displayed

mcccclean:
### Changes

When displaying a completed survey response, answers for village and programRegistryClinicalStatus questions were showing up with the bare IDs rather than the names.

The issue was that the display logic was using an object to track which field to look at for displaying an answer, and defaulted to ID if a model wasn't present in the array. Rather than add `village: name, programRegistryClinicalStatus: name` to an array that was 99% `name` already, I updated the logic to just default to `name` instead, and only fall back to ID if name isn't present. This simplifies the logic a bunch and should prevent this issue from coming up again when we add new models to the logic.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


Alastair-L (approved):
Really clean - nice!

PR #5151, opened 2024-01-12
Trigger CD Down on unlabeling

passcod:
Currently a PR is un-auto-deployed when it closes only, and [there's weird interactions with the github behaviour that means that most of these CD Downs get silently cancelled.](https://github.com/orgs/community/discussions/26657)

Separately, there's a behaviour where adding the `auto-deploy` label deploys a PR, but removing it _doesn't_. This was meant to provide flexibility, being able to control what gets deployed, for example pushing a flurry of review changes to a PR but leaving it in a broken state, then once completely fixed getting the auto-deploy updated again.

However, this has proved confusing. There have been many instances of devs confused that their branches didn't undeploy after removing the label, or worse not noticing that the branch didn't undeploy and blaming something else. There's also a greater need to more easily control undeploying, and removing the label is a decently intuitive UI.

So this PR enables this. It also removes the restriction that draft PRs can't auto-deploy, so now control is purely with the tag. Finally, it removes the `workflow_dispatch` triggers, as those never did work: AWS can't authenticate for `workflow_dispatch` runs, only pull requests and branch pushes.

### DIFF
--- a/.github/workflows/cd-down.yml
+++ b/.github/workflows/cd-down.yml
@@ -1,7 +1,6 @@
 name: CD Down
 
 on:
-  workflow_dispatch:
   pull_request:
     types:
       - closed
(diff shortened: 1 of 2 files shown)


mcccclean (approved):
LGTM, one optional suggestion

PR #5154, opened 2024-01-14
Fix: Broken long notes

tcodling:
### Changes

It was discovered that some notes were showing up with nothing in the content like this:
image
On investigation I discovered that this was a result of an infinite loop that was bugging it out and making the whole table row render nothing. I moved this to a useEffect that listened to suitable variables in order to stop this. Not sure if this is new with tamanu web or not

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


dannash100 (approved):
This looks good - just really need to test that lazy loading works correctly with larger note sets - when various notes are expanded and unexpanded.
Mabye just add to testing notes.

PR #5159, opened 2024-01-15
feat(vaccines): NASS-985: Add ability to exclude vaccines from certificate

dannash100:
### Changes
Add hideFromCertificate column to schedule vaccines as discussed in recent tech design, this is a pretty specific & rigid solution - aiming to be as simple as possible to match our current requirements for this feature.

I also went for filtering by the boolean the the FE after the query - this is because both VaccineCertificate and Covid19VaccineCertificate use the same query and search params - and then the covid cert FE filters out by certifiable  `const vaccinations = vaccineData?.data.filter(vaccine => vaccine.certifiable) || [];` so basically just mirrored this in the introduced filter for vaccineCertificate.

Tests somewhat N/A here as importer exporter just use basic model shape and work out box really.

Testing sheet
[vaccination-hide-from-certificate-examples.xlsx](https://github.com/beyondessential/tamanu/files/13934672/vaccination-hide-from-certificate-examples.xlsx)

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


tcodling (approved):
this is so crispy. i think the filter on the frontend like the covid cert was 💯 the way to go

PR #5171, opened 2024-01-16
Fix all lint warnings (except mobile)

passcod:
### Changes

A lot of these are trivial. Some are not. One commit per package.

Mobile is excluded because it has over 500 warnings so I want to do it in a different PR at a later time (or leave it to someone else).

We can now (and this PR does) switch these from warnings to errors:

https://github.com/beyondessential/tamanu/blob/8be3f3cc17136d9617107d1e0bebb51d5ab5621c/eslint.config.js#L64-L70

### Callouts

These are things I did fix that I think are great teaching examples for the kind of things that these lints catch:

#### Explicit field deletion due to name conflict

Before:
```js
export const SurveyResponseSelectField = ({ field, patient, options: _, config, ...props }) => {
  const api = useApi();
  const [options, setOptions] = useState([]);
};
```

"Okay, so we're taking `options` out, but naming it `_`, and then `_` is linted as unused. _why_ is it being renamed like this? Oooh, because we define a `const options` later. Yeah, we can make this obvious and explicit instead."

After:
```js
export const SurveyResponseSelectField = ({ field, patient, config, ...props }) => {
  delete props.options;

  const api = useApi();
  const [options, setOptions] = useState([]);
};
```

#### Hidden potential race condition

[In slack](https://beyondessential.slack.com/archives/CDGS6P56K/p1705365781032499)

```js
async function createPanelLabRequests(body) {
  const { panelIds, ...labRequestBody } = body;
  const panels = await getPanels(panelIds);

  const response = await Promise.all(
    panels.map(async panel => {
      const testPanelRequest = await createPanel(panel.id);
      labRequestBody.labTestPanelRequestId = testPanelRequest.id;

      // potentially more async work

      const newLabRequest = await createLabRequest(
        labRequestBody,
        // ...
      );
      return newLabRequest;
    }),
  );
  return response;
}
```

This is a really cool newish lint. It highlights this line:

```js
labRequestBody.labTestPanelRequestId = testPanelRequest.id;
```

Now, the code as it exists _doesn't_ have a bug. Where I've put

```js
// potentially more async work
```

There is in fact _no_ async work. If there _was_, what could happen is that in the yield point between

```js
labRequestBody.labTestPanelRequestId = testPanelRequest.id;
```

and

```js
const newLabRequest = await createLabRequest(labRequestBody);
```

...the variable `labRequestBody`, which is from the scope _outside_ the inner async function, could have its `labTestPanelRequestId` property changed by another instance of that async function, and now your code has crossed the streams 👻 and you have spooky bugs.

(If you know a bit of Rust, this kind of manual flow and scope analysis is what the compiler does for you. If you think of `labRequestBody.labTestPanelRequestId =` as requiring a mutable reference to `labRequestBody`, then the code as it is can't compile, because it would require multiple concurrent mutable accesses and that's exactly what Rust's ownership model prevents.)

Now, as I said, the code currently _doesn't_ have async work in between these two points, and because Node.js is single-threaded, that means that the value written to that property cannot be overwritten concurrently, and therefore there's no bug.

But critically, this is _very much not obvious_. If someone comes to this piece of code, and they need to do some async work there, I would bet they would think nothing of putting an `await` there. It's completely unintuitive that adding an `await` here introduces a bug, regardless of what is being awaited.

### Checklist

- [x] Code is finished
- [x] Tests pass


mcccclean (changes requested):
Looks good overall! Keen to get this merged quick as it touches a lot of files.

The only pending thing is [per this slack comment from Edwin](https://beyondessential.slack.com/archives/CDGS6P56K/p1705522652468549?thread_ts=1705356949.203389&cid=CDGS6P56K) - maybe we want to find/replace the `const a = Object.assign({}, b)`s with `const a = { ...b }`s.

edmofro (approved):
Codeowner aspects all lgtm. Thanks @passcod this was a big effort to get us clean on the linter!

mcccclean (approved):

PR #5176, opened 2024-01-16
Refactor vaccine modal multiform for clarity

passcod:
### Changes

This came out of a single lint (which warns on the absence of `key` when iterating in JSX). This file was very confusing to me, so much so that I couldn't figure a good candidate for that key value. This is my attempt at simplifying the file, optimising for human clarity, without changing anything about what it ultimately does.

### Checklist

- [x] Code is finished
- [x] Tests pass


tcaiger (approved):
Looks much more readable to me. I'm still not a fan of iterating on large objects to render templates. I would rather just repeat out the template code a bit to make it more readable but this is a great improvement.

PR #5186, opened 2024-01-19
NASS-1091: Properly implement query on vitals form

tcodling:
### Changes

We had a bug where the vitals form was permanently staying in a loading state. It actually turns out that this is because the pad query on this form NEVER actually worked as it was never supplied a patientID. This made the isLoading value always show up as true and therefore never displayed the vitals form

Only after Daniel tweaked the query to not run when not supplied a patientId did this bug actually show itself. So the fix for this bug is actually just properly implementing the pad data query

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


tcaiger (approved):
Nice one

PR #5188, opened 2024-01-19
EPI-726: Use Frankenstein date for procedure table

NavarroEmilioLuis:
### Changes

Surprisingly self explanatory 🤔 here is the context from Linear, just in case we ever switch:

> Okay so the problem here is the way we handle these fields. I'm not sure why we are dealing with separate date - time fields but I'll take it that is necessary. Now, the problem here is that the time fields (start time and end time) will always* keep the current date, meaning that will always come out as wrong.
> 
> If the date and start time fields are not modified, then everything will display good in either version. That is because both are given the current date time as their initial value.
> 
> I think the possible solutions should be:
> 
> - Unite this two fields for the table and tooltip
> 
>   - This could happen from the backend or the frontend
> 
>   - A bit hacky, but actually adjusts to the form model of having two separate fields
> 
> - Modify the form data to adjust both fields
> 
>   - Nice to have "accurate" data in the database
> 
>   - Would make the two fields duplicate
> 
> Creating an extra field on the database is simply not an option, I already deeply disagree with having two different fields keeping date time. That said, that's the UI and I think it's going to be better to more accurately match it so first option it is. I'll look into where is more plausible and apply it.
> 
> PS: backend uses a helper, so frontend it is.
> 

*Unsure if that's the case, wonder what would happen if 00 either way is selected or if you could keep scrolling, but that doesn't matter much.


Alastair-L (changes requested):
LGTM but 3 notes:

1. Looks like there's still a sneaky console log in there

2. I'd much much prefer adding some more context in the comment itself, instead of saying "blame this". If there isn't any additional context required - then I reckon just remove the "blame this" (because people will know to do so if they're stuck, and having the comment implies they should do it even if they do understand what's going on). 
Finally - if you do want to create a link outside the codebase, I'd much prefer an actual link, i.e.:
```
Additional reasoning: https://github.com/beyondessential/tamanu/pull/5188
```
However, I don't reckon you need any of that as it's fairly straightforward.

3. Just a sanity check that you have considered the option of just storing one date in the backend and doing the frankensteining in the set form instead of the view? All good if you'd thought about it and discarded it, just wanted to bring it up


The only blocking comment is the console log, the others are up to you

Alastair-L (approved):
Actually I'm going to preapprove just in case I leave before I can ;(

Make sure to fix up the console log though!

PR #5189, opened 2024-01-19
No issue: Skip failing test

Alastair-L:
### Changes

This failure is only relevant when the question is misconfigured, so it should be fixed but shouldn't block release IMO

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (changes requested):
Did some investigation as to exactly why it broke - agreed that it's safe to ship with the test disabled but have update the logic to clarify what's up to the next person to come across it.

PR #5195, opened 2024-01-21
Fix Server Start Dev Scripts

tcaiger:
### Changes

The server `start-dev` script starts two watches concurrently. The first is `build-watch` which uses swc to build the app source to the `dist` folder. The second is `start-watch` which restarts the server when there are changes in the `dist` folder. The `start-watch` was just re-running `yarn start` and causing an error for Port already in use. This change updates the `start-watch` to use nodemon so that the server gracefully restarts.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "eslint-plugin-react": "^7.33.2",
     "eslint-plugin-react-hooks": "^4.6.0",
     "globals": "^13.24.0",
-    "nodemon": "^2.0.19",
     "prettier": "^1.17.1",
     "yaml": "^2.2.1",
     "yarn": "^1.22.18"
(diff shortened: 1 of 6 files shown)


NavarroEmilioLuis (changes requested):
Hmmm you missed meta server, which if resolved I believe can help with removing `chokidar` from our deps 🤔 

Still, this works but the logging looks messy. Having a closer look it seems we're building shared what... 3 times before starting the thing? I think we should also refactor `watch-shared-and` to avoid triggering subsequent builds.

PS: but definitely thank you for tackling this! I think the first one should be the change request and the second thing is just a nice to have?

NavarroEmilioLuis (approved):
Thanks for taking the time to check on it, I think this is good enough! Just one final check but can pre approve!

PR #5210, opened 2024-01-22
NASS-1090: Use new fonts for khmer language

tcodling:
### Changes

Adding in a Khmer specific font. In order to set a font for only khmer characters i defined the font to only apply to the khmer unicode range. We prioritise that font and then go to roboto if the text isnt a Khmer character. This may potentially need to be made more dynamic down the road if/as we add more languages but this is fine for our current use case

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


tcaiger (approved):
Ka Pai!

PR #5217, opened 2024-01-23
EPI-665: Create new patient channel model

NavarroEmilioLuis:
### Changes

Creating new table in both server and mobile!


mcccclean (approved):
LGTM! 

One suggestion, doesn't need a re-review, but it would be good to add a comment to the mobile model on the details field to indicate that it's actually JSON.

PR #5218, opened 2024-01-23
TAN-2431: Program registry add mobile backend models

Alastair-L:
### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Okay so there is still some way to go here. A lot of changes needed but most are more or less the same: table names are wrong, a small fix to the `deletedAt` field, possibly drop the foreign key relationships or update their `referencedTableName` to match the actual table name, etc.

Won't comment on the types nor the model files because those would have to be modified accordingly, too. Please check about the references and if they would work or not without having them defined in the database - these do work or are not needed in the server. But maybe we do want them here, I don't know!

mcccclean (commented):
Just leaving some followup comments on Emilio's review :pray:

NavarroEmilioLuis (changes requested):
Thanks for going through the changes. I have a couple of other checks to make in the meantime. I'll also start reviewing the models now, shouldn't get in the way anymore

NavarroEmilioLuis (changes requested):
Alright some changes for model/type. I won't go into the others as they might have similar things, you get the idea 🙏

NavarroEmilioLuis (changes requested):
Okay almost there. I found there are still some discrepancies on models vs mobile migrations (I think just one but can't fully remember). In this case I'm not sure whether we want to drop the constraint or add it, please check what makes more sense for the model.

I've reopened change requests that weren't addressed, you can see those above. Also please get rid of the decorator type in all models.

i.e. `@ManyToOne<type>()`

NavarroEmilioLuis (changes requested):
Thanks for grabbing the things, we're almost there, just two more change requests that were missing from last time

NavarroEmilioLuis (approved):
Thank you very much for grabbing those! Good to go!

PR #5222, opened 2024-01-23
No issue: Expose getNameColumnForModel function mobile

Alastair-L:
### Changes

From this slack convo: https://beyondessential.slack.com/archives/C03KJSEPV9A/p1705980423326189

I also don't know if this is the "right" way to do this - but I think that's out of scope for this fix?


mcccclean (changes requested):
Good approach - I agree that returning it as a string is a little weird but yeah it's the cleanest way to accommodate the autocomplete field. LGTM, just a note about the comments.

mcccclean (approved):

PR #5224, opened 2024-01-24
feat(fhir): SAV-521: Allow configuring disable materialisation fhir resources

chris-bes:
### Changes
- SAV-521: Allow configuring disable materialisation fhir resources

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Thanks for doing this work... checking the `resourcesThatCanDo` update I was thinking to myself that we've only discussed materialistation and this covers it, but would we ever want to also disable other interactions? Given the urgency of this, I don't think it should be a blocker.

There is one change I think would make life easier though

NavarroEmilioLuis (approved):
Thanks for grabbing that, good catch on the test.json5 file!

PR #5229, opened 2024-01-25
NASS-1081: Mobile settings usage

dannash100:
### Changes
- Into epic/db-defined-config
- Swap out localisation context with settings context
- Make getSettings use a generic type so that we can pass in the type like `getSetting<boolean>` rather than maintain multiple utility functions like `getBool` and `getString` I feel this is an optional part of this pr but I like it.
- Fix a few random types

### Checklist
- [x] Code is finished
- [x] Tests are written

_Upon merging:_


tcodling (changes requested):
there might be some logic i missed that accounts for tihs but seems that there are a few places that call `field.xxx...` instead of `localisation.field.xxx...`

tcodling (changes requested):
still quite a few that are missing the localisation. prefix. Noted "here" on each one

tcodling (approved):

PR #5232, opened 2024-01-25
Fix: Update mobile API_PREFIX for local dev

tcaiger:
### Changes

Update mobile api prefix to point at `v1` on local dev and `api` on production.

The API is served from the /v1 path, but on production the url needs to go through the /api path to be rewritten by vite config to the correct path. The server will be updated soon so that the API is served from /api after which this condition won't be necessary

See more details here
https://beyondessential.slack.com/archives/GE9NHB95F/p1706157759491369

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


mcccclean (approved):
LGTM with an optional comment suggestion.

PR #5238, opened 2024-01-28
EPI-726: Fix procedure table sort

NavarroEmilioLuis:
### Changes

Take 2... This approach is just a short-to-mid-term fix, as there are discussions to change the procedure form in the future later this year. Here we will just adjust newly submitted procedures to match the current workflow: the date selected should also reflect on the startTime and endTime fields - with an extra heuristic that a procedure might end the next day it was created.


tcodling (approved):
LGTM. Agree its a bit of a quirky fix with all the slicing but seems fine given the explanation and plan for changes this year.

PR #5247, opened 2024-01-30
TAN-2320: Include all keys in validation

Alastair-L:
### Changes

Fixes the issue in this thread:
- https://linear.app/bes/issue/TAN-2320/program-registry-validation-steps-when-importing-program-registry#comment-86c2b8d2

Also note there'll be another PR removing the use of `isAdditionalDataField`

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (approved):
Looks good overall! I think that we could have two variables with partial application so we don't declare the allowed locations inline... But looks fine this way really, so up to you - I think it aids in the intent, having a `readablePatientDataColumnString` and a `writablePatientDataColumnString` (what a mouthful).

PS. I realize that this comment can be very cryptic. So here is a clearer example:

Instead of having to inline:

```
column: patientDataColumnString([
  ...Object.keys(PATIENT_DATA_FIELD_LOCATIONS),
  ...Object.values(READONLY_DATA_FIELDS),
]),
```

We can do:

```
const readablePatientDataColumnString = patientDataColumnString([
  ...Object.keys(PATIENT_DATA_FIELD_LOCATIONS),
  ...Object.values(READONLY_DATA_FIELDS),
]);

// ... some code
column: readablePatientDataColumnString(),
```

or even:

```
const readablePatientDataLocations = [
  ...Object.keys(PATIENT_DATA_FIELD_LOCATIONS),
  ...Object.values(READONLY_DATA_FIELDS),
];

// ... some code
column: patientDataColumnString(readablePatientDataLocations),
```

NavarroEmilioLuis (approved):
Thanks! Like your names better tbh!

PR #5096, opened 2023-12-21
TAV-939: Detect non-deterministic migrations

stretchkennedy:
### Changes

Adds an automated test to detect non-deterministic migrations by running them twice and seeing if the result is different (minus meaningless tables and columns, like `pg_class` or timestamps).

The work necessary to finish this component of the card is to keep adding generated data until the tests pass, debug any problems as you go, and test it more thoroughly than I have against non-deterministic migrations.

### Checklist

- [ ] Code is finished
- [ ] Tests are written
- [ ] UI Screenshots added to the Linear issue
- [ ] Testing notes added to the Linear issue

_Upon merging:_

- [ ] Update the changelog (find it [here](https://github.com/beyondessential/tamanu/releases) in the appropriate _draft_ release)
  - [ ] with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)
  - [ ] any config/settings changes and manual deployment steps
  - [ ] any db schema or other changes that could impact downstream data analysis
- [ ] Update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly) or [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q) as needed
- [ ] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed

PR #5257, opened 2024-01-30
KAM-33: Add CI workflow to generate data and dump a copy of the resulting db to S3

rail-rain:
### Changes

This is a part of an effort to make a better version of #5096. It uses the code from the original PR to generate fake data, makes a dump of it and upload it to S3.

Open question: how and where to house the data generation script.

### Checklist

- [x] Code is finished


passcod (changes requested):
That looks great, I have just a few notes :)

passcod (approved):

PR #5258, opened 2024-01-31
EPI-672: reminder messages add reminder contact modal

BhargavDY:
### Changes

Please note that the design is done, but API integration still remains, so this ticket is partly completed.

image
image

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Hey, I know this took some work but it definitely needs some changes to adjust more to our codebase. There is also some design change requests here and there but I think they'll help more than not - unfortunately, I think that if applied correctly, the updates to `ButtonRow.jsx` and `ModalActionRow.jsx` are not going to be needed at all.

NavarroEmilioLuis (changes requested):
I want to start saying good work, I really think this iteration is far better than the first one, I think it's super great idea that you created a different component for the table bit, however I think that we could improve things a little bit.

I understand the work that has been put on this so feel free to push back if you disagree with some changes. Overall, I think we should shy away from the presentational/container pattern (plus actually just shy away from using container lightly to avoid being misleading - most of these are not really containers). But my main concern here is that we are spreading logic between components, where we could have all logic in one place! I'll leave more detailed notes where things are relevant.

Let me know if anything is unclear, mostly it will be renaming and moving stuff around, it should make this simpler!

NavarroEmilioLuis (approved):
Thanks for grabbing al the changes, I think everything major is addressed and all looks good. This was a pretty quick turnaround so kudos!

PS: I left one comment but don't think a re-review is necessary. You can get rid of it if you want or not, your call! (unless I'm mistaken and it is actually needed).

PR #5259, opened 2024-01-31
KAM-127: Add /api as duplicate of /v1 route

passcod:
### Changes

- Change all references of /v1/ to /api/
- Remove the rewriting in vite
- Add a route to central and facility duplicating the /v1/ route to the /api/ route

This immediately makes the vite rewrite and confusion therein obsolete, and should also make the switching for mobile unneeded.

The reason for a duplication rather than straight replace is to leave some time for other components (notably the tester envs, which need a config change to switch to _not_ rename the /api path when routing to backends) to flexibly move, rather than coordinate a synchronous change.

A larger question is should we do the final "delete /v1" _tomorrow_, once the tester deploys have moved, or should we leave that duplication in place for the medium term, so that integrations that call our API can move at their own pace rather than break when we deploy 2.x.

### Checklist

- [x] Code is finished
- [x] Tests pass

_Upon merging:_

### DIFF
--- a/docs/Surveys.md
+++ b/docs/Surveys.md
@@ -27,7 +27,7 @@ machine.
 
 ```
 ###
-POST http://localhost:4000/v1/admin/importSurvey
+POST http://localhost:4000/api/admin/importSurvey
 Content-Type: application/json
 
 {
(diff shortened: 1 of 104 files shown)


edmofro (approved):
lgtm, just one stylistic comment

anilvkaran (approved):

PR #5274, opened 2024-02-04
NASS-1122: Fix - survey response regression

tcaiger:
### Changes

The bug was on Surveys that were submitted with PatientData questions that were configured Autocomplete components. They were set with empty strings as initial values and this causes an error when trying to view that answer because the survey response endpoint tries to look up the submitted answer but it doesn't exist. This fix is both on the form submission side to change the default value to null and on the submission view side to not error if an autocomplete answer has an empty string submitted.

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


tcodling (approved):
LGTM once you get rid of the .only

PR #5281, opened 2024-02-05
EPI-726: Fix procedure table

NavarroEmilioLuis:
### Changes

This fixes the table sort issue and also the display on the tooltip.

Same as #5188 but against release/2.1


tcaiger (approved):
Yikes pretty spooky but looks like the best option for what we are working with. I had a quick look but I'm mostly trusting Alastair's review for this

PR #5287, opened 2024-02-06
Fix main: Restore checkServerAlive method

passcod:
### Changes

The `checkServerAlive` method went missing in the API client PR.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
Ahhh, nice catch!

PR #5299, opened 2024-02-08
2.1 fix: watermark causing vaccine certificate to crash

dannash100:
### Changes
In certain cases the watermark exceeded view width and height values could cause the  vaccine certificate to never load.
I debugged this by creating a pdf worker to take the pdf render off the main thread - and noticed that while it did not outright crash the app - it still never loaded.

After reading around it seems there can be some oddities with images exceeded the content window causing a full crash. [Similar issue report](https://github.com/diegomura/react-pdf/issues/2242)
Doing this fix of limiting the size of the fullbleed watermark to view size resolved the issue.

Here is the draft pr for worker
[WIP: PDF rendered in worker](https://github.com/beyondessential/tamanu/pull/5300) - although it needs much more work than is appropriate for 2.1 regression 🙏 

I checked for visual regressions here and there is no change from before except it doesn't crash in the case identified by testers

### Checklist

- [x] Code is finished
- [n/a] Tests are written

_Upon merging:_

### DIFF
--- a/packages/shared/src/utils/patientCertificates/Layout.jsx
+++ b/packages/shared/src/utils/patientCertificates/Layout.jsx
@@ -51,6 +51,8 @@ export const styles = StyleSheet.create({
     bottom: 0,
     minWidth: '100%',
     minHeight: '100%',
+    maxHeight: '100vw',
+    maxWidth: '100vw',
     backgroundColor: '#aaaaaa',
     opacity: 0.05,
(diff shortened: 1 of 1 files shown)


tcaiger (approved):
So random. I can't believe you tracked this down. You're a genius. I'd love to see that worker update get finished. Could you make a ticket for it and put it in the refinement queue.

PR #5300, opened 2024-02-08
perf(pdf): NASS-1234: Ability for PDF to be rendered in worker and implemented for vax cert

dannash100:
### Changes
- An opt in system for rendering pdfs in worker
- This does not block main thread and leads to big performance increase in cases where pdf is rendered in background (before its opened)
- This is visible when navigating to vaccine page so I utilise here
- Add `perf` to conventional commit regex and contributer info

**I have to make some exceptions for prod build Hold on**
Ok sick i sorted it - just need to make sure it skips shim in storybook

Render pdf off main thread using web worker adapted from [doc linked blog post](https://dev.to/simonhessel/creating-pdf-files-without-slowing-down-your-app-a42)

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


tcaiger (approved):
Love it!

PR #5316, opened 2024-02-12
ci(packaging): Fix missing permission and improve for manual runs

passcod:
### Changes

- There was a missing `contents: read` permission which prevented the workflow from running
- As we can't upload to AWS with `workflow_dispatch`, switch in this case to upload it as an artifact (kept for 15 days aka enough for the release cycle)
- Use the bestool to sign the file instead of the previous complicated minisign/rsign setup

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
This LGTM but I'm not familiar with `bestool`, obviously is something done by us and I assume written by you so secure enough but just wanted to raise extra awareness about that bit mostly being on your own 😅 also, the retention policy: we only need that artifact for internal deployments? Or will they get pushed to prod? Cause then it seems too short?

Approving cause I don't think changes are needed but would be good to leave a trace of these questions I think

PR #5317, opened 2024-02-12
perf(print): NASS-1243: prevent printouts triggering unnecessary expensive renders and clean-up there loading

dannash100:
### Changes

I am going to auto deploy this and get it tested by those who have reported these really slow pdf loads before merging this one.

**Issue 1**:
There was a major issue where react-pdf could end up queuing up multiple expensive renders of the same printout - this could then be further exasperated by `keepMounted` printouts firing each time a seemingly unrelated printout was opened. I.e on vaccine pane when opening `Id label` printout - it would queue up a render for Vaccine certificate and Covid vaccine certificate along with the id label one.

This could lead to load times of up to 10-20+ seconds given the right circumstances.

**Issue 2**: 
Loading was a bit all over the place for pds leading to multiple flashes of different loaders and some niche useEffect timeout to fake a 2 second load to smooth out.
There was also code to account for the drip of data - causing re-renders as any of the prerequisite data is populated.

**Solutions:** 
- Incorporated the preparation of data (useQueries isLoading state) into the pdfViewer component and utilised `instance.loading` to display the same loading spinner during the actual render of the pdf.
- I also made it so that the pdf only need to be rendered once after all the required data is ready, and no longer need to rely on a refresh pdf useEffect.
- I have removed the keepMounted from pdfs due to the fact that because they are always mounted it lead to many extra renders due to the pdfs not being unmounted from the dom which was leading to weird interactions and side effects in react pdf. (leading to a increasingly laggy opening delay the more times pdfs are opened)

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### Deploy


tcaiger (approved):
thanks so much for doing this - it is soooo much better. Just left a couple of trivial suggestions but I'm keen to approve it and let you make those little changes.

PR #5322, opened 2024-02-12
ci(migrations): No ticket: Avoid hitting rate limits with GH API due to KAM-37

rail-rain:
### Changes

This uses `git log` over GitHub API by fetching more with `actions/checkout`. The depth of 4000 is enough to cover one to two year of commits. (Fetching everything is possible, but I think it's overkill.) It's still not clear to me if fetching more is more efficient, but it seems to be true that GH imposes much more relaxed rate limits on cloning than the API.

### Checklist

- [x] Code is finished


passcod (approved):
🎉

PR #5325, opened 2024-02-13
feat(translations): NASS-1101: TranslatedSelectField component

mlie788:
### Changes

Created a TranslatedSelectField component to convert select field options to TranslatedText. This PR also includes refactoring of outdated options found along the way.

Instances and places where TranslatedSelectField is incompatible with the use case of SelectField or just not replaced:
- Disabled select fields (displays as [object Object])
- Places where select fields are dynamically set (e.g. PatientField in PatientDetailsForm) where the prefix can't be determined
- Components where the base of the component is SelectField
- Parts of Tamanu where I couldn't visibly see the use of the SelectField (program registries, report editor db schema field, ...)

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


tcodling (approved):
edit: accidental review i cant seem to delete 😆

tcodling (changes requested):
So in general this looks really good. The only real thing i want to look into changing is the importing of the SelectField. It was previously imported from ../Field and has been swapped out with a new import from another directory.

I think we should export the new TranslatedSelectField from ../Field and then make sure that any usages that we want to stay with old behaviour are renamed to BaseSelectField (like already in this pr)

Michelle wont have time to implement these changes as EOD on her last day so might talk this through with the others when in office

tcodling (changes requested):
So in general this looks really good. The only real thing i want to look into changing is the importing of the SelectField. It was previously imported from ../Field and has been swapped out with a new import from another directory.

I think we should export the new TranslatedSelectField from ../Field and then make sure that any usages that we want to stay with old behaviour are renamed to BaseSelectField (like already in this pr)

Michelle wont have time to implement these changes as EOD on her last day so might talk this through with the others when in office

tcodling (approved):
LGTM! Just those linting fails

PR #5330, opened 2024-02-14
SAV-541: Return records last updated in passed window to Medici Report

chris-bes:
### Changes
- Return records last updated in passed window to Medici Report

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
I understand most of the query is the same, but you just added the filter on encounter ID on all CTEs and main query which sounds great. We're also adding the changes to stop reading from notes and just use encounter history which seems good to me as well.

There are a couple of things here and there that I'm not sure but asked a question, and there are two FHIR methods that probably need change from what I understand

NavarroEmilioLuis (approved):
Looks good to me, just one minor thing in the test config that seemed off but everything else looks just great. Good catch on the trigger!

anilvkaran (approved):

PR #5331, opened 2024-02-14
Fix: fix api errors showing

tcaiger:
### Changes

The TamanuApi class was recently refactored and split out into a new TamanuApi package. In the new extractError method the way that errors are parsed and thrown means that the custom errors that are passed back to the client version of Tamanu Api are different to what they used to be and one of the key changes is that they didn't have the error status included. This change adds the status to all custom errors using the existing mapping of errors and statuses and then updates the client version of the api class to read those statuses.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


dannash100 (approved):
Good thinking 99

PR #5285, opened 2024-02-06
fix(mobile): EPI-706: Mobile navigation issues

BhargavDY:
### Changes

All 3 issues raised by testing team are resolved. These were the issues;
Issue 1 - The patient list was not loading at all, even after the initial sync was finished.
Issue 2 - Recently viewed patient was not visible on the home tab.
Issue 3 - "Start visit" was not bringing the user to the patient screen after adding a new patient.  

### Checklist

- [ ] Code is finished
- [ ] Tests are written
- [ ] UI Screenshots added to the Linear issue
- [ ] Testing notes added to the Linear issue

_Upon merging:_

- [ ] Update the changelog (find it [here](https://github.com/beyondessential/tamanu/releases) in the appropriate _draft_ release)
  - [ ] with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)
  - [ ] any config/settings changes and manual deployment steps
  - [ ] any db schema or other changes that could impact downstream data analysis
- [ ] Update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly) or [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q) as needed
- [ ] Update the [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c) as needed

PR #5335, opened 2024-02-15
fix(mobile): EPI-706: Hotfix mobile navigation issues

NavarroEmilioLuis:
Getting this work to 2.2 that's all, already approved in #5285


tcaiger (changes requested):
I saw that you (@NavarroEmilioLuis) acknowledged that this approach is a bit hacky but necessary to fix the bug so I'm happy to approve the general approach based on your comment there. I left a comment about the useEffect dependencies which I think needs updating. So other than that I'm happy to approve.

tcaiger (approved):

PR #5342, opened 2024-02-18
fix(web): EPI-731: Update vaccine view field groups accessor (hotfix to release/2.2)

NavarroEmilioLuis:
These changes didn't make it into the release just getting them there


tcodling (approved):
sweeet

PR #5345, opened 2024-02-19
ci: Add alls-green job to non-determinism test so it can be added to requirements

passcod:
### Changes

Otherwise we have one job per postgres version, and would need to change requirements when adding/removing versions.

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
LGTM left two questions for you to consider

PR #5347, opened 2024-02-19
feat(central): SAV-546: create encounter service provider

kujo4pm:
### Changes

We need to add Facility information to Encounters for the Senaite integration. While they should get enough from the `Encounter.serviceProvider` this has to be a reference and thus we should include a Organizations to maintain referential integrity. 

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
Hey this looks great, left a couple of questions here and there

NavarroEmilioLuis (changes requested):
Hey man, looks good. I think you have a personal workspace linting rule that rolled back the style changes but I'm ok with that, really. If the repo is not complaining then it's good enough for me.

Now, requesting a change cause this approach won't work with the latest change, the test is breaking 😬 I left a comment about how you might solve it

NavarroEmilioLuis (approved):
Yeah that's the stuff!

PR #5353, opened 2024-02-20
feat(facility): KAM-31: Allow scaling facility API server horizontally

passcod:
### Changes

- Add `startApi` (which serves the HTTP API) and `startTasks` (which runs scheduled tasks) commands, and modify the default to be `startAll` (which does both).
- Change the pm2.config to run one `startTasks` process and N `startApi` processes.

This will make it easier for Cambodia to have their "one facility across two servers" setup, and also allow us to do no-downtime restarts of facilities (for troubleshooting and config changes etc).

### Checklist

- [x] Code is finished


tcaiger (approved):
All makes sense and looks good

PR #5355, opened 2024-02-20
fix(web): Select parameter field bug

tcodling:
### Changes

This was quite interesting. At some point between 1.38 and 2.1, we decided to destructure options and not supply it to the parameter field. I believe this means that no options for report parameter fields should have worked since this change. Felix came through and altered the format for linting purposes but in reality it should just never be destructured.

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


dannash100 (approved):
This looks sensible to me 💯

PR #5361, opened 2024-02-20
fix(docker): KAM-126: Serve web on all interfaces

passcod:
### Changes

Fixes a misconfiguration of the web container. No idea how it works in ECS, it shouldn't... but anyway, this fix is required for K8S.


dannash100 (approved):
Looks gud!

PR #5363, opened 2024-02-20
fix(provisioning): Handle file argument correctly

passcod:
### Changes

- Fix the `provision` command, which was erroring on "should be a path, got undefined" because the way arguments were passed was wrong
- Add support for JSON5 for provisioning (doesn't add a dependency as it's a dep of `config`)
- Change the default provisioning files to JSON5 (as it's more intuitive than KDL, and used by config)

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
LGTM

PR #5365, opened 2024-02-21
feat(fhir): SAV-545  include specimen retry

kujo4pm:
### Changes

The way that the FHIR API handled `_include`s has been modified. It was broken. This should fix it as least for the use cases tested within this PR. 

I have branched off [feature/tpd-141-add-specimen-resource](https://github.com/beyondessential/tamanu/tree/feature/tpd-141-add-specimen-resource) because this uses code for Specimen included there.

The approach I took was to make the tests first to prove it was broken - then propose the fix. Now the tests pass.

Previously it could only handle single level property references of a FHIR resource - think `ServiceRequest.subject`. Anything more complex like arrays or deeper references it didn't work.

I'm not 100% sure it can handle every conceivable situation but it certainly can handle the following:
- 1 depth level references as before.
- greater than 1 level depth like in `specimen.requester`
```
    requester: {
        type: FHIR_SEARCH_PARAMETERS.REFERENCE,
        path: [['collection', 'collector']],
        referenceTypes: ['Practitioner'],
    },
```
- The item for `patient.link` which didn't work before:
```
  link: {
    type: FHIR_SEARCH_PARAMETERS.REFERENCE,
    path: [['link', '[]', 'other']],
    referenceTypes: ['Patient'],
  },
```
- And arrays of the like we have in `ServiceRequest.specimen` now with the addition of the wildcard element:
```
  specimen: {
    type: FHIR_SEARCH_PARAMETERS.REFERENCE,
    path: [['specimen', '[]', '*']],
    referenceTypes: ['Specimen'],
  },
```

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
Heya okay so I think there is one small caveat with the recursion and the path of the ServiceRequest `specimen` search parameter

NavarroEmilioLuis (approved):
I think it's looking pretty fly. Thanks for the changes. I have some comments for the tests that might help in the future, or they are just saving lines of code or style things. Honestly just nitpicking so approving - you can grab them if you feel they are worth it

PR #5367, opened 2024-02-21
fix(programRegistries): NOTAM-290: Derive clinician ID from submitting user ID

NavarroEmilioLuis:
### Changes

With this change, we will always ensure that a clinician ID gets saved


edmofro (changes requested):
Suggest we also make the db column non-nullable to match this behaviour and mobile

PR #5370, opened 2024-02-21
fix(programRegistries): TAN-2482: Add missing foreign keys and constraints

NavarroEmilioLuis:
### Changes

Add non null constraint in patient program registration and all missing foreign keys for all program registry tables


edmofro (approved):
lgtm. wow, there are a lot of fks.

PR #5371, opened 2024-02-21
fix(programRegistries): NOTAM-300: Change handling of most recent registration

NavarroEmilioLuis:
### Changes

Because of software limitations we are changing the approach on how the most recent registrations are determined


edmofro (changes requested):

edmofro (approved):
lgtm!

PR #5382, opened 2024-02-22
fix(programRegistries): TAN-2429: Improve conditions hitbox and search input placeholder

NavarroEmilioLuis:
### Changes

Real issue is just the first one, Text's onPress was blocking parent handler. The other change is something that I think should've been done from the start and avoid hardcoding that


edmofro (approved):
lgtm

PR #5383, opened 2024-02-22
db(migration): Use default value 0 instead of false

NavarroEmilioLuis:
### Changes

This is needed for backward compatibility

https://www.sqlite.org/datatype3.html
2.1. Boolean Datatype
SQLite does not have a separate Boolean storage class. Instead, Boolean values are stored as integers 0 (false) and 1 (true).
SQLite recognizes the keywords "TRUE" and "FALSE", as of version 3.23.0 (2018-04-02) but those keywords are really just alternative spellings for the integer literals 1 and 0 respectively.


dannash100 (approved):
yessir!!

PR #5389, opened 2024-02-23
fix(tests): Copy the dataValues and not the model props

NavarroEmilioLuis:
### Changes

Sneaky!!


edmofro (approved):
Squash and merge when ready!

PR #5396, opened 2024-02-25
fix(programRegistries): TAN-2424: Bug displaying program forms

edmofro:
### Changes

New rendering rules for filled-out forms assumed the presence of `writeToPatient`, where it may not exist, causing a crash.

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
Your change looks sensible but the whole thing looks a bit off. Do we know what would normally be the `question.dataElement.type` when this happens?

~I'm thinking, did we intend to break here? Because there is some validation around this and I think `fieldType` should actually always be present if `writeToPatient` exists in the config 🤔~ Ah no, nevermind. I can see that `writeToPatient` is not required, spoke too soon.

NavarroEmilioLuis (approved):
Yeah okay I think I've already wasted too much time trying to check this out so happy to move on

NavarroEmilioLuis (approved):
Nice!!

PR #5403, opened 2024-02-27
fix(programRegistries): TAN-2423: Display name not id of ref data

edmofro:
### Changes

Changes `getFullPprById` to `getRegistrationForDisplay`, and instead of returning nested objects representing clinician, village, etc., it just adds fields like `clinicianName`, `villageName`, etc.

This makes it work seamlessly when trying to show the name of the relevant reference data

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (commented):
Right so this change looks clever, but have a couple of thoughts:
- Might be too program registry scoped, what about other reference data?
- The constant falling out of sync with the server version will start to become Frankenstein (I think ultimately okay with comments)

PR #5409, opened 2024-02-27
feat(central): KAM-173: Split FHIR worker from task runner

passcod:
### Changes

- Split the FHIR worker from the scheduled task runner
- Remove `--provisioning` option; the command should be used instead (the option was a hack for ECS deploys)
- Normalise names of subcommands in central (with aliases for back-compat)
- Add `--topics` option to fhir worker to limit which topics are handled

(Please review, but can't merge until K8S environments land.)

### Checklist

- [x] Code is finished


NavarroEmilioLuis (approved):
LGTM just a minor copy thing and a question around provisioning

PR #5415, opened 2024-02-28
feat(translations): NASS-1140: pre-populate Enums in Translations Table

namhoangbes:
### Changes

Add a db migration file to populate the translation data to the `translated_string` table in the database

### Checklist

- [x] Code is finished

_Upon merging:_


tcodling (changes requested):
## TLDR: This is great but we are going to park this for now and have a tech design to talk through a bit more

Hey Nam! Was just having a look through this and you did a great job considering that there was basically no detail on the card. This is pretty close to what we want but i think we should use a slightly different approach. Really sorry about that i didn't actually realise how little detail was on the card so i am pretty impressed by this!

When we said populate the enums we actually meant just the "constants" displayed on the frontend which are essentially the constants implemented in [TranslatedSelectField](https://github.com/beyondessential/tamanu/pull/5325/files) and [TranslatedEnumComponent](https://github.com/beyondessential/tamanu/pull/5315/files) or the ones with "prefix" prop applied.

I can see you have also included all of the stringIds that are supplied to the TranslatedText components? This is also a bit out of scope of this card as we will populate these through the master spreadsheet and we don't necessarily want to lock ourselves into these ids by using a migration. In this card we really only want to populate the stringIds for the constants defined in the `constants` package and the constants folder in the `web` package

Just out of curiousity, where did you get the Khmer translations from. We expected to just seed the ids for english as not every deployment needs/wants the khmer translations in the DB. We probably shouldn't include seeding these in the work but maybe these translations are still valuable?

PR #5416, opened 2024-02-28
feat(reminder): EPI-703: Reminder mobile UI components

tareq89:
### Changes

This one actually fits three cards into one. Not ideal but that's what it was!


NavarroEmilioLuis (changes requested):
Some notes for now

PR #5421, opened 2024-02-28
fix(mobile): NOTAM-300: show missing items on dark mode

NavarroEmilioLuis:
### Changes

We do not support dark mode yet, so this is just explicitly setting that so there are no surprises with the display


dannash100 (approved):
Just did a lil idiot check on google re the wording and looks perfect for needs 💯

PR #5427, opened 2024-02-29
fix(programRegistries): TAN-2465: Solve timezone issues

NavarroEmilioLuis:
### Changes

Sending the client submit date should help us keep a coherent timekeep


edmofro (approved):
lgtm!

PR #5428, opened 2024-02-29
fix(fhir) SAV-541: Fix cherry pick conflict (2.1)

chris-bes:
When cherry picking SAV-541 from 1.36 to 2.1, there were a lot of restructured files so I made a few minor mistakes picking the wrong changes... This PR is to fix that conflicts. 

Don't think it actually affects anything as tests were still passing and the mistakes were only in test files, but still want to correct them to be on the safe side.


kujo4pm (approved):
LGTM!

PR #5432, opened 2024-02-29
fix(mobile): TAN-2424: Properly display survey response answers

NavarroEmilioLuis:
### Changes

Logic was a bit fuzzy and after solving minor bugs here and there everything went too blurry. This fix introduces the same logic used in the server version.

Please also review this one: https://github.com/beyondessential/tamanu/pull/5433 as I think we would benefit from those changes


edmofro (changes requested):
@NavarroEmilioLuis this looks sensible, but loses the bug fix where a reference-data sourced field that is read only will not display the name (will display the id instead), because it isn't set to `Autocomplete` as the `writeToPatient.fieldType`.

Perhaps let's pair on it?

PR #5438, opened 2024-03-01
fix(mobile): NOTAM-308: Add background color back in

NavarroEmilioLuis:
### Changes

Changes were made [in this PR](https://github.com/beyondessential/tamanu/pull/4557/files#diff-00df5cddea4f83024cf40249100d85bf4e9614dc6a5d0f1ff14202ced70b3596L45) and they shouldn't have


chris-bes (approved):
Oopsy, thanks for picking that up 🙏

PR #5439, opened 2024-03-01
ci: KAM-176: Build ARM64 containers

passcod:
### Changes

- Build container images for ARM64
	- (Can we immediately use ARM64 in ECS deploys? Unfortunately no.)
	- And add a step that adds a manifest that "combines" two images for the same thing for different archs, which makes it such that a node using an image automatically figures out which architecture it needs and pulls the right one.
- Build the meta-server container (preparing for moving it out of Elastic Beanstalk)
- Build container images all the time, so more workflows can be powered by them
	- They take about a minute each so it's not very expensive
	- Also makes it easier to deploy to UATs
	- Plus, I add an "All images built" job that we can gate PRs on, to ensure we don't accidentally break the containers (and therefore the production builds)
- Change the way server and frontend packaging works so we only build the containers once
	- Frontend packaging now [takes 20s](https://github.com/beyondessential/tamanu/actions/runs/8106484006/job/22156523749?pr=5439) instead of [4+ minutes](https://github.com/beyondessential/tamanu/actions/runs/8105840618?pr=5444)
	- Server packaging now [takes 6 minutes](https://github.com/beyondessential/tamanu/actions/runs/8106729885/job/22157185915?pr=5439) instead of [10+ minutes](https://github.com/beyondessential/tamanu/actions/runs/8092065751)
- Use https://github.com/docker/metadata-action to build image labels instead of winging it
- Remove signing step as we don't really make use of it at the moment, and should revisit that approach
- Only build server/frontend packages for production releases. Previously web packages were built for all pushes and PRs, but we don't make use of that.
- Finally there's some optimisations around how the current version and the branch name are obtained, which reduce both time, cost, and complexity.
  - Specifically instead of calling a separate shared branch info workflow, which requires allocation of a runner plus the overhead to set up its run, and had the version as optional which required some hacks to get it not to fail when source is unavailable (such as after a branch is deleted), the code is duplicated (but call sites are reduced to only two, so it's not so bad) but runs as a step instead of a job/workflow, which means it takes milliseconds.
  - Further, the version fetch from package.json is done once when building the container image, then saved in the image metadata, so it can be retrieved given just the image when needed.
  - Finally the image is always tagged as `sha-${{ github.sha }}` which is always the same for all workflows for a commit, so retrieving it is trivial.

### Checklist

- [x] Code is finished

_Upon merging:_

  - with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)


edmofro (approved):
I've done partial diffs in vscode on the moved blocks of code, and reviewed other stuff in page, and all looks sensible...but I gotta say this is a fairly hand wavey approval. lgtm!

PR #5445, opened 2024-03-01
fix(programRegistries): TAN-2424: Use logic from web for PatientData fields

edmofro:
@NavarroEmilioLuis this copies the logic from desktop, and works almost perfectly

The display when you're filling out the survey is not so nice, as the disabled autocomplete on mobile is barely legible. We could create a different component instead - but gotta go catch a ferry!

image


NavarroEmilioLuis (commented):
I think I can get around this one - we just need to touch up or create a new component that is more visible and that's that. Just a minor note but yeah this works for me

NavarroEmilioLuis (approved):
Amazing 👍 I'll go cancel mine then

PR #5449, opened 2024-03-04
fix(programRegistries): NOTAM-301: Filter conditions by registry

NavarroEmilioLuis:
### Changes

Yeah this was missing, the provider was only ever set under the program registries view. Ended up moving the context up, as it seems fairly inoquous based on other providers and we need that in three different places.


edmofro (changes requested):
I'm not gonna stay firm on this but have one suggestion to move a filter from the front to the back end (I think back end filtering should generally be the default, right?)

edmofro (approved):
thanks!

PR #5450, opened 2024-03-04
feat(central): SAV-548: ServiceRequest to handle panels and independent tests

kujo4pm:
### Changes

This card is to update the way that we handle ServiceRequest for Lab Requests. It's meant to manage the business rules for:

> - A ServiceRequest from a Lab Request now can handle Lab Tests directly with the Lab Request. 
> - If there is no panel then ServiceRequest.code does not exist. This is ok.
> - It is not possible to have testTypes associated with both the lab request directly and through panels. This will throw an error.
> - It also includes (ServiceRequest.orderDetail[n].coding.code[n]) codes both for the externalCode but we should also be exposing the code too - only if the latter exists.
> - We'll also need to make the systems (ServiceRequest.orderDetail[n].coding.code[n].system) configurable for both. 

It's quasi-TDD beginning with tests then writing the code - so I'm pretty sure it works ok. 
### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
Overall looks good, I have a couple of observations that might improve readability. Nothing is super important but also want to clarify a few bits before moving on

NavarroEmilioLuis (approved):
Hey man, changes looks good, thanks. I think you left something in the test file? But happy to approve anyway

PR #5453, opened 2024-03-04
fix(programRegistries): TAN-2498: Properly go back after submitting form in mobile

NavarroEmilioLuis:
### Changes

Okay so this problem only happened for program registry related forms - the reason being that the code was getting blocked when trying to create a patient program registration.

This happened because:
- Wrong TypeORM synxtax
- Clinician ID is non nullable


edmofro (approved):
lgtm!

PR #5455, opened 2024-03-05
fix(programRegistries): TAN-2462: Unblock program registration view

NavarroEmilioLuis:
### Changes

This error is caused because of using a future date on the registration. This is more of a workaround, the proper fix will require a bit more thought and it's already scheduled for phase 2.


edmofro (approved):
An ok workaround, but let's aim to remove that optional chaining in phase 2

PR #5457, opened 2024-03-05
style(mobile): Fix linting stuff

NavarroEmilioLuis:
### Changes

3 type of changes:
- Lots of enums were having this error, when they are clearly being used (outside their file). See https://github.com/typescript-eslint/typescript-eslint/issues/2619
- A couple of actual errors on the migration files, which should be okay because it doesn't affect the file itself.
- Getting rid of the import errors `Cannot find module '<module>'.` which happens quite often, this does not seem to affect the way we package stuff so we're good (the app builds fine in development)

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -77,7 +77,14 @@ export default [
     files: [`packages/mobile/**/*.${TS_EXTS}`],
     rules: {
       'no-unused-vars': 'off',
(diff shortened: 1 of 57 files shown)


passcod (approved):

edmofro (approved):
Codeowner stamp

passcod (approved):

PR #5458, opened 2024-03-06
fix(programRegistries): TAN-2497: Remainder stuff

NavarroEmilioLuis:
### Changes

Two main changes here:
- Change mobile label display for patient program registration "registration status" field (this is what always was supposed to be according to Figma, web already displays it that way).
- Filter deleted program registration conditions from display


edmofro (approved):
No blocking changes, but a couple of minor suggestions

PR #5460, opened 2024-03-06
ci: KAM-100: deploy ECS with ARM64

passcod:
This gives us **20% savings** on the ECS cost, which is pretty good for a one line change!


edmofro (approved):
🎉

PR #5495, opened 2024-03-13
ci: KAM-174: deploy to k8s

passcod:
### Changes

- Rewrite the deploy process for K8S
- Abstract away the setup steps into their own repo-local actions
- Implement checkbox-controlled deploys [[see draft docs](https://beyond-essential.slab.com/posts/tamanu-auto-deploys-kubernetes-kkovoquc#h4byo-config-templates)]
	- Possible to have multiple deploys per PR, all auto-updated
	- No longer reliant on the flaky github PR label UI
	- Adds many options for customising the deploys, such as adding or removing facilities, scaling, using a different template for provisioning or config, IP restrictions, etc
	- A `%pause` directive lets you disable auto-updates but leave deploys up, e.g. if someone is testing the current state and devs are simultaneously working on the PR, that can be more stable than restarting on every new push
	- Possible to control non-PR auto-deploy branches with issues
- Write a parser for the markdown schema and some tests for that parser
	- I used `tape` for the tests instead of jest because jest fought me and wouldn't work because of the complicated rewrites and reparsing it does... `tape` _just works_
- Upgrade actions to actions/github-script@v7 to avoid the node16 warnings
- Undeploying a deploy that doesn't exist stops gracefully (green tick) instead of failing as if it errored
- If the images are already built we don't build them again
- Images are hosted in Github instead of AWS, so we can easily make them public when we go fully open source
- Images get created for release tags, becoming part of our release outputs
- Authentication to AWS for deploy is no longer required, which opens up some possibilities (and makes it easier to control via pulumi alone)

### Checklist

- [x] Code is finished

### Deploy

- [x] **Deploy (Klaus)**  %facilities=0
- [ ] **Deploy (Sima)**  %facilities=3


rail-rain (changes requested):
I commented on a few stylistic things, which already stretched me. (I swear I read all the changes.) Otherwise it looks good to me.

rail-rain (changes requested):

rail-rain (approved):
I'll say this is ready. I hope I'm not missing something massive.

PR #5501, opened 2024-03-13
ci(determinism): separate the "Tests pass" for non determinism

passcod:
### Changes

- Separate the "Tests pass" for non determinism, so that we can enable/disable that separately from the general CI. Emilio caught it earlier actually but I thought github worked differently.
- Make it possible to generate a db dump for any commit (e.g. in PRs) by including `[generate-fake-db]` anywhere in the commit message (subject or body).

### Checklist

- [x] Code is finished


edmofro (approved):
Non-blocking comments, but lgtm!

PR #5504, opened 2024-03-14
fix(web): reformat-program-registry-actions

tcodling:
Result of merge between program registries/translations. 

In doing translations work we changed the format of the actions supplied to MenuButton.jsx to an array rather than an object with the key being its label. It didnt work with our new translated text component and was also quite a different way of handling options to the rest of the app so we adjusted to match that

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


tcaiger (approved):
I never liked those object definitions for the labels and actions so this is an all round win!

PR #5516, opened 2024-03-14
revert: default facility pm2 config to single process

passcod:
### Changes

Turns out the facility server is stateful lmao

### Checklist

- [x] Code is finished


edmofro (approved):
😢

PR #5544, opened 2024-03-19
fix(web): error on clearing survey field

tcodling:
### Changes

The new ProgramRegistry context is sending requests for the programRegistries for a program even if the field has been cleared to empty. this throws an error for hitting the request with an "undefined" value. Ideally I would reorganise this logic to be a useQuery that listens to a programId set and wont fire if falsy, but it seems like it would require a decent bit of logic shuffling around (more than i am comfortable doing as i am unfamiliar with program registries and we are right at the end of regression).

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


dannash100 (approved):
Good solution for now!

PR #5546, opened 2024-03-19
fix(saveFile): TAMOC-169: fallback when showSaveFilePicker isn't available

passcod:
### Changes

- Add fallback when showSaveFilePicker isn't available (less user-friendly, but always available)
- Refactor `saveFile` util: we only ever use a single extension, so accept a string instead of an array

### Checklist

- [x] Code is finished

_Upon merging:_


NavarroEmilioLuis (approved):
LGTM

PR #5563, opened 2024-03-20
feat(programRegistries): SAV-573: Fine grained permissions for web client

NavarroEmilioLuis:
### Changes

A bit sprinty but think this should be it


chris-bes (approved):
LGTM!

PR #5567, opened 2024-03-21
feat(translation): NASS-1177: Translate yup errors

dannash100:
### Changes
Done! minus one thing pointed out

Includes some cleanups of duplicated string ids and adding of some that where missing where yup validation was required

### New features
- Translated text component can be supplied to a new yup method `.translatedLabel` without seralising to `[Object object]` this means we can supply the label we want to be used in required message within the yup schema i.e
 
  ```
  status: yup.string()
      .required()
      .translatedLabel(<TranslatedText stringId="general.status.label" fallback="Status"/>)
  ```
  - Will result in `Status is required`
  - This means we can use the same string ids as the labels and also easily reference localised fields without any extra logic
- Translated text inside replacements
  - This means we can translate stuff like `Search new clinician` using the clinician localised field
 ```
 const changeClinicianLabel = (
  <TranslatedText
    stringId="patient.changeClinician.examinerId.label"
    fallback="Search new :clinician"
    replacements={{
      clinician: (
        <TranslatedText
          stringId="general.localisedField.clinician.label.short"
          fallback="Clinician"
          lowercase
        />
      ),
    }}
  />
);
```
  - We could also pass this directly into `translatedLabel` for the error message

### New string id info:
- **New important string id** `validation.required` defaulting to `:path is required`
	- This is used in all required messages when a specific one isn't supplied
- **New stringId style for validation path** e.g `invoice.validation.percentageChange.path` 
  - these represent any labels referred to in required message that are not identical to the label name or the field has no label. In some cases the path differs from the label in that it is smaller and doesn't look messy in error message
  - this stringId will supply the templated text into the :path in required messages - in all other cases it just duplicates the`.label` or `.localisedField.label` of the field
- **New string id style for validation rule text** eg `validation.rule.min5Characters` 
  - Represents validation messages that do not follow the default required message
  - These are supplied where required using `getTranslation` inside yup schema

### Checklist

- [x] Code is finished
- [n/a] Tests are written

_Upon merging:_


tcodling (changes requested):
few comments and changes but nothing major.

Main thing for me is it would be so awesome if there was some way to not have to declare a label on everyone and use the id instead. But from playing around i see why you went with this. there seems to be downsides no matter what we do 😞

tcaiger (approved):
I've reviewed the last set of changes and looks good to me. I'm happy if you want to move this ticket to dev done or ready for test. Will still need a review from Thomas to satisfy github.

tcodling (approved):

PR #5572, opened 2024-03-21
fix(web): patient data form regressions

tcodling:
This pr covers 2 fixes related to the patient data form.

- The configurable displayId logic doesn't fit with the `requiredPatientData` functionality after a refractor. Basically fields now aren't showing properly when they dont have a 'requiredPatientData' property assigned to them. If we don't have this set in defaults, this means that the field should always show and this change just makes sure that we do that in the logic.
- We are only getting a generic "required" message when trying to submit a patient details form with empty required fields. We need to show the proper validation message with data type in the text. I have used `getTranslation` to get the label in place of the previous `getLocalisation` usage and have manually set the fallback for each validation message. **This is not ideal solution and would probably put a lot more effort into cleaning up this logic but given we have an upcoming card where we handle validation messages through yup about to be merged, i think this is probably fine for making sure we don't alter the existing functionality in this release. Very open to discussions around improving however**

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_


tcaiger (approved):
LGTM

PR #5583, opened 2024-03-25
feat(provisioner): KAM-175: load data from URL

passcod:
### Changes

This is both a requirement for K8S for data size reasons and also a nice quality-of-life improvement for e.g. onboarding, reducing provisioning in dev to this single call:

```console
$ yarn workspace @tamanu/central-server start provision $PWD/docs/provisioning/example.json5
```

### Checklist

- [x] Code is finished

_Upon merging:_

- [x] Update the [relevant runbook](https://beyond-essential.slab.com/posts/tamanu-monorepo-setup-63tub6pv#hte5m-populating-servers)

### DIFF
--- a/docs/provisioning/example.json5
+++ b/docs/provisioning/example.json5
@@ -14,22 +14,23 @@
       user: 'facility-a@tamanu.io',
       password: 'facility-a',
     },
-
-    'facility-b': {
-      name: 'Facility B',
-      code: 'b',
-      user: 'facility-b@tamanu.io',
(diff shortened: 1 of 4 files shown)


NavarroEmilioLuis (approved):
LGTM just a cautionary question

The provision file made me double think a couple of bits, but ultimately, for that purpose, I think it's better to leave that as is than to add abstractions that will save us a couple of lines. Maybe if a third option had to use the same file or URL thing but I think let's move on

PR #5592, opened 2024-03-26
feat(central): EPI-660 set up telegram bot api

khoi-bes:
### Changes

- Add TelegramServiceBot and config a public route for telegram web

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_


NavarroEmilioLuis (changes requested):
I get that this is meant to be a placeholder and isn't ready, but I think there are a couple of things we should resolve first, let me know your thoughts

NavarroEmilioLuis (approved):
Hey I'm happy to move on with this one based on the fact that the route is meant for telegram and also that it's an experimental/new feature.

My last suggestion would be that it does not get initialized or set without an `apiKey` similar to how we would treat other services. Presumably no API key means the deployment is not using said feature and there is no need to waste any resources there. However this is going into an epic, and I know we will eventually create that feature flag so happy to add it later, or you know, it's just a suggestion I think is better in the long run.

Aside from that, I'm leaving you one small nitpick you can ignore but I strongly ask you not to - also left some notes because I think the merge conflicts are actually wrong? It seems to me that we're removing latest changes so take a look and solve that before merging.

PR #5596, opened 2024-03-27
fix: Update patient contact model

khoi-bes:
### Changes

- Add sanitizePulledRecordData function for PatientConact model on mobile to sanitize jsonb data
- Update column "name" cannot be null

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/mobile/App/migrations/1705975873000-addPatientContactTable.ts
+++ b/packages/mobile/App/migrations/1705975873000-addPatientContactTable.ts
@@ -12,17 +12,15 @@ const BaseColumns = [
     name: 'createdAt',
     type: 'datetime',
     default: "datetime('now')",
-    isNullable: true,
   }),
   new TableColumn({
     name: 'updatedAt',
     type: 'datetime',
     default: "datetime('now')",
-    isNullable: true,
   }),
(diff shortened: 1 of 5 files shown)


NavarroEmilioLuis (commented):
Overall looks good to me, I want to make a couple of comments/questions before moving forward:

- Why do we want to make `name` mandatory?
- ~This is totally nitpicking, but I would much rather avoid adding default stuff (like `isNullable: true` on the metadata fields.~ turns out that's actually not the default in TypeORM!
- If this is already deployed to any type of environment, the changes won't be applied. Migrations should not be modified unless they haven't made it into any environment.
- That said, I know this is on an epic branch, and this is the way to go, but know that you will have to do some manual stuff on the test branch

I do want to request this as a change though:
- On server, these are definitely non nullable, or at least is something that we started doing a while back ago, if I recall correctly it's because without it, the default won't actually do anything, I might be wrong on that.

NavarroEmilioLuis (approved):
LGTM! Interesting that they both behave different (TypeORM and Sequelize) with the defaults, but that's ok, this seems exactly like what we want

PR #5598, opened 2024-03-27
fix(web): Template query appending question mark (Hotfix 2.3)

dannash100:
### Changes
Api client changes means that if null not specified explicitly in useTemplate get request `template/${key}?facilityId=${facility.id}` will hit endpoint with a facility id of `ref/facility/a?` with an appended question mark and will not receive the template.

This will need to be hot-fixed back to api changes 2.2 and forwards to 2.4

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### DIFF
--- a/packages/web/app/utils/useTemplate.js
+++ b/packages/web/app/utils/useTemplate.js
@@ -8,6 +8,6 @@ export const useTemplate = key => {
   const { facility } = useAuth();
 
   return useQuery(['template', key, facility.id], () =>
-    api.get(`template/${key}?facilityId=${facility.id}`),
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (commented):
Hmmmmm are you completely sure about this? I can see other usages of `api.get` that do not need to pass `null` as second argument (and they also use template strings with endpoint).

There's also something weird about having a fixed query param of `facilityId`, why don't we pass it as regular params?

NavarroEmilioLuis (approved):
Left a comment/suggestion but happy to move on with this approach

PR #5604, opened 2024-03-27
fix(programRegistries): SAV-573: Correct permission checks

NavarroEmilioLuis:
### Changes

So I guess we missed the fact that deletion should check for `write`. Also I failed to remove misguided permission checks (these didn't affect the behavior but it's better to clean it up and avoid future confusion)


tcodling (approved):
ah yep i see how that would be easy to do. LGTM!

PR #5636, opened 2024-04-02
test: Resolve intermittent failures

NavarroEmilioLuis:
### Changes

This type of error: https://github.com/beyondessential/tamanu/actions/runs/8516684545/job/23326101467?pr=5632 was common, after some investigation, it seems we weren't properly awaiting for the promises. See https://jestjs.io/docs/asynchronous#asyncawait (specially the caution bit).


passcod (approved):
oh neat!

edmofro (approved):
stamp!

PR #5672, opened 2024-04-09
fix(dataDeletion): SAV-652: Ditch DeletionStatus

NavarroEmilioLuis:
### Changes

First, did some changes that I think tidy up things more cleanly or they were pending from previous PRs. The core work here is to get rid of deletionStatus, as it wasn't really used much and it was just confusing (more info in the card)


edmofro (changes requested):
Nice refactors, fixups, and renames along the way!

edmofro (approved):

PR #5675, opened 2024-04-11
fix(dataDeletion): SAV-558: Tidy up frontend

NavarroEmilioLuis:
### Changes

Well, this wasn't working and this is a tidy up found trying to test the feature :)

PS: mostly frontend, but includes other random stuff


chris-bes (approved):
LGTM!

PR #5681, opened 2024-04-12
feat(dataDeletion): SAV-558: Handle bad scenarios

NavarroEmilioLuis:
### Changes

- Avoid restoring on facility server
- Deleted record should never be restored by central reconciliation
- "children" records should all be deleted when parent is deleted
  - This one is still pending to implement, but I just want to check whether the approach is good first
  - Instead of redoing a different handler for each model, I thought about doing all: https://github.com/beyondessential/tamanu/pull/5695


chris-bes (changes requested):
LGTM! Nothing major that I can find, just leaving a couple of minor suggestions 👍

chris-bes (approved):

PR #5684, opened 2024-04-12
fix(mobile): NOTAM-157: remove meta-dev usage

passcod:
### Changes

meta-dev is never used and shouldn't be used, especially since this ticket is over 2 years old...

### Checklist

- [x] Code is finished


tcodling (approved):
oh my god amazing hahahaha. i have commented this out so many times in dev 😆

PR #5691, opened 2024-04-15
ci: make deploy cleanup stop complaining

passcod:
### Changes

The auto-deploy cleanup routine _works_, but it [reports errors](https://github.com/beyondessential/tamanu/actions/runs/8673340406/attempts/2). It has some redundancy built-in so that doesn't stop it working, but it's annoying.

This PR stops pulumi from complaining about shorthand stack names so it won't error anymore.

### Checklist

- [x] Code is finished


dannash100 (approved):
Looks gud!

PR #5693, opened 2024-04-15
feat(server): EPI-678: reminder messages support deregistration workflow

khoi-bes:
### Changes

- Add new language configuration
- Add translation function for server
- Use websocket to trigger deregistration flow

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### Deploys


NavarroEmilioLuis (approved):
I think everything looks okay, I don't love the way it's structured but the files are already there and you're preserving the same design so I don't think we should bother much. My biggest dislike is relying on variables on an outer scope that need something else to happen first to be available. I realize this might be inevitable but it then gets weird as to what we should action and what we shouldn't action - I would probably check we have all requirements or otherwise do nothing (early return), but then again, I'm not familiar with the tasks that need to be done.

I'm approving and leaving 2 optional suggestions.

PR #5695, opened 2024-04-15
feat(dataDeletion): SAV-558: Use generic hook

NavarroEmilioLuis:
### Changes

Given that we are aiming to mark all children (and their children) records to be deleted with the parent, I thought a generic approach was more sensible, rather than explicitly declaring all models that use the functionality.


chris-bes (approved):
LGTM! This is great changes! Nothing beats a generic approach.

Just wanna flag that it has some conflicts with the base branch.

PR #5702, opened 2024-04-16
feat(central): SAV-653: specimen attached after lab request created

kujo4pm:
### Changes

SpecimenAttached is not properly set for [sav-653](https://linear.app/bes/issue/SAV-653/bug-specimenattached-in-labrequest-not-updated-when-specimen-attached).

This is a simple fix with a test in this PR but I think the field is redundant - it's really just saying
```
specimenAttached = specimenTypeId == null;
```
I would say it would not be difficult to eliminate the field entirely but that might require a fair bit of regression testing.
image

### Checklist

- [x] Code is finished
- [x] Tests are written

_Upon merging:_

### Deploys


NavarroEmilioLuis (commented):
Hey so I'm not sure why the test pass but I'm requesting changes. I also want to note that I checked the code for `specimenAttached` and I'm baffled it works because we pass 'yes' or 'no' instead of `true` or `false`... But it does work so oh well.

NavarroEmilioLuis (changes requested):
See above

NavarroEmilioLuis (approved):
Changes look good!

PR #5706, opened 2024-04-17
tweak(web): use .env to specify vite proxy target

passcod:
### Changes

Lets you create a .env file containing e.g.

```env
TAMANU_VITE_TARGET=http://localhost:4000
```

to specify the backend target instead of changing the vite config itself.

### Checklist

- [x] Code is finished


dannash100 (approved):
Great solution - I was thinking about this other day.
mabye we should add to env.example or somthing - so we have that default of localhost:4000 out the box

PR #5713, opened 2024-04-18
fix(mobile): SAV-574: Hide program registries from search if no permission 

tcodling:
### Changes

Quite unsure about the approach for this one. Ideally we would filter by permission in the query but the suggester logic really doesn't seem setup for that. I have gotten around it by adding in a front end "filter" that you can pass on suggester creation that checks permission however it feels a little bit too hacky 🤔 its the last release issue so I am trying to be conservative but im really not sure. Very open to suggestions if i am missing something. From what i can tell it hasn't really been done on mobile yet (with a suggester) we have a few select fields that filter by permission

### Checklist

- [ ] Code is finished
- [ ] Tests are written

_Upon merging:_

### Deploys

### DIFF
--- a/packages/mobile/App/ui/helpers/suggester.ts
+++ b/packages/mobile/App/ui/helpers/suggester.ts
@@ -22,11 +22,16 @@ export class Suggester<ModelType extends BaseModelSubclass> {
 
   formatter: (entity: BaseModel) => OptionType;
 
-  constructor(model: ModelType, options, formatter = defaultFormatter) {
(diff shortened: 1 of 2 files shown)


tcaiger (changes requested):
I like the idea of keeping it pretty simple and conservative for this PR. I have suggested a teak in the suggester code which I think would be a bit nicer and is compatible with the auth callback in you ProgramRegistrySuggester change

tcaiger (approved):

PR #5711, opened 2024-04-18
feat(downloads): SAV-639: implement two-step downloads for reports

passcod:
### Changes

- Split the report download action into two steps: generate the report, and save it.
- Disable the generate button until a report is selected
- Add a spinner progress indicator
- Clear success/error/download state when the form is changed

### Checklist

- [x] Code is finished
- [x] Testing notes added to the Linear issue
- [x] Cherry-pick to 2.5

_Upon merging:_

- [ ] Update the changelog (find it [here](https://github.com/beyondessential/tamanu/releases) in the appropriate _draft_ release)
  - with a ticket reference (e.g. `TAN-123: a one line description`, keep the lists sorted)

### Deploys

- [x] **Deploy to Tamanu Internal**

PR #5717, opened 2024-04-19
feat(downloads): SAV-639: implement two-step downloads for reports

passcod:
### Changes

Cherry-picking #5711 to 2.5

### Checklist

- [x] Code is finished


edmofro (approved):
stamp

PR #5747, opened 2024-04-24
feat(auth): NASS-1248: restrict users to facilities via userfacility table

passcod:
### Changes

Basically Cambodia's MVP, implemented while I was listening to that incredibly tedious meeting.

- Makes the `user_facilities` table importable
- If users don't have an association to a facility in this table they get rejected on login
- Supports M:N so one user can "belong" to multiple facilities
- Refactor: clean up admin routes
- Adds a Users admin panel with a list and a create modal — helps testing this
- Adds a `checkPermission` method on User to check permissions on users other than the logged in one, or during auth
- Adds the check to the sync endpoint so that you can't sync if you don't have permission for the facility
- Adds the check to the facility listing endpoint so you can only see facilities for which you have permissions.

### Todo

- [x] Feature-gate it in config to avoid config churn in every other deployment
- [x] Add and implement bypass permission `('login', 'Facility')`
- [x] Mobile
- [x] Central Admin panel

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### Deploys


NavarroEmilioLuis (commented):
I think everything looks good overall, left a couple of questions and optional suggestions.

anilvkaran (approved):

edmofro (commented):
To check my understanding: `user_facilities` was being used for precisely nothing before, right?

NavarroEmilioLuis (commented):
Hmmm okay I have a couple comments but I think everything looks good, will update one file, but the other comment I'm unsure if I'm missing something there 🤔

NavarroEmilioLuis (approved):
Alright I think everything is done for this one then

PS: adding merge pause cause IDK if this still needs something else

PR #5754, opened 2024-04-25
fix(dataDeletion): EPI-566: Address testing notes

NavarroEmilioLuis:
### Changes

Three small changes:
- Stop showing error dialog
- Update copy to use display ID short label everywhere (using translations, unsure if this is the place now? or maybe it should use localisation?)
- Add survey response answer - vital log relationship to propagate deletes


dannash100 (approved):
Yessir!

PR #5764, opened 2024-04-28
ci: Add cd option to only build images without deploying

passcod:
### Changes

This is occasionally useful e.g. when deploying UATs or to Itis. Currently this requires making a deploy go up and then remembering to cancel the up just after the images are done but before Pulumi starts, which is annoying. With this it will be a matter of adding `%imagesonly` to the deploy config line.

### Checklist

- [x] Code is finished
- [x] Updated the [config reference](https://beyond-essential.slab.com/posts/tamanu-auto-deploys-kubernetes-kkovoquc)

### Deploys

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -45,18 +45,6 @@ jobs:
   config:
     name: Workflow config
     runs-on: ubuntu-latest
-    outputs:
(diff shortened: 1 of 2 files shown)


dannash100 (approved):
Looks gud!

PR #5765, opened 2024-04-29
feat: EPI-758: Support new template creation for clinical notes in Admin panel

khoi-bes:
### Changes

- Rename model, api route
- Allow create new template with new type
- Write new migration to add new column "type" to "Template" table

### Checklist

- [x] Code is finished
- [ ] Tests are written

_Upon merging:_

### Deploys


tcodling (changes requested):
generally looks good. Made a few notes but the main thing is:
- `template` is a bit too general for the model name. I think it should probably be something like `document_template` since these templates are displayed in the documents tab
- With the recent merge of our db settings reader, we can now access settings on the frontend using useSettings() instead of needing to use those endpoints

tcodling (changes requested):
lgtm! Just one small thing which is that these permissions arent added to `roles.js` hardcoded permissions file. These should have been added when PatientLetterTemplate was added but was obviously missed 🙏

tcodling (approved):
lgtm!

PR #5776, opened 2024-04-30
feat(fhir): SAV-662: imaging study cancel workflow

kujo4pm:
### Changes

RISPACS have asked to be able to cancel Imaging requests on their end. This should enable us to do it. The code uses the existing code developed by @NavarroEmilioLuis to handle incoming results. This means that they can supply some data - like identifier which includes the accession code that will be ignored... I think this is ok.

For more info on business rules see: [SAV-662](https://linear.app/bes/issue/SAV-662/imagingstudy-cancel-workflow)

### Checklist

- [x] Code is finished
- [x] Tests are written
- [ ] Manual upgrade steps added to the Linear issue, if needed

_Upon merging:_

### Deploys


NavarroEmilioLuis (approved):
LGTM, left a couple of questions but I think we can move on unless they are something to think about

NavarroEmilioLuis (approved):
Changes LGTM!

PR #5790, opened 2024-05-01
fix(dataDeletion): EPI-567: Remove actions column when empty

NavarroEmilioLuis:
### Changes

Pretty self explanatory. Also removed some unneeded checks as the actions array will always be populated by that point


dannash100 (approved):
I love it!

PR #5822, opened 2024-05-07
feat(invoicing): EPI-837: simplify db schema for invoice discount

namhoangbes:
### Changes

Add `discount_markup_percentage` and `discount_markup_reason` to `invoices` table

### Deploys

### DIFF
--- a/packages/shared/src/migrations/1715069649383-addAddDiscountColumnsToInvoice.js
+++ b/packages/shared/src/migrations/1715069649383-addAddDiscountColumnsToInvoice.js
@@ -1,0 +1,22 @@
+/** @typedef {import('sequelize').QueryInterface} QueryInterface */
+import { DataTypes } from 'sequelize';
+
+/**
+ * @param {QueryInterface} query
+ */
+export async function up(query) {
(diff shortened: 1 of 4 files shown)


NavarroEmilioLuis (commented):
Is there a reason to keep `InvoicePriceChangeItem` and `InvoicePriceChangeType` if they are being deprecated?

If they were previously populated I'm assuming we should also migrate those... Though maybe that's on a future step, so LMK

PS: looks good, just want to confirm before moving forward!

NavarroEmilioLuis (changes requested):
After reading the card, there is a last comment that says we should include a migration task of truncating all invoice tables as they're not really used. So yeah I think we can get rid of the deprecated invoice models as well (or we can create a card for that if it's not already there, given you might need the models to be there until all their logic is removed)

NavarroEmilioLuis (approved):
I mean, I just wanted to double check a couple of things, happy to stamp. Please before merging check with Nam about the migration thing, cause it might not be needed? In any case I think it doesn't really need another review if we're ditching it or keeping it.

PR #5696, opened 2024-04-16
ci(android): nass-1149: Pack mobile in github action with ability to white label for cambodia

dannash100:
### Changes 

Context:
- App center is being shut down next year
- App center hooks on branches don't offer a great place to update static content needed to white label app title and icon etc

Using github action for packing android release
- Adds branding logic for app title and icons
- Determines in app branded content reading Config from env
- Default the central server during packaging

Signing for release
Working with secrets
```
SIGNING_KEY_PASSWORD 
SIGNING_STORE_PASSWORD
ANDROID_KEYSTORE
```
and var `ANDROID_SIGNING_KEY_ALIAS `

### Checklist

- [x] Code is finished
- [x] Tests are written

### Deploys

- [ ] **Deploy to Tamanu Internal**  %mobile=always

PR #5827, opened 2024-05-08
ci(release): add contents:write permission to update release

passcod:
Found while debugging #5696


NavarroEmilioLuis (approved):
I mean... can see the syntax being used elsewhere and it looks good so yeah happy to approve

PR #5828, opened 2024-05-08
fix(pdfCertificates): NOTAM-353: Use fallback footer

NavarroEmilioLuis:
### Changes

So this was not working properly, given the asset endpoint will always return an object, checking for `!queryData` will always evaluate to false. What we want is to ensure `queryData.data` is there instead.

Also took the liberty to change double not operator for `Boolean`

### Deploys

### Remember to...

- ...write or update tests


tcaiger (approved):
Nice catch

PR #5830, opened 2024-05-08
fix(settings): NASS-889: Settings for front end generated in appropriate scope

dannash100:
### Changes
Previously all logins received central scoped settings i.e global + central settings marked for front end usage

To fix:
- Front end settings should be scoped to facility and not use the central servers initialised reader when loging in.
- This also means that if local fallback login is used it will still return settings.
- It is however valid to return settings from central in the case of admin panel central server login

### Deploys

### Remember to...

- ...write or update tests


tcaiger (approved):
Makes good sense. Ka pai

PR #5847, opened 2024-05-09
fix(programs): NASS-1253: fix calculated questions

tcaiger:
### Changes

Convert calculation inputs to an empty string if they are undefined so that the calculations don't fail if there is an empty input. This change results in feature parity between the survey calculations in the front end ui and the survey logic on the server.

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
LGTM... I stopped a bit when looking at defaulting to empty string because of types but the test case seems to properly handle it as 0 so I think we're good

PR #5851, opened 2024-05-12
fix(reports): NASS-1224: Report fields not using defined labels

flynncs:
### Changes

Fixes issue where report fields/parameters were using hardcoded `TranslatedText` labels, rather than those defined in report generation.
This does mean, however, that if a label is defined for a report field, it will no longer be translated.

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
lgmt!

PR #5858, opened 2024-05-13
fix(central): SAV-587: Adding logging to LabRequest Update

kujo4pm:
### Changes

Had skipped logging before. Now it is included

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/app/hl7fhir/materialised/create.js
+++ b/packages/central-server/app/hl7fhir/materialised/create.js
@@ -30,8 +30,9 @@ export function createHandler(FhirResource) {
     );
 
     const resource = new FhirResource(validated);
-    const upstream = await resource.pushUpstream();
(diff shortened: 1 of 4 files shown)


NavarroEmilioLuis (changes requested):
Overall looks good, but I think it would be better to handle it explicitly without the added method or, if we're adding the method, to the chore and unify/reuse stuff.

I left more detailed info below, I would say overall, it's better to just cover this scenario and that's it, in case refactor/reuse is not exactly the same (though it should be, at least in the endpoint).

In any case, I would be happy to leave the method if we move out the transaction 🤔 ultimately your call, LMK what you think

NavarroEmilioLuis (approved):
Thanks! I think this is better in the long run!

PR #5859, opened 2024-05-13
fix(vaccineStatus): NASS-1146: Upcoming vaccination issue with multiple doses not showing latest dose

dannash100:
### Changes
Into epic (this is why updating the migration directly)

Fixes issue related to vaccine status view where it would get stuck with vaccines after first administered dose - even if the second dose was administered it would still show the 2nd dose as upcoming.

The fix ensures that the latest administered vaccine is the one used when comparing `sv2.index > sv.index`

Testing on clone environments this change does not significantly blow up query time and seems to be around the same query time once the range of timings are normalized.

### Deploys

### Remember to...

- ...write or update tests


tcaiger (approved):
Nice!

PR #5869, opened 2024-05-15
fix(sync): SAV-600: Fix scaling issue with sync filtering

chris-bes:
### Changes
- Fix scaling issue with sync filtering

### Deploys

### Remember to...

- ...write or update tests


edmofro (changes requested):
Thank you @chris-bes. I have a few suggestions inline, but also would wonder what the likely performance impact is of adding the patient select query for every model rather than once. Probably we can rely on PostgreSQL caching to do a good job of that, but it would be worth running some timing tests. If it proves to be a big deal, we could create a temporary table in the sync_sessions schema to hold onto the patients, so it was just a full select on a single table for each model, which I would expect to be very fast (though probably still not as fast as having the array in js memory).

@passcod do you have any comments on the postgres cache question, i.e. if you have a CTE that is included in multiple queries run in a row, is it pretty likely the result of that CTE will be in memory after the first query?

edmofro (changes requested):
I know you haven't requested re-review yet but I just took a peek :-)

edmofro (approved):
looks great!

PR #5870, opened 2024-05-15
fix(dataDeletion): Do not filter deleted records from sync

NavarroEmilioLuis:
### Changes

Even though this might appear as bloating the DB with deleted records, the thing is that we have to be able to send deleted records down, so that we can propagate deletions the right way. Otherwise, deleted records that are synced to more than one facility will still be available* when they get removed from one facility.

* "available" because they should NOT appear on the frontend, however the backend won't be erased.

We didn't catch this sooner because the sync filters were actually missing for referrals and document metadata, while encounters always synced regardless of their status (which is the way to go)


chris-bes (approved):
Makes a lot of sense! This LGTM!

PR #5880, opened 2024-05-16
feat(upcomingVaccines): nass-1260: Materialised view for upcoming vaccinations refreshed by scheduled task

dannash100:
### Changes
- Keen for input tried to get done quick
  - Currently I set last_refreshed on every column - we will need this data when doing front end - but i wonder if theres another way?  
- Gets around expensive query by refreshing table every 10 mins by default
- This will be follwed up by a front end ticket

### Deploys

### Remember to...

- ...write or update tests


edmofro (changes requested):
Super fast, super clean! Just the question you raise about having `now()` in every row...

edmofro (approved):
all lgtm!

PR #5884, opened 2024-05-16
merge: 2.6 into 2.7

passcod:
That was missed before doing the release, which has caused some builds to fail.


tcodling (approved):
lgtm 🙏  not sure how this happened

PR #5891, opened 2024-05-17
feat(omniserver): SAV-674: Provide facilityId explicitly - medium mode

edmofro:
### Changes

Where appropriate and straightforward, provide facilityId in the query or body of the request, rather than relying on the one in the token. This makes things more explicit and is more RESTful.

Note there are still some tricky cases to go, and one I'm just not going to bother with (`filterByFacilityWhereBuilder`, which is prolific in suggesters and seems like it's isolated enough to not set a hugely visible precedent, but the blast radius of changes to check through would be huge)

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Looks good just two lint errors

PR #5896, opened 2024-05-20
feat(omniserver): SAV-674: Pass facility to triage creation handler

edmofro:
### Changes

As well as moving the facility detection from config out to being passed in by the front end, I've also done some refactor so the department defaulting happens in the route handler rather than the model.

The model create override was doing far too much in my opinion. The code smells were all over the place, it was opaque, and there were specific workarounds to use it normally because it was doing something unexpected. 

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Looks great g

PR #5899, opened 2024-05-20
feat(omniserver): SAV-682: Handle multiple facilities in logging context

edmofro:
### Changes

~~I spoke with @passcod about the best way to deal with this, and the ideal here would be to have it inserted as the specific facility back at the logging call site based on that session. However, this would be a big blast radius, and the impact of not supporting that doesn't seem huge in the context of the omniserver having initially a relatively small uptake - I doubt Cambodia will opt in to honeycomb anyway. I think we can call telemetry that can disambiguate the session's facility a "feature" that's not yet supported.~~

Update: the session.facilityId will be set on the trace whenever it is available.

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Looks good!

PR #5904, opened 2024-05-20
fix(dataDeletion): EPI-565: Properly hide deleted vitals in mobile

NavarroEmilioLuis:
### Changes

This was missed unfortunately. I used the opportunity to prettify the query a bit. I think only checking for survey response answer deletion is enough given that deleting survey responses or the encounter that contain them should trigger a cascade effect


tcodling (approved):
lgtm 🙏 so much easier to read too!

PR #5925, opened 2024-05-27
fix(upcomingVaccinations): Ensure that country timezone is used when querying upcoming_vaccine view

dannash100:
### Changes
- Due to many other features using `CURRENT_TIME`, `CURRENT_DATE`, `NOW()` or `CURRENT_TIMESTAMP()`  (including thigns like signer renewal)  I am unsure about setting the timezone of sequelize to `countryTimeZone` 
- I explored other low touch ways of doing this 
  - Using some kind of input to view (not a thing)
  - Using a custom setting and using AT TIMEZONE current_setting($custom_setting) (would require allowing this config within postgres config)
- This is kinda not ideal but this works to set and revert the timezone
  - The reason why I don't want to just assume that the timezone will be reverted is because of pooling and the potential of sharing connection sessions ( This is somewhat an assumption of stuff ive read when looking into session lifetime in a sequelize instance with pooling enabled )  
  
**Short answer:** 
I think we do this as temporary solution and do a tech discussion re the change over of using sequelize timezone to country time zone and explicitly set to UTC where necessary

### Deploys

### Remember to...

- ...write or update tests


edmofro (changes requested):
Two non-substantive changes. Otherwise lgtm!

edmofro (approved):

PR #5929, opened 2024-05-27
fix(translations): release-fix-v2.8: use string util rather than component for replacement

tcodling:
### Changes

Currently our replacements logic for translations only wants to take either strings or TranslatedText components (otherwise they will show as [object Object]. Maybe we should find a way to support this in the future but for now we need to use the underlying util that generates a string for DateDisplay rather than the component itself.

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Nice! we could also default inside get date display too if we wanted
i.e = {}

PR #5932, opened 2024-05-28
repo(packageManager): SAV-644: Move to npm

NavarroEmilioLuis:
### Changes

Okay so this seems to do the trick. Now: we have to merge and release to ensure it actually works completely. Another big gotcha is the way npm handles the command line flags, I tried to go with what I thought was most simple: everything explicitly, in about the same order (prefer `--workspace` and explicitly adding the separator `--` to each script)

### Deploys


passcod (changes requested):
looks good in aggregate. a few wtfs

passcod (changes requested):

passcod (approved):
one last optional thing but other lgtm

edmofro (approved):
stamp

passcod (approved):

PR #5949, opened 2024-06-03
refactor(vaccine): NASS-1146: improve upcoming vaccination view performance

namhoangbes:
### Changes

- refactor to use join instead of row num over partition


edmofro (approved):
Thanks @namhoangbes!

@julianam-w credited with helping me understand this 😀

PR #5952, opened 2024-06-04
tweaks: EPI-825: Prevent empty first line when adding potential items

huytranbes:
### Changes

Prevent empty first line when adding potential items

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (commented):
Not much of a suggestion but I think intention is a bit unclear and we might benefit from changing the whole function flow. We don't have to push this far to preserve just one call to the setter IMO.

NavarroEmilioLuis (approved):
Thanks for applying the changes! It's a bit of a curious case that we have to deal with. LGTM

PR #5959, opened 2024-06-06
fix(adminPanel): TAMOC-195: Issue with translations

NavarroEmilioLuis:
### Changes

This adds support to having parentheses inside language names... and possibly every other quirky character we might throw at it

### Deploys


tcodling (commented):
hmm imma be honest i really dont understand what is going on here 😆  Going to try and figure it out but i do see there are some failing tests so curious if it is doing exactly what you expected?

tcodling (approved):
lgtm! thanks for catching that!

PR #5960, opened 2024-06-06
fix(appointments): Release fix v2.9: Use styled components instead of withStyles for appointment tooltip 

tcodling:
### Changes

We have been having some weird behaviour with the appointment calendar tooltip for a while now. It was bought up in regression testing that the box was showing grey (unstyled) and i discovered that this spot is the only place in the codebase using `withStyles` which is now depreciated. I am relatively certain this is the cause of the weirdness so while i was here i converted this to styled components to keep in line with our codebase and to remove the last usage of `withStyles` 

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
Nice

PR #5961, opened 2024-06-06
fix(permissions): Release fix v2.9: Fix diagnosis export permission

tcodling:
### Changes

We handle 'diagnosis' in quite a strange way. We have a bunch of places in the app where we have to handle the fact that it is sometimes referred to as 'diagnosis' and sometimes as 'icd10'. In this case it is expecting 'diagnosis' but is recieving 'icd10' so we are getting a permission error.

I would love to standardise these one day (although it feels like it was done like this for a reason?) but i feel this fix is in line with our current pattern. See `referenceDataImporter.js`, `ReferenceDataExporter.js`, `loaders.js`

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/app/admin/index.js
+++ b/packages/central-server/app/admin/index.js
@@ -65,7 +65,7 @@ adminRoutes.get(
 
     for (const dataType of Object.values(includedDataTypes)) {
       // When it is ReferenceData, check if user has permission to list ReferenceData
-      if (REFERENCE_TYPE_VALUES.includes(dataType)) {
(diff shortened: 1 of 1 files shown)


edmofro (approved):
Yeah this is nasty, but only because there's existing nastiness. Agree we should align in future, could you add that to the [maintenance sprint item](https://beyond-essential.slab.com/posts/tamanu-maintenance-sprint-items-j1ac0rkb)?

PR #5971, opened 2024-06-10
fix(mobile): TAMOC-196: Create corrective migration

NavarroEmilioLuis:
### Changes

The `deletedAt` column is missing from the `translated_string` table when upgrading the app. This happened because the feature was inside an epic branch (didn't hit main for a while) and the migration was created early on - this made it look like all was good, but there were several releases without it, causing havoc on upgrades, but not installing from scratch.

### Deploys


edmofro (approved):
Nice one @NavarroEmilioLuis, great investigation and fix!

PR #5976, opened 2024-06-11
fix(reports): release-fix-2.9: Apply startofday when explicitly filtering by a date

tcodling:
### Changes

I actually think this is an old bug that has been made apparent by the correct application of the default date range which this was compared to. I have added startOfDay from date-fns to ensure that the filter date is not filtering out stuff from the start of day due to timezone differences. 

I think it is worth down the road replacing all these usages of new Date() with a more appropriate function for our handling of dates (parseISO)? but out of scope for a late release fix

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/shared/src/utils/reports/getReportQueryReplacements.js
+++ b/packages/shared/src/utils/reports/getReportQueryReplacements.js
@@ -1,4 +1,4 @@
-import { subDays, startOfDay, subYears, addDays, endOfDay } from 'date-fns';
+import { subDays, startOfDay, subYears, addDays, endOfDay, parseISO } from 'date-fns';
 import { REPORT_DEFAULT_DATE_RANGES } from '@tamanu/constants';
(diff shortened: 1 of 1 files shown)


dannash100 (commented):
Just did a play and parseISO will do the same thing here
i.e ? parseISO(params.fromDate)

```
_.startOfDay(new Date('2015-10-10'))
// Sat Oct 10 2015 00:00:00 GMT+1300 (New Zealand Daylight Time)

new Date('2015-10-10')
// Sat Oct 10 2015 13:00:00 GMT+1300 (New Zealand Daylight Time)

_.parseISO('2015-10-10')
// Sat Oct 10 2015 00:00:00 GMT+1300 (New Zealand Daylight Time)
```

dannash100 (approved):
Beautiful

PR #5982, opened 2024-06-12
fix(settings): Add default to non-null new column in mobile

passcod:
### Changes

2.8 hotfix to mobile

When there was already data in the table, this migration was failing trying to add a non-null column with no default.

- [ ] Deploy #deploy %mobile=always


NavarroEmilioLuis (commented):
LGTM but this change then requires uninstall/reinstall the app on devices the migration might have already run (I think this is fine, just confirming we do want it - as Nauru was keen to avoid doing that).

NavarroEmilioLuis (approved):
Right cause this migration errors without it so it would never run a next step. Welp, LGTM then

PR #5983, opened 2024-06-12
fix(ci): fetch full git history to generate release notes

passcod:
### Changes

💡 Locally, we have the full git history, so git-cliff works. In Actions by default, we only pull a shallow git history, for performance, but that means that git-cliff only sees the latest commit... which is the one we just created for the release, and isn't counted for release notes. So release notes are empty!

Merging to 2.7 as that's the first branch with that workflow, will then port to 2.8, 2.9, and main


NavarroEmilioLuis (approved):
LGTM

PR #5988, opened 2024-06-12
fix: Add quotes to default scope

edmofro:


NavarroEmilioLuis (approved):
Oh wow... that's lame but yeah, can see we have antecedent in the codebase, looks good!

PR #5989, opened 2024-06-12
feat(charts): SAV-703: New program data element types for complex charts

NavarroEmilioLuis:
### Changes

Even though charts is for desktop only, we program data elements themselves will be synced down to mobile so they need to be valid

### Deploys


kujo4pm (approved):
One idea - put all the constants in alphabetical order - otherwise LGTM

PR #5990, opened 2024-06-13
fix(forms): SAV-705: Remove survey info from encounter reason

NavarroEmilioLuis:
### Changes

Some of the surveys might be sensitive, hence we're getting rid of them.

### Deploys


tcaiger (approved):
Nice one!

PR #5996, opened 2024-06-13
ci(changelog): Clean up release note output

passcod:
### Changes

Cliff was outputting way too much, and not the correct things either, so fix that by checking out the exact tag we want and instructing it to write the "--current" release notes.

However, in this mode Cliff sometimes outputs more than one version's worth of release notes, and there's no way to limit that behaviour, so post-process the release notes to keep the top section only.

While we're at it, remove "no-issue" and "release-fix-vX.Y" labels often present in PR titles but with zero value in the changelog.

Finally, the cliff template was discarding commits without a scope (`topic(scope)`), because the template syntax is obnoxious, so fix that.

### DIFF
--- a/.github/workflows/publish-release-version.yml
+++ b/.github/workflows/publish-release-version.yml
@@ -60,14 +60,17 @@ jobs:
 
     - name: Generate release notes
       run: |
-        version=${{ steps.tagged.outputs.current_version }}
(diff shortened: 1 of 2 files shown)


tcaiger (approved):
LGTM - love the new system

PR #6004, opened 2024-06-14
feat(translation): NASS-1194: Register translatable enums

dannash100:
### Changes

Register enums for tracking enums and enum selects across codebase
Involved quite a lot of refactors and some cleanups to enums based select fields so keen for a detailed review, I will be going through and QA'ing in background

- Registry uses a `Set` and compare object reference using `.has(thing)` method to enforce that in dev we get an error when creating a translated select or translated enum with no registered enum behind scenes.
- Importing the registry means we can see the state of translatable enums across codebase

- There are some edge cases that are not addressed by this pr see https://linear.app/bes/issue/NASS-1194/populate-enum-translations-through-registry
  - Some selects are set up to be treated as enums but are actually populated by reference data
  - A few selects are populated by localisation values.
 Update: These issues where addressed in [fix(translations): NASS-1127: Clean up translated ref data selects with superfluous prefix prop ](https://github.com/beyondessential/tamanu/pull/6025)

### Deploys

### Remember to...

- ...write or update tests


tcodling (changes requested):
epic stuff. I have gone through and made a few comments/questions/suggestions. 

I take it making use of this registry goes down in the other card?

tcodling (approved):
lgtm other than tha log

PR #6028, opened 2024-06-20
feat(translations): NASS-1250: Translate mobile reference data

flynncs:
### Changes

Implemented translated reference data component from web.
Swapped usage of reference data to use this component.
Note: There will be some places that are missed and need to be swapped 

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
lgtm! only thing is that i was a bit unsure about the pattern of caluclating the id and then supplying to getTranslation, however it is such a rare situation it doesn't seem worth altering the context to add a new feature

PR #6043, opened 2024-06-21
feat(invoicing): EPI-883: allow add note to invoice item

namhoangbes:
### Changes

- Ability to add note to each invoice line item


khoi-bes (changes requested):
You must add a migration for the new "note" column

khoi-bes (approved):

khoi-bes (approved):

PR #6060, opened 2024-06-27
fix(mobile): release issue 2.10 add back custom section logic

tcodling:
### Changes

Most of this pr is just restoring old logic for the custom section handling. When we were working on cambodia mode we changed up this logic to allow custom fields to be hardcoded within the other sections (general and pad). This meant that we were checking by field rather than by section in order to grab the relevant component and update the correct information. In doing this we accidentally removed the logic that allows for entire new custom sections (not used in cambodia). This just restores the old logic and renames the flag from `isCustomFields` to `isCustomSection` to make it a bit clearer and to differentiate the logic for whole custom sections and individual hardcoded custom fields

Also worth noting that this is a 2.10 release fix that i am trying to restore to its original behaviour. I know Tom has done some great work in 2.11 that resolves some of the hacky parts/unstable code as part of this feature.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/mobile/App/ui/components/Forms/PatientAdditionalDataForm/PatientAdditionalDataFields.tsx
+++ b/packages/mobile/App/ui/components/Forms/PatientAdditionalDataForm/PatientAdditionalDataFields.tsx
@@ -83,15 +83,16 @@ const CustomField = ({ fieldName, required }): ReactElement => {
 
   if (loading) return <ActivityIndicator />;
 
+  return getCustomFieldComponent(fieldType);
+};
+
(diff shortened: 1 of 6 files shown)


dannash100 (changes requested):
SOme ideas!

jaskfla (approved):
Just two minor comments from me, but neither are critical and you may well have arguments against them so happy to pre-approve 🙏

dannash100 (approved):
## giving tick because Jasper say so

PR #6077, opened 2024-06-30
fix: release 2.10: Upcoming vaccination last updated effected by timezone

dannash100:
### Changes

Treatment of dates for relative to now date display needs to just use new date iso string. Then displays correctly no matter the orientation of server timezone and clients timezone.

Just fixing sum wackyness with test
### Deploys

### Remember to...

- ...write or update tests


jaskfla (changes requested):
Just one concern with the ISO 8601 format, but nothin’ else to comment on from me

jaskfla (approved):
🙏

PR #6079, opened 2024-07-01
fix(mobile): release fix v2.10: Fix patient details heading

tcodling:
### Changes

When refactoring patient details for cambodia mode I accidentally introduced a change in the text for the new patient form. I have not correctly assigned the appropriate translated text string.

This is a great example of why i was hesitant to implement the cambodia patient details in this way. I had to make the same change in two places so hopefully we dont run into bugs in the future from these diverging files. In this pr i also added a comment to LocalisedPatientDetailsLayout to explain why it is like this as I imagine this will be a bit shocking for someone coming across it. Am also going to add this to the maintanence sprint so it hopefully gets resolved sooner rather than later

### Deploys

### Remember to...

- ...write or update tests


jaskfla (approved):
Makes sense to me!

PR #6089, opened 2024-07-02
fix: release 2.10: vaccine statuses not updating on vaccine given on mobile

dannash100:
**To fix:**
- Add back the refetch of administered vaccine on focus
- Add specificity to Vaccine cells key and id informing memo

- Fix some types as I went

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (commented):
I really don't have any context around this issue, I left a couple of suggestions here and there but nothing too serious. Final question is, how bad is performance and do we completely need this approach? Or are we missing some params we could be using to get the same out of it?

NavarroEmilioLuis (approved):
Changes look good, I made some more comments, sorry. But I don't think any of them are blockers (most are just genuine questions)

PR #6094, opened 2024-07-04
test(web): NASS-1201: `<DownloadDataButton>` tests

jaskfla:
### Changes

I had hoped to come up with a pretty comprehensive suite to test the actual data export logic/code, but ran into quite a few issues getting the `renderToText` util function to work in the testing environment, which breaks the mapping function that does all the heavy lifting in `onDownloadData` to generate the spreadsheet for export.

Hence, this testing suite is a bit of a stub. All it really does is test the rendering of the React button itself. It does manage to test the translation of that button label, though.

More importantly, it establishes a way to get `<TranslatedText>` to render in the testing environment, where the context providers that are globally accessible in Tamanu Web/Mobile are missing.

If we continue to increase testing coverage of React components, it might eventually make sense to extract out custom `render` function here—and make it create a fresh instance of the contexts (and providers) for each unit test to more effectively isolate the tests from each other.

Merge commit message: 
```
test(web): NASS-1201: `<DownloadDataButton>` tests

deps(web): add React Testing Library, remove Enzyme
```


tcodling (commented):
honestly this looks great to me! you did an amazing job with this. Its definitley not my thing tho so might be keen to discuss with another dev in town tommorow as this seems like some really foundational work!

tcodling (approved):

PR #6103, opened 2024-07-05
chore(mobile): Add local dev option to server select with special alias 10.0.2.2 to localhost 

dannash100:
### Changes

We could also do this using like a local server override or something - but this was the best way I thought to add the local host server at the start of a __DEV__ run.

I havnt thought much on this since I made up this draft, but I use this IP to save time in dev myself using a stashed server overrides file. 

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
incredible. So much better than what we have already. Good call keeping the whole list and adding to the front

PR #6107, opened 2024-07-05
fix(postMigration): VIT-227: error creating trigger name longer than 64 chars

namhoangbes:
Fix a long-standing error:
`ERROR: trigger "set_patient_program_registration_conditions_updated_at_sync_tic" for relation "patient_program_registration_conditions" already exists`
This happens because PostgreSQL's maximum identifier length is 63 bytes, which equals 64 characters. The identifier `set_patient_program_registration_conditions_updated_at_sync_tick` exceeds this limit and is automatically truncated to `set_patient_program_registration_conditions_updated_at_sync_tic`. As a result, the table is falsely detected as not having a trigger, leading to re-creation of the trigger and causing the error.

### DIFF
--- a/packages/shared/src/services/migrations/runPostMigration.js
+++ b/packages/shared/src/services/migrations/runPostMigration.js
@@ -44,7 +44,7 @@ const TABLES_WITHOUT_TRIGGER_QUERY = `
       FROM
         pg_trigger p
       WHERE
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (approved):
Nice! Thank you very much for solving this one. Want to be very cautious but I can't think any other safety measure.

I guess let's just make sure this gets explicitly tested when doing the release? Can you make it a card to track it?

PR #6124, opened 2024-07-10
fix(surveys): TAMOC-210: Handle surveys with dates on mobile

NavarroEmilioLuis:
### Changes

This properly handles submitting surveys, referrals and vitals in mobile

### Deploys


dannash100 (approved):
Looks good some non blocking comments

PR #6129, opened 2024-07-11
fix(mobile): release 2.11: Mobile patient details form broken

tcodling:
### Changes

We made a tweak (adding a section key for comparison logic now that the title is translated) to the Cambodia patient details layout on mobile that we didn't carry over to the generic layout. This bug is a result of that.

This is another example of why i was somewhat opposed to this design/hack for the "cambodia" patient details as it is very bug prone.

### Deploys

### Remember to...

- ...write or update tests


jaskfla (approved):
Left two small questions, one about the return type from a map—but since the bit I’m asking about exists in the old code too I figured it’s safe to approve

jaskfla (approved):
You found that merge impressively quickly! Does make more sense now 😎

I think the consensus was to remove the explicit `void` return type in line 88, but that’s less mission critical, of course

PR #6130, opened 2024-07-12
fix(translations): NASS-1295: fallback when replacement is loading

jaskfla:
### Changes

Now falls back to the figure dash ( ‒ ) when the total page count is still loading.

Does **not** fix the current page number not updating. There’s some existing logic there, which doesn’t seem to be working on my system, but I didn’t want to mess with it. Figured it probably deserves its own card.

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
tiny optional tweak but otherwise nice one! Also just need to rerun that flaky test

PR #6133, opened 2024-07-14
fix(genericSurveyReport): TAMOC-208: Do not include test patient in report

NavarroEmilioLuis:
### Changes

Production deployments should now have a common test patient with a stable UUID, this patient should be removed from reports. The scope of this card is to only include this report.

### Deploys


edmofro (changes requested):
Just one suggestion, I think this should be a named constant with a comment saying it's a stable UUID for the test patient

edmofro (approved):

PR #6138, opened 2024-07-15
fix(facility): release fix 2.11: Error codes instead of server message

khoi-bes:
### Changes

In the khmer certificate font pr, I introduced some interceptor promise logic to the TamanuApi. This resulted in server errors coming through as just codes rather that the full server error string. Here's the pr for fixing it

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/api-client/src/TamanuApi.js
+++ b/packages/api-client/src/TamanuApi.js
@@ -159,7 +159,7 @@ export class TamanuApi {
         responseInterceptorChain[j++],
       );
     }
-    await responsePromise;
+    await responsePromise.catch(() => {});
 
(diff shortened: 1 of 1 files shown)


tcodling (changes requested):
Thank you so much! Only thing is this needs to be pointed at v2.11 instead

tcodling (approved):

dannash100 (changes requested):
Looks like this has changes from main we dont want on 2.11
You could rebase or just make the change in a new pr off 2.11 probably easier :)

dannash100 (approved):

PR #6158, opened 2024-07-17
fix(translations): translate straggling “Loading” label

jaskfla:
### Changes

Translates a straggling untranslated “Loading…” label

### Deploys


dannash100 (approved):
Approved with small request

PR #6169, opened 2024-07-19
fix(test): NASS-1079: fix Storybook

jaskfla:
### Changes

Fixes Storybook, including `test-storybook` script.

Does not fix stories themselves. There are a few broken ones, perhaps due to the component interfaces drifting away from what’s demonstrated in the stories?

**Also**, as part of this fix, scripts used with `watch-shared-an` can now take any number of arguments—not just a single command. For example, from root, you can now run
```bash
yarn storybook --port 6006
```
and the arguments `--port 6006` will be passed along to `watch-shared-and`

### Notes

The next time we update to Storybook will be a bit of a rigmarole.

v7 deprecated `storiesOf` and v8 has completely removed it. Storybook does provide some migration scripts (which are `npx`-able), but it will only get us partway there. This is because it follows a new paradigm that relies on statically analysable code, which a bunch of `default` exports.

Their script transforms our stories into ES modules with multiple default exports, so needs some manual intervention to split into separate files.


dannash100 (commented):
Yes looks good! Can you check that the vaccine cert loads correct still 🙏

dannash100 (approved):
Giving tick agian for recent changes :)

PR #6172, opened 2024-07-22
feat(diagnosis): EPI-811: Add clinician name to diagnosis workflow

huytranbes:
### Changes

Add clinician name to diagnosis workflow

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (changes requested):
Hey this looks good to me but I think there is one issue with UI logic and also would prefer to remove redundancies in migrations - this last one isn't completely on you as there was a precedent, but if you can, please check if mobile can do without it too. Otherwise should just apply to server.

NavarroEmilioLuis (approved):
Thank you for checking and changing those! Looks great!

PR #6187, opened 2024-07-23
perf(web): prefer WOFF2 font

jaskfla:
### Changes

Took the TTF files that are already in `/fonts/` and compressed them to WOFF2. If the user’s browser supports it, this reduces the weight of Fira Code from 177 kB to 72 kB

Small potatoes in the grand scheme of things, but was such a trivial optimisation

(Leaving Khmer fonts are they’re due to be removed)

### Deploys


dannash100 (approved):
Love it! Worth testing in deployed environment that they load correctly

PR #6195, opened 2024-07-23
tweak(mobile): SAV-771: Misc fixes

NavarroEmilioLuis:
### Changes

So most of this is `style` but at least the migration file is a true fix. We probably never encountered any issues because we never have run down migrations on mobile?

In here:
- Specify <Partial> when using a subset of models (MODELS_MAP)
- Fix migration file by passing the column name instead of the foreign key object
- Change the way to check if a prop exists before using it
- Specify <Omit> key from type
- Create and use a specific type for an array of models

### Deploys


rohan-bes (approved):
Thanks for the tidy up! :pray: 

A couple of comments, but nothing major so pre-approving :+1:

chris-bes (approved):

PR #6206, opened 2024-07-25
feat(web): EPI-913: Documented allergy to appear in red

khoi-bes:
### Changes

- Add new setting for highlight allergy
- Show allergy in red color if matching the setting

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (changes requested):
Hey so my comment was longer than expected. I think it would be better if we changed the strategy a bit. I know this works as is but seems like a good opportunity to make some arguments - great that it's using the settings table, too.

NavarroEmilioLuis (approved):
Changes look great! I'm guessing `undefined` means it'll fallback to default so that's also good.

PR #6212, opened 2024-07-26
fix(invoicing): release-2.13: EPI-948: Change Insurer payment importer to use Invoice number

namhoangbes:
### Changes

Update the invoice number generator function and also change the insurer payment importer to use invoice number instead of invoice id

### Deploys


khoi-bes (approved):
LGTM!

PR #6225, opened 2024-07-31
feat: EPI-905: Create new multiselect filter component

khoi-bes:
### Changes

- Create new multiselect filter component and apply for 'Related condition' and 'Status' search fields in the program registry patient search table
- Update corresponding api

### Deploys

### Remember to...

- ...write or update tests


tcodling (changes requested):
Lgtm other than some small linting issues to resolve 🙏 

Was thinking of maybe suggesting abstracting out the styling as it seems a lot of it is shared with MultiSelectField however I think this would just make it even more confusing so probably better to leave.

tcodling (approved):
lgtm! Would be good to remove those electron comments but shouldn't be a blocker 🙏

edmofro (approved):
Much better, thanks @khoi-bes

PR #6226, opened 2024-07-31
test(sync): SAV-752: Add script to benchmark performance of sync

NavarroEmilioLuis:
### Changes

Okay so the strongest addition here is just to bring the `shell` command to facility server. This allows us to run JS scripts easily withing the app context.

I think this script is a good first step towards sync benchmarking - the simulation happens at the facility level. This does not help us to stress test facilities, but that's not the goal here (though I'd argue it'd be a good template when we do that).

### Deploys

### DIFF
--- a/packages/facility-server/app/index.js
+++ b/packages/facility-server/app/index.js
@@ -9,6 +9,7 @@ import {
   migrateAppointmentsToLocationGroupsCommand,
   migrateCommand,
   reportCommand,
+  shellCommand,
   startAllCommand,
   startApiCommand,
   startSyncCommand,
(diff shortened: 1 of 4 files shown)


edmofro (changes requested):
Ok it's not the finest toothed comb review of all the details of each random generator, but overall this is looking great! I have a question about whether there's much overlap with the faking done for migration determinism testing? Conceptually it feels like there is, I wonder if we can have the migration determinism fakers support the random generators in here somehow, as those are required to be kept updated as new models are created so would make it easier to add more in here too. Could you look into it?

edmofro (approved):
Yep all good! I would slightly push for changing the name `shell` to `nodeRepl` or `nodeShell`, but probably it's only me who'll make that mistake. Approved and over to you on that one!

PR #6234, opened 2024-08-01
fix(web): release 2.13: can not export inpatient data

khoi-bes:
### Changes

In Inpatient table, we can't export the data due the diet column, because it's rendering a tooltip and a tooltip require only one child (as object) instead of an array of children, so it will cause error if you pass the array of children to the tooltip

### Deploys

### Remember to...

- ...write or update tests


rohan-bes (approved):
Thanks for the fix :pray:

PR #6256, opened 2024-08-07
feat(sync): SAV-793: Restart update sync_lookup table process if it takes too long

chris-bes:
### Changes
- Restart update sync_lookup table process if it takes too long

### Deploys

### Remember to...

- ...write or update tests


edmofro (commented):
I really appreciate picking this up but was there a reason you think we'll need it? I was hoping we can forget about the stalled syncs entirely! Do you have a suspicion the same random long-snapshots will show up in the sync lookup, and restarting will get rid of them?

edmofro (commented):
I've decided this is actually amazing, thank you @chris-bes fantastic insurance policy that fully solves the potential clogging up of resources.

Does it throw an error that is nice, or should we catch and gracefully handle setting a nice error message on the sync session?

edmofro (approved):

PR #6265, opened 2024-08-09
fix(multiselect): TAMOC-218: Filter by label

NavarroEmilioLuis:
### Changes

Okay so this change should correct only program registries multiselect while keeping everything else the same. This way we can change places one by one, as needed.

### Deploys


chris-bes (approved):
LGTM! Searching by label should definitely be the default behaviour!

PR #6266, opened 2024-08-09
feat(omniserver): Import/export user_facilities using custom loader

tcodling:
### Changes

This is mostly copied from the labTestPanelLabTestType import/export logic. 

The main difference here is that I needed to figure out how to delete records facilities from the join table as it appears we haven't implemented logic like this yet. Not 100% sure if I put it in the best place but I essentially used the loader to add a 'allowedFacilityIds' property to the row, then when the row for that user is being processed, i add 'deleted' UserFacility rows for all facilities linked to that user not contained in this row.

### Deploys

### Remember to...

- ...write or update tests


dannash100 (commented):
Much better I love it! Just have question re supplying the generated id for create

dannash100 (approved):
Beautiful work and great improvement

PR #6267, opened 2024-08-09
fix(web): NASS-1271: fix exports on slow networks globally

jaskfla:
### Changes

- Updates `saveFile` to accept an async data fetcher as parameter instead of directly receiving the data/blob
  - This lets `saveFile` immediately get a file handle, rather than waiting for the file data to be prepared before opening the Save File dialog. (Waiting too long causes a SecurityError timeout due to [Transient User Activation](https://developer.mozilla.org/en-US/docs/Glossary/Transient_activation).)
  - Getting file handle first, then preparing file data is how [Chromium engineers recommend handling file downloads](https://issues.chromium.org/issues/40175286)
- File preparation can take place asynchronously after user has been presented with Save File dialog
- Updates all usages of `saveFile` to match this new interface
- Fixes `Failed to execute ‘showSaveFilePicker’ on ‘Window’: Must be handling a user gesture to show a file picker` error in all places **except report export**. No regressions, but error can still happen here because the default filename is derived from the API call that retrieves the data. #6297 fixes this.
- Shows a toast when file download is complete (For clarity: not when Chrome puts empty `.cswrap` file on disk)
- Fixes NASS-1264, NASS-1270, NASS-1271

### Deploys


dannash100 (approved):
Beatiful work - I got some tiny comments, but nothing to block you

PR #6268, opened 2024-08-12
fix(omniserver): NASS-1304: Remove last usages of facilityId from request

dannash100:
### Changes

Use facilityId from body or query instead of from request

Thomas:
Also we combined the `facilityId` + `markedForSyncFacilityId` passed to patient search as they will never be different in theory (at least in our current implementation)
- `facilityId` = the current facility used for filtering the patients so we can find beds used in this facility
- `markedForSyncFacilityId` =  used in query to get "markedForSync" status from `patient_facilities` to see if they are synced to this facility

I cant see any reason that these should be different. Although i do think patientSearch endpoint could use a bit of a cleanup to be clearer what is going on

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/__tests__/apiv1/Patient.search.test.js
+++ b/packages/facility-server/__tests__/apiv1/Patient.search.test.js
@@ -159,15 +159,17 @@ describe('Patient search', () => {
   });
   afterAll(() => ctx.close());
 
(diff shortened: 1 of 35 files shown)


jaskfla (approved):
Looks good—liking the consistency of the `markedForSyncFacility` → `facilityId` rename 🎉

PR #6270, opened 2024-08-12
fix(mediciReport): Fix missing locations in location history for Medici Report (hotfix 2.9.5)

rohan-bes:
### Changes

We previously we relying on a `select distinct on (encounter_id)` to get the locations, which would lead to selecting the most recent location but not the history. Now using a `json_agg` much like other history fields

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (commented):
I think this looks OK from a code perspective. But just giving you a heads up that the change seems whack?

NavarroEmilioLuis (approved):
LGTM

PR #6273, opened 2024-08-13
fix(immunisationsTable): release-2-14: Fix query

NavarroEmilioLuis:
### Changes

This one was painful to debug as there were no apparent changes to the code involved in the table. The SQL verbose output also wasn't good enough because the query was so long it didn't show up complete. Ended up doing a binary search between commits to find the culprit, only to scratch my head a little longer.

This is the PR that introduced the error: https://github.com/beyondessential/tamanu/pull/6213 and it's where we add diets as its own thing.

It didn't show up on our unit tests because we didn't have coverage with the table pagination API, which is ultimately what causes the issue. It seems to works great everywhere else when including the diet association, however, when limiting the query for some reason all goes south. The error complains about a missing scheduledVaccine table that is definitely there... I don't fully understand what is going on, so I decided to just smash the error - the query is complicated enough and this is a release fix.

Although tempting, I did not want to just exclude `diets` from the include, as including everything is what got us into this mess. I tracked everywhere where we use the `getAdministeredVaccines` method and realized we only ever use a nested include from encounters.

In the first commit, you can see a facility server test failing, which is the one this PR aims to fix

### DIFF
--- a/packages/facility-server/__tests__/apiv1/PatientVaccine.test.js
+++ b/packages/facility-server/__tests__/apiv1/PatientVaccine.test.js
@@ -267,6 +267,14 @@ describe('PatientVaccine', () => {
       expect(result.body.count).toEqual(5); // there are 5 recorded given vaccines in beforeAll
     });
 
+    it('Should get administered vaccines with pagination', async () => {
+      const result = await app.get(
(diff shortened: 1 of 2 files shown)


chris-bes (approved):
LGTM! Thanks for double checking the usage!

PR #6276, opened 2024-08-13
fix(invoices): EPI-950: port to release 2 14

NavarroEmilioLuis:
### Changes

This was merge to main after the cutoff and we missed it!

Also, used the time to check whether we missed anything else from current 2.13 to 2.14 but nope, just this one


dannash100 (approved):
# OHK OHK

PR #6297, opened 2024-08-16
fix(web): NASS-1271: fix report export timeout error

jaskfla:
> [!NOTE]
> This one branches off #6267.

### Changes

Applies the fix from #6267 to report exports from the admin panel.

(Previously, the suggested filename came from the API response. This forced us to wait for the API to finish responding to get a file handle. On slow networks, this can take too long and cause a SecurityError when presenting the Save File dialog.)

### Deploys


tcodling (approved):
lgtm!

PR #6298, opened 2024-08-16
fix(sync): SAV-806: Fix sync status toast

edmofro:
### Changes

Fixes the two issues reported in SAV-806

1. Our sync session start time was being recorded from part way through the sync logic, and notably was after some async work, so there was a significant period when the sync status would note the sync as running, but without a valid time.
2. The time "ago" was comparing Date.now() on two different computers, the facility server and the client. These clocks can be slightly out, so it wasn't accurate, and in this case was producing negative numbers.

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
LGTM

chris-bes (approved):
code owner stamp

PR #6320, opened 2024-08-21
feat(pm2): calculate default memory and scale from os resources

passcod:
### Changes

Using `'max'` spawns as many threads as there are CPU cores, but we explicitly don't want that, because we typically co-locate Postgres on the same machine as Tamanu, and also the task-runner is extraneous.

But the pm2 config is a Node.js script, so we can in fact query the actual number of cores and set the default amount of instances as half of available, or a minimum of 2.

And then just because, we also query the total available memory and use that to define the Node.js allocation, instead of hardcoding as 8G.

### DIFF
--- a/packages/central-server/pm2.config.cjs
+++ b/packages/central-server/pm2.config.cjs
@@ -1,4 +1,10 @@
-const memory = process.env.TAMANU_MEMORY_ALLOCATION || 8192;
+const os = require('node:os');
+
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (changes requested):
Lots of questions to avoid shooting ourselves in the foot. Hopefully. Also one change request cause I think this will break as is!

NavarroEmilioLuis (approved):
Great!

PR #6333, opened 2024-08-25
epic: Greenmask config

passcod:
- KAM-251
- KAM-252
- KAM-253
- KAM-254
- KAM-255


edmofro:
@passcod is this still a thing that should be open?

PR #6334, opened 2024-08-25
fix(mobile): Make survey notifiable default 0 and not false

dannash100:
### Changes

We saw some errors on mobile build when klaus was testing on bluestacks.
sqlite expects 0,1 to represent boolean values

I think its the issue? altho it doesn't error on my local
image (61)

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
strange that its not erroring on your local 🤔 looking at other migrations though it does seem like this is the correct way to declare this though so i think a good change. maybe get klaus to test on this pr if unsure

PR #6336, opened 2024-08-26
fix(sync): SAV-810: Change id generator HOTFIX 2.9 (#6329)

chris-bes:
SAV-810: Change id generator HOTFIX 2.9


edmofro (approved):

rohan-bes (approved):
Just had one comment about a very edge case scenario that I would be interested to hear your thoughts on. But all looks good!

PR #6338, opened 2024-08-26
feat(tasking): EPI-932: task sets importer

namhoangbes:
### Changes

- Alter reference data relatons unique index
- Throw human readable error on in task set importer
- 
### Deploys


khoi-bes (changes requested):
If you check this file `importRows.js` in line 212th, you can see the way it get the existing record in dtb to determine this record should be updated or created. With the new refactor, we should update the `existingRecordLoaders.ReferenceDataRelation` to ensure it can get the correct existing record, and we also need to be careful here as it may affect the current behavior when importing ReferenceDataRelation directly

khoi-bes (approved):

PR #6339, opened 2024-08-26
feat(tasking): EPI-936: build mark task as completed workflow

namhoangbes:
### Changes

- implementing task completion workflow

### Deploys


dannash100 (approved):
Nothing blocking approval just some thinking. Nice!

PR #6345, opened 2024-08-27
fix(mobile): release 2.15: Fix bug with large autocomplete results

tcodling:
### Changes

We discovered that if a reference data type has a lot of records. It will throw errors when being searched with an empty string (or any query that returns more than 999 values) due to a sqlite limitation when inserting ids into a query. To get around this i actually a added a limit of 25 to the suggester so that it acts the same as desktop. 

Currently even though its an autocomplete field, it actually returns every value in the db so this is just updating to reflect what we do on desktop.

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Yes great idea using same logic as desktop 🙏

PR #6352, opened 2024-08-28
ci(android): fix android workflow failing on main

passcod:
### Changes

The android workflow fails when running on main because there's no PR to post a comment to.


NavarroEmilioLuis (approved):
LGTM

PR #6353, opened 2024-08-28
fix(survey): EPI-970: missing visualisationConfig column in vital survey export

namhoangbes:
### Changes

- Add visualisationConfig column in vital survey export

### Deploys


khoi-bes (approved):
LGTM!

PR #6356, opened 2024-08-28
fix(appointments): release 2.15: Status bug breaking styling

tcodling:
### Changes

Discovered on regression testing that some of the logic depending on 'status' was breaking on the appointments calendar. This was because the options were changed in a way where it was essentially passing around a capitalised version of the constant which was not triggering logic when it needed. Here i fixed the options array

This still needs to be incorporated into a "translated" select but i think out of scope of this release fix

### Deploys

### Remember to...

- ...write or update tests


jaskfla (approved):
One suggestion, but quite small so pre-approving 🙏

PR #6359, opened 2024-08-29
feat(pm2): opt into multi-process facility deployment

passcod:
A series of small improvements to our PM2 configs:

- bringing the memory calculation to facility server like in #6320 
- using the canonical `startApi` instead of `startServe` alias
- shortening process names so they're easier to type/read without losing meaning
- using the lower-overhead fork mode
- using a single line to set cwd

This last one means that we can more easily have fully-custom pm2 configs for deployments that persist between upgrades by copying the config to outside release (e.g. to `\tamanu\pm2.config.cjs`) and adjusting the cwd line to

```js
const cwd = 'release-v2.16.0/packages/central-server';
```

and then upgrading by changing that version.

---

Some minor code/config changes:

- change the default facility sync port to 5000 instead of 4001 (so it doesn't conflict with facility api scaling up)
- always respect the standard `PORT` environment variable over the config files

---

And finally a deployment process change:

- change facility-server to use multiple processes (1 for sync, 1 for tasks, 2+ for api)

This part requires a caddy change, but will enable no-downtime restarts of facility when changing config, especially for sync and tasks, and allow us to tweak load-balancing policy.


rohan-bes (approved):
Looks good @passcod! I'm also happy to back out my change and put up a separate PR once this one has been merged. It feels like it might be a little out of scope (being a functional change that really just affects the FHIR stuff)

Harnesh787 (approved):

edmofro (approved):

PR #6360, opened 2024-08-29
repo: add kamaka codeownerships

passcod:
### Changes

- for PM2 configs, to avoid missing changes to deployment processes which may require e.g. caddy or container configuration
- for Github Actions, as a safety

Maybe we should have a github kamaka team so it doesn't fall to only me.


edmofro (approved):
There's now a team, approved either as is or with the handle swapped out

PR #6373, opened 2024-08-29
feat(invoicing): EPI-975: add support for imaging area pricing

namhoangbes:
### Changes

- add support for imaging area pricing

### Deploys


dannash100 (approved):
Looks good. Is the lack of permissions something we need to consider?

PR #6376, opened 2024-08-30
feat(container): cleanup obsolete functionality

passcod:
### Changes

Cleanups to the container image removing old features and using new ones.

- Use `PORT` to set the port instead of the `local.json5` file
- Read configuration from `/config` and package source [directly with `config`](https://github.com/node-config/node-config/wiki/Configuration-Files#config-directory), which eliminates needing to copy files in the entrypoint, and allows some advanced uses.
- Remove obsolete entrypoint section that wrote OTEL variables based on the `/meta` path, which has long since disappeared.
- ~~Remove healthcheck from entrypoint, which was only used by ECS (K8S uses service healthcheck).~~ Reverted that removal, so that ECS keeps working.

### Deploys


dannash100 (approved):
Looks good :)

PR #6316, opened 2024-08-20
feat(settings): NASS-1330: move subset of config to settings (part 3)

jaskfla:
### Changes

👉 See NASS-1330.

### Deploys

- [x] **Deploy to Tamanu Internal**  %config=nofeatures

PR #6380, opened 2024-09-01
refactor: named params for makeCertificate functions

jaskfla:
> [!NOTE]
> Merges into #6316, because this change came out of that review

### Changes

Changes the signature of `make...Certificate` functions to use named parameters (it was beginning to get unwieldy).


tcodling (approved):
lgtm! i love this change.

PR #6383, opened 2024-09-01
fix(settings): `slidingFeeScale` default

jaskfla:
> [!NOTE]
> Spotted in review for #6316, so merges into that

### Changes

- `slidingFeeScale` default value did not match schema for the setting
- moves setting to where all defaults are defined under new settings structure

### Deploys


dannash100 (approved):
Amazing work

PR #6390, opened 2024-09-03
feat(adminPanel): NASS-1323: non-generic sidebar icons

jaskfla:
### Changes

Not all the icons used are included in MUIv4’s set of Material Icons, so there’s a mixture of `<SvgIcon>`s and plain ol’ `.svg` files that get linked as ``s like the other icons

It’s slightly more React-y and lets us recolour icons at render-time, but I might actually prefer the simplicity of having them all as plain SVGs? Let me know what you think. **Do check out https://github.com/beyondessential/tamanu/pull/6390/commits/00e39a06c65c2e2475a22e69d3cd3ee8b3afe91a in particular**

(The comprehensive solution would be to update MUI, but that’s clearly out of scope)

Screenshot 2024-09-03T131838

### Deploys


dannash100 (approved):
Looks gud!

dannash100 (changes requested):
Seems to be relevant test error 🤔

dannash100 (approved):
Yay amazing work

PR #6403, opened 2024-09-05
feat(appointments): NASS-1381: appointment tiles & status tokens

jaskfla:
### Changes

```bash
yarn run storybook -p 6006
```

Then 👉 http://localhost:6006/?path=/story/appointments-appointment-tile--confirmed

Just note that in Storybook they appear hilariously wide. This is expected, because I think the width is more appropriately set where it’s used, not at the component level. (There could be an argument for `min-` and `max-inline-size`, though, but I think appropriate values for those will come as we start using these components)

### Deploys

### Remember to...

- ...write or update tests


dannash100 (changes requested):
Few small changes - most are knit pick but Date stuff is requested.
Amazing detailed work really love ya work 💯 💯

dannash100 (approved):
Looks great thank you!

PR #6405, opened 2024-09-05
feat(dbmodel): KAM-260: Add a script to generate source model from database

rail-rain:
### Changes

This adds a JavaScript to generate a Source model by reading DB schema from Postgres. This is better because running the original `generate_source` requires Python and a full dbt project, which is a lot for a this small task.


passcod (approved):
looking great!

PR #6412, opened 2024-09-06
feat(referenceData): NASS-1324: add new reference data type for booking type

jaskfla:
### Changes

Defines `bookingType` reference data type

### Deploys


dannash100 (approved):
# 2 ez

PR #6426, opened 2024-09-11
feat: remove meta server

passcod:
### Changes

It now lives out of tree: https://github.com/beyondessential/tamanu-meta-server


dannash100 (approved):
💯💯

PR #6430, opened 2024-09-12
tweak(greenmask): reclassify to sensitivity levels

passcod:
As per discussion in https://linear.app/bes/issue/MAUI-4577/test-out-new-greenmask-dump-of-demo2


edmofro:
@passcod bump on this too

PR #6435, opened 2024-09-12
fix(sync): Fix syncAllLabRequests encounters sync (Release 2.16 fix)

chris-bes:
### Changes
- When an encounter is first created, the sync lookup data will have `isLabRequest` = false. When a new lab request is attached to it, SyncLookupRefresher currently does not pick up that encounter and change `isLabRequest` to true, causing FK constraint issue when syncing down because the encounter is not synced. This PR is to update the update lookup table logic to also include the encounters when a lab request has been attached to it

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
+++ b/packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
@@ -970,7 +970,7 @@ describe('Sync Lookup data', () => {
 
     let regularSyncPatientsTable;
 
-    beforeAll(async () => {
+    beforeAll(async () => {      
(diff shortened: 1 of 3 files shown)


NavarroEmilioLuis (approved):
Ohhh so this is what we were seeing!! Looks good to me, left a comment with no real action to take.

I was going to say that it would've been great if we only perform this join if the feature is on, but unsure if it's a simple as changing the query or if it would get in the way of the lookup logic.

edmofro (approved):
I have a suggestion but it is not substantive, and I might not be able to check later, so approving this with my codeowner stamp and happy for you and @NavarroEmilioLuis to take it from here

PR #6436, opened 2024-09-12
ci(android): ignore error for non critical steps

passcod:
### Changes

Steps that post comments or add links to releases should not block uploads to S3 or builds in general if they fail.


NavarroEmilioLuis (approved):
LGTM

PR #6466, opened 2024-09-18
db(docs): fully document database with dbt-compatible model

passcod:
### Cards

- KAM-260
- KAM-261
- KAM-264
- KAM-281
- KAM-282
- KAM-291

# DBT model generation & checking

```
yarn dbt-generate-model
```

The core tool is a script which generates and updates a dbt-compatible "source model", which is a YAML format specifying all the schemas, tables, and columns present in the database, along with some "data tests" such as relationships, accepted values, and non-null/unique.

There's one YAML file per database table. The tool generates the model by querying the local database, creating missing table files, adding and removing columns, adjusting field types, etc. It leaves everything it doesn't explicitly controls alone: additional data tests, unknown fields, and most importantly descriptions.

Because there's sometimes differences between the facility and central server schemas, two source models are kept, one for each server type. However, to reduce maintenance burden, if the facility and central files are found to be identical, then the facility one becomes a symlink to the central one instead.

Alongside each YAML file is a MD or markdown file in a specific format:

```markdown
{% docs table__table_name %}
Description of the table.
{% enddocs %}

{% docs table_name__field_name %}
Description of the field.
{% enddocs %}
```

Filling these sections adds documentation to the tables and their fields.

## TODOs

```
yarn dbt-check-todos
```

When the model generator finds new columns and tables not present in the files, it will add them to the YAML, and will also add the relevant markdown files or sections. It places a `TODO` in those sections:

```markdown
{% docs table__table_name %}
TODO
{% enddocs %}

{% docs table_name__field_name %}
TODO
{% enddocs %}
```

The check-todos script reads the source models and lists every TODO that remains. CI will not pass if there are any TODOs left, so all PRs must fully document new tables and columns to be able to merge.

## Docs

```
yarn dbt-generate-docs
```

This configures and generates a dbt docs static site for the source models. In CI, this is used to verify the syntax of the source model is correct and compatible with dbt.

# Documentation

The bulk of this PR is:
1. an import of the original "source models" that Team Maui painstakingly wrote and documented by hand
2. the result of the tool described above generating TODO-filled files for every single missing table and column
3. the verification and documentation of every single table and field, by area of interest

Essentially, once this PR is merged, the Tamanu database will be fully documented, and no further PR will be able to land without maintaining this 100% documentation level.

The dbt source model will also be integrated back into Maui's activities; notably while developing this PR a number of mistakes and inaccuracies were discovered in the original files (columns that no longer existed, broken or incorrect relationships, imprecise, too-lax, or too-strict value checks) and of course the full schema validation done as part of this PR has also added a lot of coverage!

Further integrations are planned, such as:
- validating the greenmask database anonymisation configuration against this source model, to ensure it always remains up to date
- making migration determinism checking smarter
- publishing the database docs on release
- syncing the model to constants


edmofro (approved):
This is brilliant stuff. Giving it a stamp as all the individual PRs have been reviewed separately I believe

NavarroEmilioLuis (approved):
Thanks, and sorry!

PR #6469, opened 2024-09-19
fix(pm2): add restart delay to avoid using 100% CPU when boot-looping

passcod:
### Changes

When the process fails to start immediately, pm2 has a default restart delay of 0ms, so it uses 100% CPU for no useful reason. A 5s wait between restarts is reasonable, and will ensure we can actually use the machine (e.g. to fix the problem).

Second commit brings over the task() abstraction to central which reduces a lot of the code duplication.


dannash100 (approved):
Perfect just the lint

PR #6487, opened 2024-09-23
feat(dbmodel): KAM-281: Make source model generation edit-aware

rail-rain:
Running the script causes 400+ changes to `database/`. I inspected some while testing, but there may be some edge cases I missed.

### Changes

Adds a script that reads the DB and update the dot source models to match.


passcod (changes requested):
just a few minor notes

passcod (approved):

PR #6488, opened 2024-09-23
fix(sync): SAV-828: Sync task should be in the same process with tamanu-sync

chris-bes:
### Changes
- Sync task should be in the same process with tamanu-sync

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/app/subCommands/startAll.js
+++ b/packages/facility-server/app/subCommands/startAll.js
@@ -14,6 +14,7 @@ import { startScheduledTasks } from '../tasks';
 import { version } from '../serverInfo';
 import { ApplicationContext } from '../ApplicationContext';
 import { createSyncApp } from '../createSyncApp';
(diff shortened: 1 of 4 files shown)


NavarroEmilioLuis (changes requested):
I'm not very familiar with this separation, but I think this needs to be modified a bit?

NavarroEmilioLuis (approved):
Oh okay gotcha. LGTM, might want to get a second opinion

edmofro (approved):
I was thinking we would change the SyncTask to call the api rather than move the task itself into the same process. However this is simpler to implement, and probably just as cohesive, so approved from codeowner perspective!

passcod (approved):

rohan-bes (approved):
Approving since it looks like this will work.

A few refactor/code cleanup changes suggested but perhaps they could be address when we merge back to main :+1:

PR #6503, opened 2024-09-24
feat(invoicing): EPI-1012: implementing tracking for user that record patient payment

namhoangbes:
### Changes

- Implementing backend logic for tracking user that record patient payment

### Deploys


NavarroEmilioLuis (commented):
Hey! Thinks LGTM, I have one small question that might need a change and also a suggestion about including a unit test with this.. I would expect we already have a unit test for this model, so it should be just checking that it gets included? It's a bit trivial so I'm okay if it doesn't get added

PR #6510, opened 2024-09-25
fix(programRegistry): 2.17 fix: Error on registration deletion on program registry table

dannash100:
Tiny change required when updating registry related to omni server change.

Basically because it creates a record each time, it needs prerequisite info for registry endpoint

### Deploys

### Remember to...

- ...write or update tests

fix: upsert invoice


NavarroEmilioLuis (approved):
LGTM... we need the ID to be able to update later because we could have the same name or something?

PR #6513, opened 2024-09-26
fix(vaccine): 2.17 fix: Mobile missing scheduleVaccineId on multiple edits without returning to vaccine table

dannash100:
### Changes

To reproduce issue:
- Administer vaccine
- View vaccine
- Click edit and change to `NOT GIVEN`
- Do not click back to table and instead click edit agian
- Change to `GIVEN`
- Administered vaccine is missing its scheduledVaccineId

Ive played with this for a few hours and the issue is that `vaccine.scheduleVaccine` becomes null after multiple updates without returning to table view. This means that if it is submitted the table will error due to Administered vaccine with null scheduledVaccineId

My feeling is that this is an existing issue, having existed as far back as implementation of vaccine edit changes. It was noticed in 2.17 

The scheduled vaccine is already fetched in the submit so easily enough to pass that through to the subsequent load.
This fixes the issue in the cleanest way I could find.

 ~~Still I would really love to get to the bottom of why it becomes null in first place, but I feel pretty happy with this as a solution with the least risky changes to logic.
Keen for discussion.~~ 

**Update** latest commit addresses underlying issue! Basically the administered vaccines link to scheduled vaccine wipes the key. - as scheduled vaccine is not included into administered vaccine!

Leaving note to back apply where needed to effected versions.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/mobile/App/ui/navigation/screens/vaccine/newVaccineTabs/NewVaccineTab.tsx
+++ b/packages/mobile/App/ui/navigation/screens/vaccine/newVaccineTabs/NewVaccineTab.tsx
@@ -135,7 +135,6 @@ export const NewVaccineTabComponent = ({
         navigation.navigate(Routes.HomeStack.VaccineStack.VaccineModalScreen, {
           vaccine: {
             ...vaccine,
-            scheduledVaccine: scheduledVaccineRecord,
             administeredVaccine: {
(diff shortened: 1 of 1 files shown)


jaskfla (approved):
Phwoar, looks like a tricky one to catch 👌

PR #6521, opened 2024-09-30
fix(bugsnag): make integration functional

passcod:
### Changes

- The bugsnag code was plain broken because I forgot how dynamic imports work.
- Once that was fixed, a few other issues cropped up:
- The bugsnag middleware broke logins proxied from facility to central.
- Bugsnag was only used in servers, not in the frontend.
- Fixing that involved inventing the ability to have an environment variable passed to the frontend code in production (or in the kubernetes deploy).
- Once that capability was available, it became possible to update the NODE_ENV dynamically, which is nice to classify errors and allow different behaviours e.g. in testing deploys and actual production.
- Something I've always wanted was to have the exact version show up in the frontend version, that is, the git commit. With frontend env variables, that was possible, so now if you define `REVISION` it will show up in the UI as a hover:

image

Which corresponds to this commit:
https://github.com/beyondessential/tamanu/commit/e56334141316e159d34b59eb84d9722269a43ada

- This is also showing the frontend version; in production this can be different from the server version (which is what the non-hover thing is showing).
- Docker build yelled at me (with a warning) so I calmed it down (`as` -> `AS`)
- Editing the caddyfile kept trying to use spaces, so I fixed that (`.editorconfig`)
- Add the `%env` option to the PR auto-deploy mechanics so that `NODE_ENV` can be customised
- Enable Bugsnag by default in auto-deploys (Bugsnag was actually really useful in fixing the errors I found _in this very PR_!)

### Deploys

### Remember to...

- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly)
- ...update the [auto-deploy options reference](https://beyond-essential.slab.com/posts/tamanu-auto-deploys-kubernetes-kkovoquc)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,3 +16,7 @@ indent_size = 2
 
 [*.md]
 trim_trailing_whitespace = false
+
+[Caddyfile*]
+indent_size = 4
+indent_style = tab
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,9 +50,12 @@ COPY --from=build-server /app/node_modules/ node_modules/
 # set the working directory, which is where the entrypoint will run
 WORKDIR /app/packages/${PACKAGE_PATH}
 
-# explicitly reconfigure the port
-RUN echo '{"port":3000}' > config/local.json
+# explicitly configure the port
+ENV PORT=3000
 EXPOSE 3000
+
+# read configuration from source and from /config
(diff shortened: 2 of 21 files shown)


dannash100 (approved):
Looks great! Cool new horizons

PR #6530, opened 2024-10-02
fix(invoicing): EPI-975: fix wrong sourceId for potential imaging request area

namhoangbes:
### Changes

- Fix the issue where adding imaging request item from the potential item list cause the other imaging request with the same type to be vanished.

### Deploys


huytranbes (approved):
LGTM!

PR #6533, opened 2024-10-02
feat(sync): SAV-305: introduce (sync, Facility) permission, feature-gated

passcod:
### Changes

This is a first step to authenticating sync.

We introduce a feature flag: `auth.restrictUsersToSync`. If this is false or unset, all users can sync. Later, we can default it to true.

We introduce a permission: `verb:sync noun:Facility`.

Without an object, it grants the ability to sync any facility.
With an object, it grants the ability to sync a specific facility.

Sync is also subject to user-facility mappings when the `auth.restrictUsersToFacilities` feature is on.

Superusers are exempt.

I note that Users and Permissions are PULL_FROM_CENTRAL, so there's no risk of user privilege escalation via sync.

### Deploys

### Remember to...

- ...write or update tests
- ...update the [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q)


dannash100 (changes requested):
Looks great! Just a request to define the default and use settings reader for incoming db settings schema

edmofro (approved):
Glad @dannash100 reviewed this first because I wouldn't have caught those couple of specifics, but other than that lgtm!

dannash100 (approved):
🌞

PR #6552, opened 2024-10-08
tweak(settingsAdmin): Fix few issues with admin panel

dannash100:
### Changes

So the only annoying thing is that we have to make that controlled flag.
This basically means that it will only update internal state of selections when the external value prop is updated.
This means that it won't remove the inputs value when cleared until we actually set the value state to null etc.

We need something like this for the dismiss changes logic.

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
lgtm!

PR #6561, opened 2024-10-10
fix(enums): NASS-1410: Add missed patientDetails enums

tcodling:
### Changes

In the translated enum implementation we missed marital status which needs to be translatable. While working on this i discovered many missed enums in the patientDetails and birthDetails views so i cleaned all those up while here. This involved using the proper select field component and also moving some constants out of the web package to constants

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Yes love it!

PR #6589, opened 2024-10-19
db(source): KAM-291: document FHIR tables

passcod:
### Changes

Fill in documentation for all remaining TODOs in the fhir namespace.

This also includes a full description of the FHIR job queue system, which will surely be useful and was previously only in the original PR/commits.


dannash100 (approved):
Looks great - is that new CI step expected to be failing?

PR #6594, opened 2024-10-21
db(source): document logs schema

passcod:


dannash100 (approved):
Looks good!
Just want to confirm that it is intentional to have that central-server path inside `database/model/facility-server/logs/debug_logs.md` is it generically aplicable?

PR #6600, opened 2024-10-22
feat(db): EPI-1022: Create db change responsiveness via pub/sub

namhoangbes:
### Changes

- Implementing db notify

### Deploys


dannash100 (commented):
Will have a closer look when online thursday 🙏 Great cleanups and consolidations to what I added during upcomingVaccinations stuff 😆

dannash100 (changes requested):
@namhoangbes needs to be rebased or merged to remove diff from main if possible.

dannash100 (approved):
Looks great!
It will be a very cool feature going forwards

chris-bes (approved):

PR #6601, opened 2024-10-22
feat(booking): NASS-1371: Support adding a booking via a location cell

dannash100:
### Changes

Just needed a few tweaks to work.
- Had to specify a key due to the internal handling of locationId not responding to new initialValues inside location + area field **update** went with effect of form.initialValues field prop as will work when not in formik context and doesn't mess with key trick
- Had to make sure to not break if there was a appointment with no end time, created on appointment calender & Make sure to not fetch before pre-req params are ready inside the booking time field
- Invalidate query rather than track refresh count
- consolidate few queries 

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
LGTM

PS: unsure why the comments from last review reposted

PR #6603, opened 2024-10-22
fix: EPI-983: Dashboard is redirected after login even when it is hidden

huytranbes:
### Changes

Fix the issue that the app redirect to dashboard even when dashboard setting is set to 'hidden', because setting data is not cached before login for the first time

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (commented):
Hmmm this might work but I think we should try out some other stuff out first, lemme know what you think. I might've misunderstood the issue but I think the redirect should achieve this on its own

NavarroEmilioLuis (approved):
Thanks for double checking and sorry for the lateness🙏

PR #6604, opened 2024-10-23
fix: EPI-932: Record deleted task still appear in task set table

huytranbes:
### Changes

Fix task still appears in task set table when deleted_at has value in the reference_data_relation table

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Looks good, i thought paranoid: true default would handle this tho 🤔 - oh but mabye not, we do often make it explicit dont we

PR #6620, opened 2024-10-25
fix: release-2.18: Fix infinite loading survey responses printout modal

huytranbes:
### Changes

`isVillageQueryLoading` returned from useQuery() returns true even if villageId does not exist

### Deploys

### Remember to...

- ...write or update tests


rohan-bes (approved):
A comment that maybe we want to try using a different value from `useQuery` to determine whether we should display a loading spinner or not. But I don't think it's too much of a big deal so approving regardless of which option you go with :+1:

PR #6621, opened 2024-10-25
fix: release-2.18: EPI-1044: Fix calculated question still display if left empty

huytranbes:
### Changes

- Make calculated answer returns null if all the question fields are left empty. It somehow returned 0.0 when calculated from empty strings
- Add missing display handler for RESULT, CHECKBOX and CALCULATED question type

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/shared/__tests__/utils/calculations.test.js
+++ b/packages/shared/__tests__/utils/calculations.test.js
@@ -78,6 +78,45 @@ describe('Survey calculations', () => {
       expect(calculations.TEST).toEqual(1024);
     });
 
+    it('should supports 0 value variables', () => {
+      const survey = makeDummySurvey([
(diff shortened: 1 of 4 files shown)


rohan-bes (commented):
Nice stuff @huytranbes! I just had a question about the answer filtering logic with `0` values

rohan-bes (changes requested):
Suggestion for a way to more safely implement the `extractVariables` function.

Also requesting that we could add some unit tests for the new behaviour:
1. Ensure parser supports 0 value variables
2. Ensure parser returns `null` if all variables are undefined

rohan-bes (approved):
Looks great! Thanks for the work of implementing the suggestions :pray:

PR #6628, opened 2024-10-27
db(source): document invoices

passcod:


NavarroEmilioLuis (approved):
LGTM, there a bunch of 'Reference of the...' but I think we are not applying these (let me know and I'll mark them otherwise)

PR #6630, opened 2024-10-27
db(source): document facilities, locations, etc

passcod:


NavarroEmilioLuis (approved):
LGTM just one copy thing here as well

PR #6631, opened 2024-10-28
feat(fhir): SAV-820: Only rematerialise active Imaging/Lab requests

rohan-bes:
### Changes

Introduced the concept of whether or not a FHIR resource is `live`. If a resource is not live, it won't be rematerialised if upstream changes are made. We are introducing this to cut down on the number of records we rematerialise so that we don't force clients to re fetch large numbers of resources when common reference data is updated.

Initially we are just implementing this for ServiceRequests, where the rule is that a request is only `live` when it would be considered `active` in Tamanu.

### Deploys

### Remember to...

- ...write or update tests


passcod (approved):
lgtm, well-architected

PR #6633, opened 2024-10-28
db(source): document public.encounters

passcod:


NavarroEmilioLuis (approved):
Good. Style suggestion based on the appointments file style!

PR #6640, opened 2024-10-30
db(source): document all the admin/system tables

passcod:


dannash100 (changes requested):
Just some config changes i dont believe are relevant once shared is built

dannash100 (changes requested):
Two more clashes with main settings changes

dannash100 (approved):

PR #6684, opened 2024-11-11
fix(fhir): SAV-847: Filter out unresolved resources from FHIR search API (hotfix 2.16.12)

rohan-bes:
### Changes

This fix is being implemented to avoid a bug that occurs when a client fetches a resource from our FHIR API before it has been fully resolved. In this scenario, they end up stuck with a resource where the IDs in its references can't be used to fetch the referenced resource.

Some core changes here to the way we resolve references in FHIR:
- Added a `resolved` flag to every resource
- When a resource gets resolved by the resolver, we update it to have resolved=true
- When a resource rematerialises, we check if its references exist and if so we keep them rather than setting to the original unresolved value (this is to avoid resources frequently 'disappearing' from the API if they kept getting marked as unresolved)
- When a resource is resolved, we bump its last_updated so that it can be refetched

**UPDATE**
So after running into issues with bumping the `last_updated` via the database, and feeling a bit uncomfortable about managing the resolution logic in multiple places, I decided to go with changing the logic to only use the javascript resolution logic and to drop the database functions. This ultimately seems like it'd be simpler in the long run, but it is a core change. I've left in a bunch of the migrations for now but will drop them before merging (just to keep a record and for discussion during code review). 

### Deploys

### Remember to...

- ...write or update tests


chris-bes (changes requested):
Functionality wise they all make sense to me and I don't see any implications! I just have a few minor suggestions 🙏

chris-bes (approved):
LGTM!

PR #6707, opened 2024-11-12
feat(appointments): NASS-1386: Cancel appointment workflow

flynncs:
### Changes

I've extracted all common components of the cancel appointment/booking modal into a separate file, then created a new CancelAppointmentModal & CancelLocationBookingModal - reasoning being that although they currently are very similar, there are differences in pretty much every text, the hook for mutating, etc. I didn't want to pass in a bunch of props. Additionally, the appointment cancel modal will deviate in a later change for adding the repeating appointments part. 

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/Appointments/CancelModal/BaseModalComponents.jsx
+++ b/packages/web/app/components/Appointments/CancelModal/BaseModalComponents.jsx
@@ -57,12 +57,16 @@ const StyledConfirmCancelRow = styled(ConfirmCancelRow)`
   margin-top: 0;
 `;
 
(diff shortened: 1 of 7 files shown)


tcodling (approved):
Generally looking really good. Some naming that threw me off but i really cant find much wrong with this! Good call extracting out the base modal components.

I wonder if the files could be organised a bit better though? feels a bit strange having a folder for multiple kinds of cancel modal. I wonder if we should just have a "modals" folder that could include this and the warning modals? I think you could either do in this pr or we could do seperately off the epic 🙏

PR #6737, opened 2024-11-18
fix(api): find all appointments which overlap with queried range

jaskfla:
### Changes

Currently, when querying appointments by time (with `after` and `before`), these were only checked against an appointment’s start time.

This means that if there is a booking that spans from yesterday to tomorrow, if we request a list of today’s appointments, that three-day booking won’t show up.

This PR creates a `RANGE` from the `start_time` and `end_time` and looks for any appointments which overlap at any point with interval [`after`, `before`).

### Deploys

### Remember to...

- ...write or update tests


dannash100 (changes requested):
Just a suggestion to look into

dannash100 (approved):

PR #6747, opened 2024-11-19
fix(scripts): Update script to be able to work with NPM

NavarroEmilioLuis:
### Changes

So I needed to test migrating data up and down and this came up, so after checking it out it, this is my way of unblocking this

PS: you can still do the thing on your own but yeah this is faster


dannash100 (approved):
Looks gud!

PR #6750, opened 2024-11-19
fix(mobile): release-v2.20: fix sync errors on mobile

khoi-bes:
### Changes

- There are two reason for errors on syncing process on mobile:
1. Import `taskTemplate` will create two `translated_string` records (cuz we need to add one record to `reference_data` table and the other one to `task_templates` table). The first `translated_string` record (from `reference_data` instance) is valid but the second one (from `task_template` instance) is not cuz we can't extract name from this instance. This case will happen for `all referenceData type` which need to extend its data like `taskTemplate`, so just leave the `translated_string` records duplicated and set the text to "" to avoid syncing error on mobile.

image

2. TableIndex for referenceDataRelation hasn't been updated on mobile

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Perfect! Oh this explains the null text error that occurred as well

PR #6754, opened 2024-11-20
ci: KAM-302: toolbox

passcod:
### Changes

- Include the alerts folder in the servers images and the packages
- Package the alerts separately as well (the current packaging was actually broken, this copies the cd-package-frontend which is a known-working config)
- Use `metadata-action` to generate short version tags in the conventional style, so `v2.22.0`, `v2.22`, `v2`
- Adds a toolbox container image with tooling, dbt, bestool, the alerts, the database models.
  - This can be used for debugging in a controlled environment wherever we have access to docker.

For example:

```console
kubectl \
	-n tamanu-some-autodeploy-namespace \
	debug deployment/facility-1-api \
	--image ghcr.io/beyondessential/tamanu-toolbox:v2
```

This will create an ephemeral container in the context of an auto-deploy, with access to the bestool, alerts, dbt, etc. There's still some work upfront to connect to databases etc but it should be a useful tool in many cases.

Additionally, building an arbitrary container image with bestool included is now as easy as:

```dockerfile
FROM ubuntu
COPY --from=ghcr.io/beyondessential/tamanu-toolbox:v2 /usr/bin/bestool /usr/bin/bestool
```

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
Well it's green so it must work. I'm not too familiar with these but nothing jumped out of me either. Left two comment/questions but non-blocking.

PR #6760, opened 2024-11-21
feat(charting): SAV-698: Conciliate charting stuff

NavarroEmilioLuis:
### Changes

So most of the workflows were only partially working, navigation screwed things up and these issues came up. I'm just wary about solving first before doing other stuff.

Things fixed:
- When creating a new chart record entry, the mutation didn't actually set the expected format
- The user preferences table was still using user ID as primary key, which makes it so that we can't remember more than one category at once
- Selecting different chart types didn't update the table

### DIFF
--- a/database/model/central-server/public/user_preferences.yml
+++ b/database/model/central-server/public/user_preferences.yml
@@ -10,7 +10,7 @@ sources:
           - system
         columns:
           - name: id
-            data_type: character varying(255)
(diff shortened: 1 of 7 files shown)


chris-bes (changes requested):
LGTM! Just 1 question that I've got

chris-bes (approved):

PR #6795, opened 2024-11-25
tweak(locationBooking): Ensure that location field state is cleared between multiple create booking 

dannash100:
### Changes

Pretty annoying thing with state managed inside location group & location field.
Even tho we changed it to reinit as initalValues changed - it would still cause stuck state when initalValues are {} - in the case of multiple create bookings.

await is required - to preserve animation and also to ensure there is a render between the null and the {}

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
clean i actually really like this

PR #6796, opened 2024-11-25
fix(syncLookup): TAMOC-251: Save facility ID for appointments

NavarroEmilioLuis:
### Changes

We missed a spot to be able to discriminate which record should be pulled!


chris-bes (approved):
LGTM!

PR #6808, opened 2024-11-26
fix(translations): TAMOC-250: Improve translated suggestions query

NavarroEmilioLuis:
### Changes

This was made by Edwin and I just executed and tested😆


edmofro (approved):
Approved as MVP for hot fix! As discussed, we should really be ordering by where in the string the search term matches for translations, like we do for non-translated reference data. Then, we should be interleaving the two search results so all starts-with results are up top. However, this will fix the performance issues and is no worse than what is in the wild

PR #6860, opened 2024-12-02
feat(tasking): Implement notification hooks for ImagingRequest and LabRequest updates

khoi-bes:
### Changes

- Move pushing notification logic to afterUpdate hooks

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/shared/src/models/Notification.js
+++ b/packages/shared/src/models/Notification.js
@@ -89,16 +89,17 @@ export class Notification extends Model {
         case NOTIFICATION_TYPES.IMAGING_REQUEST: {
           userId = metadata.requestedById;
           const encounter = await models.Encounter.findByPk(metadata.encounterId);
-          patientId = encounter.patientId;
+          patientId = encounter?.patientId;
           break;
         }
         case NOTIFICATION_TYPES.LAB_REQUEST: {
           userId = metadata.requestedById;
           const encounter = await models.Encounter.findByPk(metadata.encounterId);
-          patientId = encounter.patientId;
+          patientId = encounter?.patientId;
           break;
         }
-        default: return;
+        default:
+          return;
       }
 
       await this.create({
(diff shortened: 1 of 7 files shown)


namhoangbes (approved):

packages/shared/src/models/Notification.js
```diff
@@ -89,16 +89,17 @@ export class Notification extends Model {
...
           const encounter = await models.Encounter.findByPk(metadata.encounterId);
-          patientId = encounter.patientId;
+          patientId = encounter?.patientId;
           break;
         }
         case NOTIFICATION_TYPES.LAB_REQUEST: {
           userId = metadata.requestedById;
           const encounter = await models.Encounter.findByPk(metadata.encounterId);
-          patientId = encounter.patientId;
+          patientId = encounter?.patientId;
```
edmofro, line 98: Why is the `?` necessary? There should never be a time where an imaging request or lab request _doesn't_ have an encounter, right?

edmofro (commented):
I have another question on this one too

packages/shared/src/models/Notification.js
khoi-bes: yes your are right, but in theory, the `patientId` can be null in the encounters table. As I remember, it was failed in some test cases so I have to use the `?`

edmofro: Oh right it's nullable in the DB! I see that now. It's bad that we have that as nullable, have added a maintenance sprint item [here](https://beyond-essential.slab.com/posts/tamanu-maintenance-sprint-items-j1ac0rkb#hapt2-new) to fix that.

However, that should never be the case in reality though - all encounters should have a `patient_id`, and I've never seen one without. If test cases were failing that means the tests need fixing, not the code. We should _never_ add defensive code unless there is a real use case where the data should be null/undefined. This isn't one of those cases.

Can you please update it to remove the defensive programming and fix any tests that break as a result?

khoi-bes: alright, let me fix

edmofro: We shouldn't use a js logical change to fix a type error

PR #6927, opened 2024-12-10
repo(pm2): Do not start fhir processes when fhir worker is disabled

passcod:
### Changes

This came up recently in on-call. If a server has the FHIR worker disabled, then by default we'll start two useless processes, which will eat some memory and do nothing. This change reads the config from the pm2 manifest to only start FHIR worker processes if they'll be used.


NavarroEmilioLuis (approved):
LGTM

PR #6933, opened 2024-12-11
feat(desktopCharting): SAV-698: Add charting graphs

NavarroEmilioLuis:
### Changes

At first glance some choices might not seem obvious but this is the most 'proper' and smaller approach! We might want to do an extra refactor to change the naming but that's secondary


rohan-bes (approved):
Looks good! Just a few minor comments

PR #6937, opened 2024-12-11
fix(test): failing front-end unit test

jaskfla:
### Changes

#6844 introduced a small change to the `<DownloadDataButton>` that wasn’t reflected in its unit test. This:

- updates the test to expect the new (translated) label; and
- adds a new (very similar) unit test for the new branch of code

### Deploys


dannash100 (approved):
# 100

PR #6953, opened 2024-12-15
fix(scheduling): NASS-1458: Update user_preferences table to allow per facility preferences

tcodling:
### Changes

We discovered in implementing a couple of new `user_preferences` that the current table schema is not ideal for situations where we need to filter settings by facility. To make it work with current implementation you needed to have each "per facility" setting nested in a facility keyed object. Not only does this make the database structure confusing to follow, it also requires lots of fetching/resubmitting data or custom endpoint logic in order to work around this. After  discussion with team we decided to add a `facility_id` column to the table to allow easy fetching of preferences by facility when needed. 

This works very similar to the settings table where if a user fetches their user preferences, they will recieve a merged object of their general settings overwritten with any facility specific settings defined for the user.

### Deploys

### Remember to...

- ...write or update tests


dannash100 (commented):
Some lil things I spotted, such amazing work on this tho - such a complex task

dannash100 (approved):
Great work phew!

PR #6955, opened 2024-12-15
feat(desktopCharting): SAV-700: Update charts importer

NavarroEmilioLuis:
### Changes

Includes more logic to avoid shooting ourselves in the foot. This is covering that all charting questions have a constant ID (we need these to be able to do grouping logic)

Note to self, TODO: Update any imported charts on the epic directly on the database before merging!


rohan-bes (approved):
Couple of minor comments but happy to approve all the same :+1:

PR #6968, opened 2024-12-17
feat(desktopCharting): SAV-698: Disable graphs for complex charts

NavarroEmilioLuis:
### Changes

Just a small thing I missed. There is work planned ahead to support complex charts, so instead of blocking it everywhere, I decided to just return an empty state for visualisation configs, allowing us to remove it in the future pretty easily


dannash100 (approved):
I trust ya

PR #6972, opened 2024-12-19
fix(desktopCharting): SAV-821: Remove menu button on conditions and cleanup

NavarroEmilioLuis:
### Changes

Addressing QA, the menu should only appear when an instance is selected and that instance does not have any responses.

Also changed something that I'm sure it's going to bite us


dannash100 (approved):
Perfect!

PR #6973, opened 2024-12-19
feat(outpatientAppointments): NASS-1383: Repeating appointment models

dannash100:
### Changes
Covers generateRepeatingAppointment on model and validation of schedule

Still todo in follow up pr
- Considered rework maximum initial appointments config into relative date in future to match tasks,
But more I thought about it mabye not as appropriate for appointments - given you could have an 8 weekly and only have to create 1 appointment say. Keen for opinions... Mabye its an important part of the scheduled task. Ill have a think

- Make a scheduled task like repeated tasks to create future appointments that go beyond initial maximum when they arrive

### Deploys

### Remember to...

- ...write or update tests


jaskfla (approved):
Just tiny deets, but looking solid! 🏆

PR #6980, opened 2024-12-19
tweak(locationBookings): avoid flash of content when loading time slots

jaskfla:
### Changes

Current implementation always falls back to skeleton time slot toggles (with the default pulse animation from MUI). When switching dates in the Location Bookings view, this only appears briefly and makes the GUI look janky.

This PR introduces a smarter placeholder which uses an arbitrary day to render placeholder disabled time slot toggles. Under the hood it uses today as that arbitrary day, but the time times of day that make each time slot are guaranteed to be identical.

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
nice!

PR #7005, opened 2025-01-05
refactor(locationBookings): clean up `TimeSlotPicker` change logic

jaskfla:
### Changes

Updates `<TimeSlotPicker>` to improve readability by treating `startTime` and `endTime` as the one source of truth. Now uses `useEffect` to synchronise these values (if they exist) to the visible status of the rendered component, rather than updating the underlying data and the component state in parallel

This results in (at least) one more render cycle each time a start/end time is updated, and—as with so many `useEffect` hooks—can be prone to infinite render cycles. Despite these, I find this to be more readable/maintainable—particularly when looking down the barrel of NASS 1522 where we will attempt to preserve the start time when switching a booking overnight and non-overnight

(I previously attempted NASS 1522 using only `onChange` handlers, but it spread the logic out into too many different places, and ran into problems with infinite render loops anyway 🤷)

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/Appointments/LocationBookingForm/DateTimeRangeField/TimeSlotPicker.jsx
+++ b/packages/web/app/components/Appointments/LocationBookingForm/DateTimeRangeField/TimeSlotPicker.jsx
@@ -68,9 +68,9 @@ export const TimeSlotPicker = ({
   date,
   disabled = false,
   /**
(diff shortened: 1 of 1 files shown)


passcod (approved):
Approving for the current code, which lgtm, but pending optional resolution of the legal selection prop.

PR #7026, opened 2025-01-08
deps: add missing direct dependencies to all packages

passcod:
### Changes

Missing dependencies found with (fish shell, using [ripgrep] and [jq]):

```fish
for pkg in (rg "from '" | rg -vF '://' | cut -d\' -f2 | rg -v '([.]/|@tamanu|node|^/|^~|[.]{2}|^types$)' | cut -d/ -f-2 | sort -u)
	if not jq "(.dependencies + .devDependencies) | keys[] | select(. == \"$pkg\")" package.json -e >/dev/null
		echo $pkg
	end
end
```

(with some false-positives filtered by human)

---

Also, all the nodejs imports were switched to the `node:` syntax, which disambiguates between external deps and Nodejs internals. That was done with [sd] and [ripgrep]:

```fish
sd "from '(assert|child_process|crypto|fs|http|os|path|repl|url|util|zlib)'" "from 'node:\$1'" (rg -l "from '(assert|child_process|crypto|fs|http|os|path|repl|url|util|zlib)'")
```

---

Finally, all direct usage of `node-fetch` was removed since that's native in both browsers and Node.js now.

---

Also, because I did the additions of packages using `npm i`, it re-ordered the dependencies fields alphabetically.

[ripgrep]: https://github.com/BurntSushi/ripgrep
[jq]: https://jqlang.github.io/jq/manual/
[sd]: https://lib.rs/crates/sd


edmofro:
@passcod last one

passcod:
Needs redoing from scratch unfortunately

PR #7027, opened 2025-01-08
refactor(adminPanel): NASS-1563: Move search input out of form

edmofro:
### Changes

We currently make use of a SearchField, which works within the context of a formik form. Because it therefore needs to be part of the form, we have to do all sorts of weird things to deal with that. We should have a standalone SearchInput so we can do things in a normal way.

### Deploys

### Remember to...

- ...write or update tests


tcodling (changes requested):
Mainly just my big comment about trying to follow the existing field pattern if possible and drying up the code. Totally open to pushback though if you did it like this for a particular reason!

tcodling (approved):
lgtm!

PR #7034, opened 2025-01-10
fix(encounterView): Ensure view doesn't break while survey data is loading

NavarroEmilioLuis:
### Changes

Pretty small change that not always happens. This hasn't been reported yet, but I got it while doing other work


tcodling (approved):
Ah yep i noticed this on my local and it was throwing me off 😆 Thanks!

PR #7053, opened 2025-01-13
feat(appointments): NASS-1423 Cancel repeating appointments

tcodling:
Logic for cancelling repeating appointments.

The basic flow is:
- When opening cancelling modal for appointment, if the appointment has a schedule linked, we display the special repeating appointments modal.
- You will have the option to delete either just this appointment or this and all future appointments
	- Selecting this appointment will cancel as normal
	- Selecting this and all future appointments will set them all to cancelled and end the linked appointment schedule using model method

Daniel and I have discussed possibly separating this out into a delete endpoint if it gets to complicated with the modify logic but we have been a bit on the fence. We may end up doing this later but have decided this is fine for now

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/models/Appointment.ts
+++ b/packages/database/src/models/Appointment.ts
@@ -66,6 +66,7 @@ export class Appointment extends Model {
       'appointmentType',
       'bookingType',
       'encounter',
+      'schedule',
     ];
   }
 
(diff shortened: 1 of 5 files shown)


dannash100 (changes requested):
A few things :)

dannash100 (approved):
Approved - Im gonna pull this into my branch and do a bunch of tweaks

PR #7055, opened 2025-01-13
feat(sync): SAV-807: Handle duplicated display IDs

chris-bes:
### Changes
- Handle duplicated display IDs

### Deploys

### Remember to...

- ...write or update tests


edmofro (changes requested):
This came together super super nicely, thank you @chris-bes!!!

I have some high level thoughts. This problem of having sync side effects has come up in a few places over time. I really like where you've put that alignment function, but I think we should flag it clearly as something that could be used as a generic hook. So something like `prePersistSyncHook`, or `incomingSyncHook`, or `runIncomingSyncSideEffects`. E.g. @dannash100 recently was looking at a way to delete extra appointments created on the central server based on a schedule that was now outdated on the facility server. An `incomingSyncHook` would have been the perfect place, and would have worked really nicely to delete those extra appointments - but in that case it wouldn't touch the actual incoming data, so isn't really about alignment.

My second thought is that we should not allow that hook to do any db manipulation itself, other than to the sync snapshot table. In this case what you've done is very likely fine, but I think that we should retain the idea that all changes for one sync session flow through one mechanism, in case we rely on that for other things like stats or something. Because of that, I think the `incomingSyncHook` should be wrapped in a read only transaction, but be able to return some sort of data structure containing changes to records, which are then smartly added or updated in the snapshot table. Or a transaction that is read only for all but the snapshot table, if that's possible? I might be overthinking this one.

What do you think? If you feel you need to move on from this, I am happy for you to just make a small tweak to change the name of the alignment function and leave the rest.

edmofro (changes requested):
Yes loving this after the updates, thank you!

edmofro (approved):

edmofro (changes requested):
Just went through to catch any lingering references to alignment

edmofro (approved):

anilvkaran (approved):

PR #7060, opened 2025-01-14
fix(patientView): release fix v2.23: Patient view reload bug

tcodling:
Simple fix for stopping the full screen error when patient screen has no patient in context. 

This is a weird fix because it returns to the 2.22 behaviour but also exposed an exisiting bug that the patient refresh is broken. Previously you were able to refresh on the patient without issue but now it logs you out and when you log back in the patient screen will be loaded but without a patient in context.

I had a quick look into reimplementing the proper behaviour but it wasn't obvious and this is a release fix so just going to make a card for now.

### Deploys

### Remember to...

- ...write or update tests


edmofro (approved):
Yeah there's a deeper fix here in that
- Reload should keep the patient in context, but also
- We shouldn't be on the patient screen if there's no patient in context

But as you say, release fix so 👍

PR #7074, opened 2025-01-16
fix(deaths): Release fix v2.23: Patient death button display

tcodling:
After my last tweak it stopped showing the "Record death button". This was because after adding in the "enabled" option for react query, `isLoading` is always set to true when enabled option for query is false. I have just replaced `isLoading` with `isFetching`. From looking online this is a common pain point for people and this is often how it is solved

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
100

PR #7085, opened 2025-01-21
fix(desktopCharting): SAV-686: QA stuff 1601

NavarroEmilioLuis:
### Changes

Although it looks like lots of changes, there aren't really that many! Most come from moving tests to their own file plus adding one test to that file.

Refactored a couple of logics that were wrong. Here we make use of the table loading state and avoid flickering until we fetched everything. Also updated the initial chart selection and initial instance selection plus solving a couple of display bugs.


chris-bes (changes requested):
Looks great overall! Just a couple of suggestions

chris-bes (approved):
LGTM!

PR #7086, opened 2025-01-21
feat(adminPanel): NASS-1461: Allow importing only new default translations (backend)

edmofro:
### Changes

Each commit is self-contained :-) 
One small front end tweak as I changed the direction of the query parameter.

Note that this actually allows `skipExisting: true` to be applied to any reference data import, but only the translations front end is using it.

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
LGTM

PR #7091, opened 2025-01-22
fix(desktopCharting): SAV-821: Improve and correct display

NavarroEmilioLuis:
### Changes

- The delete menu should be shown when there are no entries recorded against the instance, the logic was wrong before 😓
- Saw some flickering when submitting/navigating so thought I'd get ahead of it


rohan-bes (commented):
Looks good! Had a bit of a question about the `useCallback` but I mightn't be following exactly what it's doing so if it's working all good :+1:

rohan-bes (approved):
Omg that pesky co-pilot! Bad robot :robot: :hammer: :stuck_out_tongue_winking_eye:

PR #7114, opened 2025-01-27
tweak(surveys): SAV-890: Filter clinical statuses by program registry…

NavarroEmilioLuis:
### Changes

Cherry picking changes from #7104


tcodling (approved):
Thanks so much for this Emilio!

PR #7117, opened 2025-01-27
feat(repeatingAppointments): NASS-1422: Enable modify repeating appointments

dannash100:
### Changes
Ive gone for not keeping these form modifiers like isRepeatingAppointment and endsMode and stuff outside of formik state for most part and just determined from the form state that informs it.

I think this is better - it means we don't have to do alot of whacky stuff when initialising the edit form no more. See my last version of this where I init form with
```
    setSelectedAppointment({
      ...selectedAppointment,
      modifyRepeatingMode,
      isRepeatingAppointment: !!schedule,
      schedule: {
        ...schedule,
        endsMode: schedule.untilDate ? ENDS_MODES.ON : ENDS_MODES.AFTER,
      },
 ```
 Which is like why pass this info in agian? But doing it within state and not muddying form values is messy too in its own way
 
 Also its gonna do the diff on the schedule on the endpoint to avoid more front end fluff ye
 
 
 I think ultimately i like this approach better tho and sseems to be working good

### Deploys

### Remember to...

- ...write or update tests


edmofro (commented):
Damn son this is soooo much better, I love it

edmofro (approved):
lgtm!

this one feels like a fair bit of dev testing could be useful

PR #7119, opened 2025-01-28
fix(mobile): 2.24 fix: Error when writing character `(` in select filter

dannash100:
This lead to an unterminated capture group.
- This version removes the templated regex to do case insensitve comparison - it does the same thing as far as I could visualize and test but really keen for second option
- Alternative solution to this is to sanitise the search input before templating into regex
`text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');` Mabye lower touch? I just thought it looks bit ick

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
I think this is better than the regex for sure. Also good improvement taking `parts` outside the iteration

PR #7121, opened 2025-01-28
fix(registerPatient): 2.24 fix: Mobile newly registered patients not showing up when searching by dob

dannash100:
### Changes

[See Thread](https://beyondessential.slack.com/archives/C03KJSEPV9A/p1738029544261489)
Low touch solution to this one. 

It was saving them as date times in local sqlite3 db
They where trimmed to size by time they hit central tho so doesn't effect sync or anything

Heres the state after creating 4 patients on mobile
Screenshot 2025-01-28 at 4 09 56 PM

They then do not make a complete match with dob filter.

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (commented):
Hmmmm unsure! The change looks good but I'm not sure if we should be looking into a more generic solution? I'm curious whether the date and datetime string helpers are doing what we think they are

NavarroEmilioLuis (approved):
Yeah cool let's go with this for the hotfix!

PR #7122, opened 2025-01-28
fix(alerts): sync alerts need to use facilityIds now

passcod:
### Changes

- Sync: When we switched to putting `facilityIds` in `debug_info` in the `sync_sessions` table it broke the sync alerts. Unfortunately it did so silently such that the sync alerts always return zero rows == no errors.

- FHIR: The job `fhir.resolver` just tends to sit there all the time even if FHIR is disabled, and that's causing a spurious alert.

### DIFF
--- a/alerts/fhir-queued-job-long.yml
+++ b/alerts/fhir-queued-job-long.yml
@@ -1,11 +1,13 @@
 sql: |
   SELECT
     id,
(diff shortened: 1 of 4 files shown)


NavarroEmilioLuis (approved):
LGTM!

PR #7123, opened 2025-01-28
fix(fhir): don’t materialise anything if the worker is disabled

passcod:
### Changes

Not certain this is a good idea or if this should be done a different way, but currently when `fhir.worker.enabled` is `false`, then materialisation jobs are still inserted into the queue, which means the job queue grows infinitely (which triggers alerts, and caused Samoa going down).


NavarroEmilioLuis (commented):
Well it kinda makes sense. The downside would be that when we turn it on, we will start having a big queue but I mean, that's expected.

I think I'd be good to have yet a third pair of eyes to confirm before approving? Probably Rohan's?

PR #7124, opened 2025-01-29
tweak(alerts): convert to markdown

passcod:
### Changes

bestool 0.26.0 adds markdown support for the alerts, which means that slack alerts no longer need to have all the ugly HTML, but emails will still look beautiful. This PR converts all the standard alerts to markdown.

It also does a few tweaks to the text and queries:
- adds `ORDER BY` clauses in places to stabilise the message and highlight e.g. the longest running FHIR jobs first
- removes redundant info from the template (e.g. when that's present in the subject)
- removes the implied "Please see the logs"
- removes the unnecessary `;` from the end of queries

### DIFF
--- a/alerts/certificate-notification-error.yml
+++ b/alerts/certificate-notification-error.yml
@@ -2,22 +2,21 @@ sql: |
   SELECT *
   FROM certificate_notifications
   WHERE status = 'Error'
-  AND created_at > $1;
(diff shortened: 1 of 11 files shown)


dannash100 (approved):
Awesome I love md.
Is it all good that we losing some `;` on the queries? I assume they are optional

PR #7125, opened 2025-01-29
tweak(scheduling): NASS-1570: Add clinician filter on clinician view

tcodling:
### Changes

This adds a few things to update the filters for Outpatient appointments
- Store the groupBy in userPreferences
- Add a new clinician filter that is used when groupBy is by Area

If I had more time I would restructure this logic quite a bit as it gets incredibly confusing/tricky to code since we essentially have 3 objects we need to keep in sync (formik values = context state = userPreferences)

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Looks good! wish we could get rid of that useEffect tho waa

PR #7127, opened 2025-01-30
fix(scheduling): Release fix 2.24: Add ability to remove saved endTime

tcodling:
### Changes

Because end time isnt sent with the body when its been deleted, the endpoint doesn't attempt to update/delete it. Here I manually check and then set the column to null. I do feel like there must be an easier way to do this though 🤔 

### Deploys

### Remember to...

- ...write or update tests


dannash100 (approved):
Yes perfect

PR #7128, opened 2025-01-30
fix(fhir): SAV-891: Bump lastUpdated of MediciReport when rematerialising

rohan-bes:
### Changes

We were actually never bumping the last updated previously.

### Deploys

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
So what I'm thinking is that: did we ever had a purpose for not including it?

Updating it makes perfect sense as a FHIR resource but because this is not standard, I'm just a bit curious as if we had a requirement of only sending them in once, or if we just assumede that discharges mean no more updates?

I'm happy to just stamp either way as you might've already given thought to this

edmofro (approved):
lgtm! And Félix will be happy to see you remove that hard coded age 😄

PR #7131, opened 2025-01-30
fix(patientRegistration): 2.24 fix: Allow a table column with title of `''` to hide title and not fallback to key

dannash100:
### Changes

Previously this was falsy so we always had to set key to "" which is problematic in the case that two titles are hidden (key duplicate error) This fixes it

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
nice. saw a couple the other day where they made both the key and title '' but this seems betta

PR #7145, opened 2025-02-02
tweak(mobile): Mobile small cleanup 270125

NavarroEmilioLuis:
### Changes

Well this was meant to be a different thing. I was working in some other place and wasn't feeling all the warnings. Plus, I ended up resolving some other lint issues from GitHub.

### Deploys

### DIFF
--- a/packages/mobile/App/ui/components/AutocompleteModal/AutocompleteModalField.tsx
+++ b/packages/mobile/App/ui/components/AutocompleteModal/AutocompleteModalField.tsx
@@ -50,7 +50,6 @@ export const AutocompleteModalField = ({
 }: AutocompleteModalFieldProps): ReactElement => {
   const navigation = useNavigation();
   const [label, setLabel] = useState(null);
-  const refDataType = getReferenceDataTypeFromSuggester(suggester);
 
   const onPress = (selectedItem): void => {
     onChange(selectedItem.value);
@@ -68,6 +67,7 @@ export const AutocompleteModalField = ({
     (async (): Promise<void> => {
       const data = await suggester.fetchCurrentOption(value);
       if (data) {
+        const refDataType = getReferenceDataTypeFromSuggester(suggester);
         setLabel(
           <TranslatedText
             stringId={getReferenceDataStringId(data.value, refDataType)}
@@ -78,7 +78,7 @@ export const AutocompleteModalField = ({
         setLabel(null);
       }
     })();
-  }, [value, refDataType, suggester]);
+  }, [value, suggester]);
 
   if (readOnly) {
     return <ReadOnlyField value={label} />;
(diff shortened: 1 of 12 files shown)


packages/mobile/App/ui/components/AutocompleteModal/AutocompleteModalField.tsx
```diff
@@ -78,7 +78,7 @@ export const AutocompleteModalField = ({
         setLabel(null);
       }
     })();
-  }, [value]);
+  }, [value, refDataType, suggester]);
```
dannash100, line 81: Eek some of these ones scare me they have caused issues in past when applying lint rules to a codebase. I think best to check these ones if possible.

dannash100 (approved):
Looks good - just warning that I have seen bugs like infinite rendering and stuff when blanket fixing dep rules so just worth double checking where possible.

packages/mobile/App/ui/components/AutocompleteModal/AutocompleteModalField.tsx
edmofro: That's interesting to hear @dannash100, I always blindly apply. What kind of issues have you encountered?

dannash100: Infinate render loop issues - this was on a large code base that hadn't applied the rules in past tho

dannash100: and a couple unexpected state issues but I think codebase was full of side-effects and general loosness.

edmofro: Yeah interesting, good to be aware of!

NavarroEmilioLuis: Yeah, Daniel and I spoke about this on DMs. I said that I will confirm with a round of test to ensure this is not happening. But I don't think it should - usually those problems occur whenever you set state within a use effect and the state value is part of the deps. That's not the case here - I don't think the suggester instance will be different, the other value is just a string and I hope that react is smart enough.

But even then, that other value isn't used anywhere in the component and uses an imported function, so I'll just move it within the use effect.

NavarroEmilioLuis: Okay, so I checked and everything is fine with this update. Any objections to merge it directly without a tester? @dannash100 @edmofro

edmofro: If you've tested it, yep all good to merge imo

edmofro (approved):

PR #7157, opened 2025-02-04
fix(outpatientAppointments): 2.24 fix: Update endTime date on startTime change

dannash100:
### Changes
This already happened on end_time change - as the field is a time only field we had to ensure that the calendar date was set from the start_time.
However when changing start_time. the end_date would remain the original calendar date of start_time and not update.

I noticed this when fixing some issues identified in modify repeating ticket but it also applies to 2.24

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
Seems a bit verbose so if you could simplify through `date-fns` functions would be ideal but otherwise looks legit

PR #7166, opened 2025-02-04
feat(labs): SAV-893: Add ability to flag a lab request as sensitive

NavarroEmilioLuis:
### Changes

Title sums it up. I think I'll create a helper to reduce the noise on the tests.

### Deploys


packages/facility-server/app/routes/apiv1/labs.js
```diff
@@ -35,6 +36,10 @@ labRequest.get(
   '/:id',
   asyncHandler(async (req, res) => {
     const labRequestRecord = await findRouteObject(req, 'LabRequest');
+    const hasSensitiveTests = labRequestRecord.tests.some(test => test.labTestType.isSensitive);
```
chris-bes, line 39: This is repeated quite a lot in this file, I reckon let's move all of them to be inside the model LabRequest:
```
    hasSensitiveTests() {
return this.labRequestRecord.tests.some(test => test.labTestType.isSensitive);
}
```

Then we just need to check:
`if (labRequestRecord.hasSensitiveTests())` instead of having to assign it to another variable

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -586,9 +583,34 @@ createSuggester(
...
+            FROM lab_requests
+            INNER JOIN encounters
+              ON (encounters.id = lab_requests.encounter_id)
+            INNER JOIN lab_tests
+              ON (lab_requests.id = lab_tests.lab_request_id)
+            INNER JOIN lab_test_types
+              ON (lab_test_types.id = lab_tests.lab_test_type_id)
+            WHERE lab_test_types.is_sensitive IS TRUE
+              AND encounters.patient_id = :patient_id
+          )`,
```
chris-bes, line 604: This query looks just like the main query in `idBaseFilter`, except with the extra joins and check for `lab_test_types.is_sensitive IS TRUE`? I reckon we can just use this query in the main `idBaseFilter`, and conditionally check for lab_test_types.is_sensitive IS TRUE based on the `SensitiveLabRequest` permission? Just a bit wary that having 2 sub queries might have bad impact on the performance.

chris-bes (changes requested):
LGTM! Just leaving a few suggestions

packages/facility-server/app/routeHandlers/labs/getLabRequestList.js
```diff
@@ -26,6 +26,12 @@ export const getLabRequestList = (foreignKey = '', options = {}) =>
       newOptions,
     );
 
+    const canListSensitive = req.ability.can('list', 'SensitiveLabRequest');
+    const filteredLabRequests = labRequests.filter(labRequest => {
+      if (canListSensitive) return true;
+      return labRequest.tests.every(test => test.labTestType.isSensitive === false);
+    });
```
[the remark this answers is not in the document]
NavarroEmilioLuis: Yes but no because we might need to include the notes. So it seemed a bit awkward to return as is and apply the filtering in two different places

packages/facility-server/app/routes/apiv1/labs.js
NavarroEmilioLuis: I'm pushing back on this one, it's only one more line, and `this` doesn't necessarily includes both `tests` and nested `labTestType`

packages/facility-server/app/routes/apiv1/suggestions.js
NavarroEmilioLuis: Hmmm I need to double check the behavior. So you're right, it is very similar, except this one is a `NOT IN`, unfortunately flipping it doesn't provide the same results:

If we say get me `lab_test_types.is_sensitive IS FALSE`, then, we might have one sensitive and one non-sensitive in the same category (we shouldn't, but it could happen). In that scenario, we would still suggest the category, even though it has sensitive tests under it. Whereas this approach removes the category with just one sensitive test under it.

It seems to me that we might want the latter, and we can drop needing two subqueries - or even are okay suggesting the category anyway and can undo the endpoint!

packages/facility-server/app/routeHandlers/labs/getLabRequestList.js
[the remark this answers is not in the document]
chris-bes: Hmmm yeah this is not that significant, I'll leave it up to you but if you want I reckon you can do something like below which is a bit clearer to me when reading:

```
 const permittedLabRequests = canListSensitive  ? labRequests : labRequests.filter(labRequest => {
      if (canListSensitive) return true;
      return labRequest.tests.every(test => test.labTestType.isSensitive === false);
    });
```

packages/facility-server/app/routes/apiv1/suggestions.js
chris-bes: Hmm I'm a bit confused, maybe let's chat through slack

NavarroEmilioLuis: Resolution was: proper configuration should not permit a category with both sensitive and non-sensitive, so we can use the shorter query version and avoid the double subquery!

I still kept the two separate objects, even if they repeat themselves, as it's a bit cluttered having to add strings in more than one place

chris-bes (approved):
LGTM

PR #7169, opened 2025-02-06
fix(suggester): NASS-1535: Fix suggester vunerabilities

tcodling:
### Changes

Remove the string interpolation directly into sql from queries. The only thing I was a bit on the fence about is we have a couple of places where we directly interpolate constants into SQL as values but I have left them for now.

There is one place above these code changes where `searchQuery` is interpolated however it goes in to a findAll where clause so it should be automatically escaped from what I have read online. I tried to be extra safe and run it through Sequilize.escape or use replacements but I couldnt get either strategy to work

### Deploys

### Remember to...

- ...write or update tests


packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -343,7 +343,7 @@ createSuggester(
       if (!types?.length) return;
 
       const caseStatement = query.types
-        .map((value, index) => `WHEN '${value}' THEN ${index + 1}`)
+        .map((value, index) => `WHEN '${Sequelize.escape(value)}' THEN ${index + 1}`)
```
tcodling, line 346: Seemed like the best way since its not easy to add to the replacements object

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -703,31 +703,40 @@ createNameSuggester(
...
+      id: {
+        [Op.notIn]: Sequelize.literal(
+          `(
           SELECT DISTINCT(pr.id)
           FROM program_registries pr
           INNER JOIN patient_program_registrations ppr
           ON ppr.program_registry_id = pr.id
           WHERE
-            ppr.patient_id = '${query.patientId}'
+            ppr.patient_id = :patient_id
```
tcodling, line 726: Someone went through and caught most of the places like this in a previous pr. Here I just caught the last one

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -78,7 +78,7 @@ function createSuggesterRoute(
         : [];
       const suggestedIds = translations.map(extractDataId);
 
-      const whereQuery = whereBuilder(`%${searchQuery}%`, query);
+      const whereQuery = whereBuilder(`%${Sequelize.escape(searchQuery)}%`, query);
```
tcodling, line 81: Not sure if I am getting a bit loose with the Sequelize.escape but this seems like the safest/clearest way to go about it to work with any `whereBuilder` created

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -342,18 +343,26 @@ createSuggester(
...
+      return {
+        sql: [
+          Sequelize.literal(`
+            CASE "ReferenceData"."type"
+              ${caseStatement}
+              ELSE ${types.length + 1}
+            END
+          `),
+        ],
+        replacements,
```
tcodling, line 364: Bit more roundabout than id like but feels like the best solution given the current structure of suggester logic

dannash100: Could possibly use the extraReplacementsBuilder here too - 
like 
```
extraReplacementsBuilder: (query) => query.types.reduce((acc, value, index) => {
        acc[`type${index}`] = value;
        return acc;
      }, {}),
```

dannash100 (commented):
Just realised you might be able to use extra replacement builder for the order replacements too!

packages/facility-server/app/routes/apiv1/suggestions.js
tcodling: oooo ok

dannash100 (approved):
Looks good!

PR #7172, opened 2025-02-09
repo(scripts): generate TypeScript migrations

passcod:
### Changes

We can do this now, and it provides better IDE support.


dannash100 (approved):
beautiful

PR #7182, opened 2025-02-10
perf: switch UUID default to gen_random_uuid

passcod:
### Changes

In #6326, it was found that a low-hanging fruit for sync performance was to change the serial generation from UUID to a bigserial.

Subsequently, [in the slack thread for the PR](https://beyondessential.slack.com/archives/C01RZFLB3DZ/p1724368895597569), we found that the `uuid_generate_v4()` function from the (built-in) extension `uuid-ossp` is **3.5 times slower** than `gen_random_uuid` from the (also built-in) extension `pgcrypto`, _specifically on Windows_. On Linux they're the same (and also performance for that one function is 30% greater on Linux given the same resources).

This PR migrates every single table and model to `gen_random_uuid`. It also adds default values for the `id` to a number of tables which had forgotten to do that (it doesn't touch tables with e.g. generated columns where the omission is intentional).

As we move to Linux this will become less important, but in the meantime this may provide some performance benefit to our Windows deployments.

### DIFF
--- a/packages/central-server/app/tasks/VaccinationReminderProcessor.js
+++ b/packages/central-server/app/tasks/VaccinationReminderProcessor.js
@@ -71,7 +71,7 @@ export class VaccinationReminderProcessor extends ScheduledTask {
     )
     insert into patient_communications (id, created_at, updated_at, channel,  patient_id, destination, hash, status, type, subject, content)
     select
-      uuid_generate_v4() id,
+      gen_random_uuid() id,
       NOW() created_at,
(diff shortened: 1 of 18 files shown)


NavarroEmilioLuis (approved):
This LGTM but it leaves me with two questions:

- Is there any particular reason we don't iterate all tables in the same go, instead break them into categories? (string_ossp, uuid_ossp, string_no_default)
- Do we care/benefit changing types to one or the other?

PR #7183, opened 2025-02-10
db: add defaults to created_at and updated_at

passcod:
### Changes

When prototyping a change, I often add records in the database directly. Also, when writing raw queries in application code.

At those times, it's really annoying when the created_at and updated_at columns don't have defaults set. This fixes that.


NavarroEmilioLuis (approved):
Thanks... LGTM

PR #7187, opened 2025-02-10
feat(repeatingAppointments): NASS-1572: Clean up appointments using sync hook

dannash100:
### Changes

- Add sync hook that will soft delete any out of bounds appointments created by generate appointments task
- This is an edge case that is possible if a repeating appointment is cancelled see reproduction path at bottom
- Also adds new date columns:
  - Cancelled at date: populated as date of the cancellation when using `this and all future appointments`
  - Generated until date: populated each time appointments are bulk created, represents the date of latest appointment - i.e where generation is currently completed too. 

#### Reproduction path

1.  In central server config (in lens)
```
   "generateRepeatingAppointments": {
      // Not sure here maybe every min or 30s? does it have to be quicker or slower than sync
      "schedule": "*/30 * * * * *",
      "generateOffsetDays": 15, // instantly identify a 2 occurrence weekly repeating appointment as needing generating
    }
```

2. In admin panel set global setting
```
appointments: {
 maxRepeatingAppointmentsPerGeneration: 2
}
```
This will mean that repeating appointment will hit limit at second generated appointment - meaning that the task will kick in to generate another two

3. Generate repeating appointment with occurrence count 4 and weekly frequency (it should create 2 appointments hitting max)

4. Make sure it has synced once to central and then before the next sync cancel repeating appointment from first appointment into future

5. After sync ensure there are no appointments that have appeared at week 3 and 4 ( I will check database too )

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/models/AppointmentSchedule.ts
+++ b/packages/database/src/models/AppointmentSchedule.ts
@@ -345,12 +345,7 @@ export class AppointmentSchedule extends Model {
     if (isFullyGenerated) {
       await this.update({
         isFullyGenerated,
-        // If the schedule was generated with occurrenceCount, set the untilDate to the last appointment
-        // and clear the occurrenceCount
-        ...(this.occurrenceCount && {
-          generatedUntilDate: appointments.at(-1)!.startTime,
-          occurrenceCount: null,
-        }),
+        generatedUntilDate: appointments.at(-1)!.startTime,
       });
     }
     return appointments;
--- a/packages/database/src/models/AppointmentSchedule.ts
+++ b/packages/database/src/models/AppointmentSchedule.ts
@@ -202,11 +202,12 @@ export class AppointmentSchedule extends Model {
         },
       },
     });
-    await this.update({
+    const updatedSchedule = await this.update({
       isFullyGenerated: true,
       untilDate: previousAppointment ? previousAppointment.startTime : appointment.startTime,
       occurrenceCount: null,
     });
+    return updatedSchedule;
   }
 
   /**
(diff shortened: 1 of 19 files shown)


packages/database/src/models/AppointmentSchedule.ts
```diff
@@ -202,11 +214,12 @@ export class AppointmentSchedule extends Model {
         },
       },
     });
-    await this.update({
+    const updatedSchedule = await this.update({
       isFullyGenerated: true,
-      untilDate: previousAppointment ? previousAppointment.startTime : appointment.startTime,
+      cancelledAtDate: previousAppointment ? previousAppointment.startTime : appointment.startTime,
```
rohan-bes, line 219: Do we also need to be setting `untilDate` to this value as well? Otherwise the next time that the `GenerateRepeatingAppointments` scheduled task runs it might re-generate these?

rohan-bes (changes requested):
Looks great! This was a tricky corner case to reason about haha

Just one issue I think I noticed tho possibly I might not be following exactly how it's working

packages/database/src/models/AppointmentSchedule.ts
dannash100: Thanks for grabbing this PR. Ohh let me look into this I think you are right

rohan-bes: Ah yep that makes sense! Nice stuff :)

rohan-bes (approved):

PR #7193, opened 2025-02-11
refactor(db): KAM-192: time-travelling determinism check

passcod:
### Changes

The original (and current) determinism check suffers from a few issues:
- it requires a separate fake-data step
- it's confusing for developers when it fails
- it can't easily be run locally
- it fails on some migrations, e.g. most FHIR ones

It also suffered from a chicken-egg problem when it was first instated that influenced its design, in that the fake-data generator wasn't good enough to be usable until it was filled in by developers hitting against the determinism checker. This aspect is resolved: after a year, the generator now covers most of the database.

This PR introduces a new determinism checker, which uses time travel instead of a separate data gathering step:
- the first migration-touching commit in a PR is identified
- the repo is checked out to just before that commit
- the database is created and fake data inserted
- the repo is forwarded to the last commit in the PR
- the database is repeatedly migrated and checked for deviations

There are several advantages and new features with this:
- the entire check can easily be run locally
- when run locally the resulting databases are kept available, so you can inspect the database state directly to figure out what went wrong
- we now check the entire database, not just the tables the migrations touched directly. this could be useful with some triggers.
- we now only check tables sequelize cares about (models, excluding FHIR resources), we don't have to ignore tables manually
- you can backfill missing generator data with git rebase
  - that is, instead of using multiple branches or special git commit messages, you can instead more simply insert your generator-fixing commits before the first migration-touching commit (the script will tell you where that is).
- we can easily repeat the migration check several times. the new default is 10; the migrations thus run 11 times: once to hash the initial state, and 10 test rounds

```
=== Test round 10 ===
Create test database determinism-test-10
$ dropdb --if-exists determinism-test-10
$ createdb -O passcod -T determinism-test-init determinism-test-10
Migrate and hash the database
2025-02-12T04:23:47.575Z info: databaseConnection username=tamanu-facility host=null port=null name=determinism-test-10
2025-02-12T04:23:47.594Z info: registeringModels count=120
Applying migration: 1739280969900-testNonDeterministic
Post-migrations overall hash: cXbXqRPbhN88L0tD31x6CXmy+HhSwkPZzgdCzkmXaJk=

Resetting repo to refactor/time-travel-deterministic-check
$ git switch refactor/time-travel-deterministic-check

🎉 Determinism check passed!
```

We also add a few more improvements:
- if no migrations are present in a PR, we quit early

```
There is nothing touching migrations here, so there is nothing to check!
```

- if migration-touching commits are present, but the final state of the PR doesn't contain new migrations, we quit early

```
❎ Found some migration commits, but no actual migrations in the end state.
Resetting repo to refactor/time-travel-deterministic-check
$ git switch refactor/time-travel-deterministic-check
🎉 Determinism check passed!
```

- hash mismatches are presented table-by-table in the case of a failure

```
Post-migrations overall hash: WABnJCk15yK7INSeMLH4oEV7HsbGXhwgBCwTiiZTNzM=
!!! Found non-determinism !!!
determinism-test-0 -> determinism-test-1
--- 1739280969900-testNonDeterministic.js ---
mobzo1y59MgQR0tUDakTsFx0UO0snbA0bdaq9wLgvlw= -> WABnJCk15yK7INSeMLH4oEV7HsbGXhwgBCwTiiZTNzM=
table patients: mismatch! e6afdbd0d2a0c594089a2131ea19e7f8 != ba62f44abbed68c233726911c7193353
❌ See this slab page for potential solutions: https://beyond-essential.slab.com/posts/how-to-resolve-an-error-in-test-for-determinism-nwksh8cf
Resetting repo to refactor/time-travel-deterministic-check
```

- we fill the tables with more data. instead of a single row each, the generator now runs multiple rounds, by default 100. this provides more opportunities to catch subtle migration bugs, and locally you can use much larger loads for performance testing.

```
Fill database with fake data 100 rounds
....................................................................................................
```

### Deploys

### Remember to...

- ...write or update tests


packages/scripts/package.json
```diff
@@ -1,29 +1,30 @@
...
+    "build": "swc --out-dir dist --copy-files --source-maps true src",
+    "clean": "rimraf dist *.tsbuildInfo",
     "clean:deps": "rimraf node_modules",
-    "test": "tape 'tests/**/*.test.*js'",
-    "download-translations": "node src/downloadTranslations.js ../web/resources/strings",
-    "generate-fake-db": "node src/fake.js",
-    "test-determinism": "node src/testDeterminism.js"
+    "test": "npm run build && tape 'tests/**/*.test.*js'",
+    "download-translations": "npm run build && node dist/downloadTranslations.js ../web/resources/strings",
+    "test-determinism": "npm run build && NODE_CONFIG_DIR=../facility-server/config node dist/testDeterminism.js"
```
NavarroEmilioLuis, line 11: Hmmmm why do we specify the facility server config in `scripts`?

packages/constants/src/surveys.ts
```diff
@@ -187,7 +187,7 @@ export const READONLY_DATA_FIELDS = {
 };
 
 export const PROGRAM_REGISTRY_FIELD_LOCATIONS = Object.keys(PATIENT_DATA_FIELD_LOCATIONS).filter(
-  key => PATIENT_DATA_FIELD_LOCATIONS[key][0] === 'PatientProgramRegistration',
+  key => PATIENT_DATA_FIELD_LOCATIONS[key]![0] === 'PatientProgramRegistration',
```
NavarroEmilioLuis, line 190: Doesn't look right? Or is this something in TS?

NavarroEmilioLuis (commented):
LGTM! Only general note would be to keep the repo changes separate in the future. There is one place that didn't look right though

packages/constants/src/surveys.ts
passcod: Yes, ! in TS means "I know this isn't unrefined"

NavarroEmilioLuis: Ah okay - tried looking it up before asking but found nothing. I did manage to find it now, should anyone need it [here it is](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator)

NavarroEmilioLuis (approved):
LGTM then - the other question would be great to know but probably has a reason🤷

packages/scripts/package.json
passcod: Thinking about this has highlighted a further issue which is that we can't test whether migrations are deterministic between facility and central, which is kind of the entire point.

We can't take a central DB and compare it fully to a facility DB because they have different schemas (it used to be the case for a very long time that the schema would be identical but the data would be missing from one side, but that was incidental and informal and was broken in the past year, so now there's tables in central that don't exist at all in facility), and because there's data in central that doesn't exist in facilities and vice versa.

So the best we can do is check that central migrations are deterministic, and facility migrations are deterministic, and hope that nobody does something silly like

```
if (isFacility) {
	await q('UPDATE table SET id = sha1(id)');
} else {
    await q('UPDATE table SET id = md5(id)');
}
```

which would be 100% deterministic within each context but utterly break sync.

I'll change the script to be able to check both central and facility so we don't hard-code the node config dir in scripts like this.

edmofro: Fingers crossed that mayhem never occurs

passcod:
For clarity I've found a bug while using this for #7182 and am fixing that and the central/facility issue.

PR #7195, opened 2025-02-12
refactor(auth): NASS-1534: Add new auth package and move central code over

tcodling:
### Changes

### Deploys

### Remember to...

- ...write or update tests


packages/auth/src/login/login.ts
```diff
@@ -45,26 +49,26 @@ const getRefreshToken = async (models, { refreshSecret, userId, deviceId }) => {
...
   const { exp } = jwt.decode(refreshToken);
-  await RefreshToken.upsert(
-    {
-      refreshId: hashedRefreshId,
-      expiresAt: new Date(exp * 1000),
-      userId,
-      deviceId,
-    },
-    {
-      where: {
```
tcodling, line 56: Was throwing type error and I think (still to test) should work without

packages/auth/src/login/login.ts
```diff
@@ -99,50 +104,43 @@ export const login = ({ secret, refreshSecret }) =>
...
+        getLocalisation(),
+        getPermissionsForRoles(models, user.role),
+        models.Role.findByPk(user.role),
+      ]);
     // Send some additional data with login to tell the user about
     // the context they've just logged in to.
     res.send({
       token,
       refreshToken,
-      user: convertFromDbRecord(stripUser(user.get({ plain: true }))).data,
```
tcodling, line 139: Stripuser was stripping password which covertFromDbRecord does anyway

packages/auth/src/login/refresh.ts
```diff
@@ -12,14 +12,14 @@ import {
...
 
-export const refresh = ({ secret, refreshSecret }) =>
-  asyncHandler(async (req, res) => {
+export const refresh = ({ secret, refreshSecret }: { secret: string; refreshSecret: string }) =>
+  asyncHandler(async (req: any, res): Promise<void> => {
     const { body, store } = req;
     const { refreshToken, deviceId } = body;
 
-    const { canonicalHostName, auth } = config;
+    const { canonicalHostName, auth } = config as any;
```
tcodling, line 22: Both `req` and `config` will need to get their types properly handled but seems out of scope of this card

packages/auth/src/login/login.ts
```diff
@@ -45,26 +48,26 @@ const getRefreshToken = async (models, { refreshSecret, userId, deviceId }) => {
...
-  asyncHandler(async (req, res) => {
+export const login = ({
+  secret,
+  refreshSecret,
+  getLocalisation,
+}: {
+  secret: string;
+  refreshSecret: string;
+  getLocalisation: Function;
+}) =>
```
passcod, line 69: I prefer the pattern of having an `emport interface LoginArgs {...}` that's then used where needed

passcod: Like you could do

```ts
export interface AuthArgs {
  secret: string;
  refreshSecret: string;
}

export interface LoginArgs extends AuthArgs {
  getLocalisation: Function<foo>;
}
```

and the AuthArgs can be used directly by the refresh function as well

packages/auth/src/utils/convertDbRecord.ts
```diff
@@ -0,0 +1,47 @@
+type DbRecord = {
```
passcod, line 1: I think interfaces are prefered over typedefs in ts style

packages/auth/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "isolatedModules": false,
+    "noEmit": true,
+    "resolveJsonModule": true,
+    "moduleResolution": "NodeNext",
+    "module": "NodeNext",
+    "skipLibCheck": true,
+    "target": "es2020"
+  }
+}
+
```
passcod, line 20: OOS but we should really consider moving a bunch of those settings to be defaults in the common file.

passcod (commented):
Love it!

packages/auth/src/login/login.ts
tcodling: Ahh yep i got you 🙏

PR #7198, opened 2025-02-13
repo: set target for js compilation to es2023

passcod:
### Changes

As per <https://compat-table.github.io/compat-table/es2016plus/> Node 20 and Chrome both fully support ES2023. So let's standardise on that.

I've left the mobile target alone (currently at ES2019) because I'm not sure if we've managed to get it building with Node 20 (I remember we had issues some time back).

### DIFF
--- a/common.tsconfig.json
+++ b/common.tsconfig.json
@@ -7,15 +7,15 @@
     "ignoreDeprecations": "5.0",
     "verbatimModuleSyntax": false,
     "isolatedModules": true,
-    "module": "ES2020",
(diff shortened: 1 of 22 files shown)


dannash100:
@passcod I think mobile is on node 20 now - its package.json engine is set to 20.9.0

dannash100 (approved):
Looks great! Just that comment re mobile - I think we can change that too

passcod:
Okay, doing this a bit differently actually.

PR #7208, opened 2025-02-13
feat(labs): SAV-894: Add importer validation for sensitive labs

NavarroEmilioLuis:
### Changes

Adding validation to importer around sensitive lab test types. Note that this type of validation falls outside of the model level validation as it needs to look into other models or the whole spreadsheet info. We don't seem to have this kind of validation elsewhere (AFAIK) so possibly very rarely we want to do such a thing, but I tried making it easy to extend.

### Deploys


packages/central-server/app/admin/importer/validateTableRows.js
```diff
@@ -0,0 +1,65 @@
...
+    // This works because LabTestType is always upserted before panels
+    const { labTestTypeId } = values;
+    const labTestType = await models.LabTestType.findOne({ where: { id: labTestTypeId }});
+    if (labTestType?.isSensitive) {
+      pushErrorFn(-3, 'Lab test panels cannot contain sensitive lab test types');
+    }
+  }
+}
+
+const MODEL_VALIDATION = {
```
tcaiger, line 55: Nice idea to make this extensible

tcaiger (approved):
LGTM

PR #7231, opened 2025-02-17
feat(programRegistry): Epic Program Registries Phase 2

tcaiger:
Epic: Program Registries Phase 2

- [x] **Deploy (Klaus)**


packages/constants/src/programRegistry.ts
```diff
@@ -23,3 +23,15 @@ export const STATUS_COLOR = {
...
+export const PROGRAM_REGISTRY_CONDITION_CATEGORIES = {
+  Suspected: 'Suspected',
+  'Under investigation': 'Under investigation',
+  Confirmed: 'Confirmed',
+  Unknown: 'Unknown',
+  Disproven: 'Disproven',
+  Resolved: 'Resolved',
+  'In remission': 'In remission',
+  'Not applicable': 'Not applicable',
+  'Recorded in error': 'Recorded in error',
```
NavarroEmilioLuis, line 36: Unsure about these keys?

packages/facility-server/__tests__/apiv1/PatientProgramRegistration.test.js
```diff
@@ -163,7 +162,12 @@ describe('PatientProgramRegistration', () => {
         clinicianId: clinician.id,
         patientId: patient.id,
         date: '2023-09-02 08:00:00',
-        conditionIds: [programRegistryCondition.id],
+        conditions: [
+          {
+            conditionId: programRegistryCondition.id,
+            category: PROGRAM_REGISTRY_CONDITION_CATEGORIES.CONFIRMED,
```
NavarroEmilioLuis, line 168: How is this working?

NavarroEmilioLuis: Oh okay I see why, this is a POST, we're sending undefined, it probably saves it as `'Unknown'` and we never check for this value later 🤔

packages/constants/src/programRegistry.ts
tcaiger: I agree - as discussed, let's change them to camelCase keys

PR #7241, opened 2025-02-19
refactor(fakeData): KAM-331: Create fake-data package

tcodling:
### Changes

As part of our testing project for creating a synthetic environment, we have decided to pull out the logic for populating a DB instance with data used in determinism testing into its own package `fake-data`. This will be used for both determinism testing and will be used in the future to populate the synthetic test environment + any other use case for generating fake data.

- Create new package `fake-data`
- Move over the functions for populating the DB for determinism testing from `/scripts` to the new package
- Split up into functions that create various groups of linked data roughly in line with how they would be made realistically
  - Introduced some parameters with randomness 

Determinism testing was done and passed in a pr branched off this with dummy migration #7264 

### Deploys

### Remember to...

- ...write or update tests


packages/fake-data/src/populateDb/generateEachDataType.ts
tcodling, line 1: All files in this directory are helper functions built out of the original logic in the old scripts/src/fake.ts

packages/fake-data/src/populateDb/generateImportData.ts
```diff
@@ -0,0 +1,130 @@
...
+  LocationGroup,
+  Models,
+  ProgramRegistry,
+  ReferenceData,
+  ScheduledVaccine,
+  Survey,
+  User,
+} from '@tamanu/database';
+
+export const generateImportData = async ({
```
tcodling, line 18: Generate one of each import data types to be used for foreign key relationships when making clinical data

packages/fake-data/__tests__/setup.js
```diff
@@ -0,0 +1,2 @@
+// suppress warning about empty config
+process.env.SUPPRESS_NO_CONFIG_WARNING = 'true';
```
tcodling, line 2: No tests with this initial package setup. Tests coming in next pr with fake helper functions

packages/fake-data/src/populateDb/generateEachDataType.ts
tcodling, line 1: Tricky to break up in the most logical/easy to follow way. Currently I have essentially broken it up into functions covering:
- All import data
- Each type of clinical data group (lab request + lab test etc)
- A function that does all import data and one of each clinical data (used in determinism testing)

packages/fake-data/src/populateDb/helpers.ts
```diff
@@ -0,0 +1,435 @@
...
+  REPORT_DB_SCHEMAS,
+  REPORT_STATUSES,
+} from '@tamanu/constants';
+import type { Models, Encounter, Patient } from '@tamanu/database';
+
+// TODO: maybe this should come here?
+const { fake } = require('@tamanu/shared/test-helpers/fake');
+
+// Each function in this file takes the models for a DB instance + any relevant linking ids
+// depending on the data type
```
passcod, line 15: I feel like each function could be a separate file; it would make smallish files which isn't bad and be clearer for maintainability

packages/fake-data/src/populateDb/helpers.ts
```diff
@@ -0,0 +1,435 @@
...
+  IMAGING_TYPES,
+  NOTE_RECORD_TYPES,
+  REPEAT_FREQUENCY,
+  REPORT_DB_SCHEMAS,
+  REPORT_STATUSES,
+} from '@tamanu/constants';
+import type { Models, Encounter, Patient } from '@tamanu/database';
+
+// TODO: maybe this should come here?
+const { fake } = require('@tamanu/shared/test-helpers/fake');
```
passcod, line 12: It probably should come in if possible yes.

passcod (commented):
A good start!

packages/fake-data/src/populateDb/helpers.ts
tcodling: I have another pr specifically for this due to the number of files changed 🙏

packages/fake-data/src/populateDb/helpers.ts
tcodling: Hmm ok yea I was thinking this but didnt quite pull the trigger. Ill make that change

packages/fake-data/src/populateDb/helpers/importData.ts
tcodling, line 1: On the fence if I should have this file with all import data, or split it up amongst the other files in this directory in new functions

packages/fake-data/src/populateDb/helpers/encounter.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+export const createEncounter = async ({
+  models: { Encounter, EncounterHistory, Note, Discharge, EncounterDiagnosis },
+  patientId,
+  departmentId,
+  locationId,
+  userId,
+  referenceDataId,
+  noteCount = chance.integer({ min: 1, max: 5 }),
+  diagnosisCount = chance.integer({ min: 1, max: 5 }),
```
tcodling, line 25: I have done some simple randomness here but am a bit on the fence about how far I should take it

Do you think some relatively simple randomness like this is fine? Or do you think it is worth doing what we were talking about where you mentioned taking in a range or number and using that to generate the random number. Its easy enough to do but am a little worried about over-engineering 🤔

passcod: Nah the way you've got it is fine. We can always make it customisable later if we feel like it.

packages/fake-data/src/populateDb/helpers/importData.ts
passcod: I could see it either way or both but let's just have it all in here for now for simplicity.

passcod (approved):

PR #7242, opened 2025-02-20
feat(programRegistries): SAV-872: Update program registry layout

tcaiger:
### Changes

Update program registry view to match design.
The update related condition, program registry and history will come in future PRs.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/views/programRegistry/PatientProgramRegistrationSelectSurvey.jsx
+++ b/packages/web/app/views/programRegistry/PatientProgramRegistrationSelectSurvey.jsx
@@ -17,25 +17,14 @@ import { useTranslation } from '../../contexts/Translation';
 import { TranslatedText } from '../../components/Translation/TranslatedText';
 
 const DisplayContainer = styled.div`
-  display: flex;
-  width: 100%;
-  flex-direction: column;
-  justify-content: start;
-  align-items: flex-start;
   border: 1px solid ${Colors.softOutline};
-  font-size: 11px;
-  padding-left: 20px;
-  padding-top: 13px;
-  padding-right: 14px;
-  padding-bottom: 20px;
+  padding: 0 15px 20px 20px;
+  border-radius: 5px;
   background-color: ${Colors.white};
 `;
 
 const StyledFormGrid = styled(FormGrid)`
-  width: 100%;
-  display: grid;
   grid-template-columns: 80% 18%;
-  width: 100%;
   justify-content: space-between;
   align-items: flex-end;
 `;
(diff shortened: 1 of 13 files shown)


packages/web/app/views/programRegistry/PatientProgramRegistrationSelectSurvey.jsx
```diff
@@ -96,18 +82,20 @@ export const PatientProgramRegistrationSelectSurvey = ({ patientProgramRegistrat
...
                   label={
-                    <TranslatedText
-                      stringId="patientProgramRegistry.selectForm.label"
-                      fallback="Select form"
-                    />
+                    <Heading5 mt={1} mb={1}>
+                      <TranslatedText
+                        stringId="patientProgramRegistry.selectSurvey.label"
+                        fallback={`Select a ${patientProgramRegistration.programRegistry.name} form below to
+                      complete`}
```
NavarroEmilioLuis, line 89: Hmmm is that copy on figma? I think the biggest issue here is that having a translation would have a generic message, and here we are just pushing the program registry name without it being translated? I'd shoot back a message to see if we can do it generic on the fallback!

Oh okay on a second look this was already there as a heading... 🤔

packages/web/app/views/programRegistry/PatientProgramRegistryView.jsx
```diff
@@ -76,27 +90,29 @@ export const PatientProgramRegistryView = () => {
...
-            patientProgramRegistration={data}
-            programRegistryConditions={conditionOptions}
-          />
-          <ConditionSection
-            patientProgramRegistration={data}
-            programRegistryConditions={conditionOptions}
-          />
-        </ProgramStatusAndConditionContainer>
-
+          {/*Todo: Make first column fill if the second column isn't there */}
```
NavarroEmilioLuis, line 95: Just double checking this TODO is meant to be or if it was left behind?

NavarroEmilioLuis (approved):
Thanks for grabbing this one, the changes look great! I left two comments that are just to double check, but happy to carry on

packages/web/app/views/programRegistry/PatientProgramRegistryView.jsx
tcaiger: ah actually that is something that I need to do . Thanks for the reminder.

tcaiger: done

PR #7244, opened 2025-02-20
fix(fhir): SAV-901: Set published_date on a Lab Request when published via the FHIR API

rohan-bes:
### Changes

Previously when a Lab Request is transitioned to `Published` by an incoming `DiagnosticReport` we weren't setting the `published_date`. We should be doing this. 

### Deploys

### Remember to...

- ...write or update tests


packages/database/src/models/fhir/FhirDiagnosticReport.ts
```diff
@@ -114,6 +114,9 @@ export class FhirDiagnosticReport extends FhirResource {
 
       if (this.shouldUpdateLabRequestStatus(labRequest, newStatus)) {
         labRequest.set({ status: newStatus });
+        if (newStatus === LAB_REQUEST_STATUSES.PUBLISHED) {
+          labRequest.set({ publishedDate: new Date() });
```
NavarroEmilioLuis, line 118: Hmmmm but `publishedDate` seems to be a string, so we probably want the `getCurrentDateTimeString()` thing? Also assuming it's datetime and not just date here

NavarroEmilioLuis (commented):
Hmmmm I think there is one thing we need to double check that is probably a blocker sorry! By the way, curious that it kinda does work based on the test 😂

packages/database/src/models/fhir/FhirDiagnosticReport.ts
rohan-bes: Good call! I swapped this out and the tests still passed so maybe it would get handled anyway but best to be explicit that it's a string :+1:

packages/central-server/__tests__/hl7fhir/materialised/DiagnosticReport.test.js
```diff
@@ -169,6 +171,38 @@ describe('Create DiagnosticReport', () => {
...
+      body.status = FHIR_DIAGNOSTIC_REPORT_STATUS.FINAL;
+      body.presentedForm = [testAttachment];
+
+      const response = await app.post(endpoint).send(body);
+      await labRequest.reload();
+      expect(response).toHaveSucceeded();
+      expect(labRequest.status).toBe(LAB_REQUEST_STATUSES.PUBLISHED);
+      expect(labRequest.publishedDate).toBe(dateFnsTz.format(fakeDate, 'yyyy-MM-dd HH:mm:ss'));
+      expect((await labRequest.getLatestAttachment()).title).toBe(testAttachment.title);
+      jest.useRealTimers();
```
NavarroEmilioLuis, line 203: Hmmmm is this line really evaluated? I don't know when, but I think I recall that logs past the last `expect` get ignored? Just double check that it's working I think otherwise we could move it after we get the response hey?

NavarroEmilioLuis (approved):
Nice! Just one more note before you merge, but good to go!

packages/central-server/__tests__/hl7fhir/materialised/DiagnosticReport.test.js
rohan-bes: Okay now I'm confused, I removed the line and it seems like the tests work regardless. I was sure I had tested that previously and it was required... :shrug:

NavarroEmilioLuis: Oh well! Let's keep it for good measure? IDK?

PR #7250, opened 2025-02-20
repo: add SBOM and check license compliance

passcod:
### Changes

- Add an [SBOM](https://about.gitlab.com/blog/2022/10/25/the-ultimate-guide-to-sboms/#what-is-an-sbom%3F) to the repo and include it in distributed packages
- Add an npm script to regenerate the SBOM (no external tools needed)
- Add a CI check to ensure the SBOM is up to date
- Copy [the default `license.json` file from CycloneDX](https://github.com/CycloneDX/sbom-utility/blob/main/license.json)
- Specifically allow GPL-3.0-and-later software, given we're using that same license
- Add a CI check to ensure all licenses used as dependencies are approved

- [ ] TODO: normalise the sbom so it doesn't change on re-run


NavarroEmilioLuis (approved):
Interesting.. okay this looks good to me but you might want to involve a third person's sanity check.

I do have two broad questions:
- Were does the SBOM generator exist? in ops?
- Any reason we copy that `license.json` from that? I guess that's standard?

passcod:
SBOM generator is a built-in command to the `npm` CLI.

passcod:
We copy the license.json... yeah, because it's standard, and because afaics there's no way to say "use the standard list and change this one setting"

passcod:
This needs more thought, the generator isn't deterministic.

PR #7252, opened 2025-02-21
refactor(dataGeneration): KAM-337:: Move test-helper/fake to data generation package

tcodling:
### Changes

In this pr
- Moved fake.js to the data-generation package and converted to typescript
- Updated imports for all the files referencing it from shared to instead get from the data-generation package

Dont be scared by the file count! They are almost all imports! Scroll through to double check but since they are core parts of testing I am pretty confident as they are all passing. **Main thing to look out for is the types on `fake.ts` as everything else is just moving the package the file is in. I really struggled with some of this to be honest and had to do some hacky types that im not too happy with so totally open to suggestions**

### Deploys

### Remember to...

- ...write or update tests


packages/fake-data/src/fake/fake.ts
```diff
@@ -416,120 +419,117 @@ const MODEL_SPECIFIC_OVERRIDES = {
...
   const id = fakeUUID();
   const record = {};
   const modelOverridesFn = MODEL_SPECIFIC_OVERRIDES[model.name];
   const modelOverrides = modelOverridesFn ? modelOverridesFn(passedOverrides) : {};
   const overrides = { ...modelOverrides, ...passedOverrides };
   const overrideFields = Object.keys(overrides);
 
-  function fakeField(name, attribute) {
+  function fakeField(name: string, attribute: any) {
     const { type, fieldName, defaultValue } = attribute;
```
tcodling, line 533: This attribute argument caused a lot of confusion when typing this file. Attribute here seems like in theory it should be a `ModelAttributeColumnOptions` type from sequilize but we seem to be getting and using a fieldName which isnt on the type but still seems to exist?

I spent much of this pr learning about ts which has been awesome but i don't want to let it slow down the project so I decided to draw a line here. If the reviewer has any ideas I would love to have a look/pair. This function is just in a hacky type state atm

packages/fake-data/src/fake/fake.ts
```diff
@@ -579,8 +579,8 @@ export const fake = (model, passedOverrides = {}) => {
       return FIELD_HANDLERS[type.type](model, attribute, id);
     }
 
-    if (type instanceof DataTypes.STRING && type.options.length) {
-      return FIELD_HANDLERS['VARCHAR(N)'](model, attribute, id, type.options.length);
+    if (type instanceof DataTypes.STRING && (type as any).options.length) {
+      return FIELD_HANDLERS['VARCHAR(N)'](model, attribute, id, (type as any).options.length);
```
tcodling, line 583: Dont like these hacky types but i cant seem to type these at all properly. I believe this is meant to be a `Sequilize.DataType`which can either be a string or an object containing dataType information but I cant make these errors go away

packages/fake-data/src/fake/fake.ts
```diff
@@ -33,23 +33,24 @@ import {
...
 
 // this file is most commonly used within tests, but also outside them
 // jest wom't always be defined, in which case we can use a random seed
 export const chance = new Chance(global.jest ? jest.getSeed() : null);
 
-export function fakeStringFields(prefix, fields) {
+export function fakeStringFields(prefix: string, fields: string[]) {
   return fields.reduce(
-    (obj, field) => ({
+    (obj: object, field: string) => ({
```
dannash100, line 45: Think this is this
```suggestion
    (obj: Record<string, string>, field: string) => ({
```

packages/fake-data/src/fake/fake.ts
```diff
@@ -214,7 +215,8 @@ export function fakeEncounterMedication(prefix = 'test-') {
 }
 
 export const fakeDate = () => chance.date();
-export const fakeString = (model, { fieldName }, id) => `${model.name}.${fieldName}.${id}`;
+export const fakeString = (model: typeof Model, { fieldName }, id: string) =>
```
dannash100, line 218:
```suggestion
export const fakeString = (model: typeof Model, { fieldName } : { fieldName: string } , id: string) =>
```
I dunno actuly probably shud be typeof attributes sumhow but minor and just skip for now

packages/fake-data/src/fake/fake.ts
```diff
@@ -238,15 +240,16 @@ const FIELD_HANDLERS = {
...
+    fakeString(model, attrs, id).slice(0, length),
 
   TEXT: fakeString,
   INTEGER: fakeInt,
   FLOAT: fakeFloat,
   DECIMAL: fakeFloat,
   'TINYINT(1)': fakeBool,
   BOOLEAN: fakeBool,
-  ENUM: (model, { type }) => chance.pickone(type.values),
+  ENUM: (_: typeof Model, { type }) => chance.pickone(type.values),
```
dannash100, line 252: Do we need to type that unused one 🤔  mabye ts not smart enough

packages/fake-data/src/fake/fake.ts
dannash100: Ohh ignore my previous comment then, I remember, yeah the problem is that they havn't type that at all! so theres nothing really we can do.
They do define a type for `getAttributes` so mabye thats the move 🤔

packages/fake-data/src/fake/fake.ts
dannash100: You might be able to say its `AbstractDataType` at first - then u might not have to do the type as any

packages/fake-data/src/fake/fake.ts
```diff
@@ -599,7 +599,7 @@ export const fake = (model, passedOverrides = {}) => {
     );
   }
 
-  for (const [name, attribute] of Object.entries(model.tableAttributes)) {
+  for (const [name, attribute] of Object.entries(model.getAttributes())) {
```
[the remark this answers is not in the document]
dannash100: oh! yes perfect

dannash100 (approved):
Just some comments but looking great, im gonna give ✅

packages/fake-data/src/fake/fake.ts
tcodling: I guess we dont need to. We just get warning not error

packages/fake-data/src/fake/fake.ts
tcodling: Type and option dont seem to be valid children though 🤔

dannash100: I would think that `type instanceof` would coerce it into the more specific one automatically 🤔

packages/fake-data/src/fake/fake.ts
dannash100: Weird I cant see fieldName in the type defs at all either. Mabye just missed?

dannash100 (approved):

PR #7254, opened 2025-02-21
tweak(programRegistries): SAV-870: Update related conditions table

NavarroEmilioLuis:
### Changes

Organizes related conditions in two different categories plus UI changes

### DIFF
--- a/packages/web/app/views/programRegistry/ConditionSection.jsx
+++ b/packages/web/app/views/programRegistry/ConditionSection.jsx
@@ -3,12 +3,13 @@ import styled from 'styled-components';
 import { sortBy } from 'lodash';
 import { ButtonBase, Divider } from '@material-ui/core';
 import { PROGRAM_REGISTRY_CONDITION_CATEGORIES } from '@tamanu/constants';
-import { Heading5, TranslatedEnum, TranslatedReferenceData, TranslatedText } from '../../components';
+import { getReferenceDataStringId, Heading5, TranslatedText } from '../../components';
 import { usePatientProgramRegistryConditionsQuery } from '../../api/queries';
 import { LoadingIndicator } from '../../components/LoadingIndicator';
 import { ConditionalTooltip } from '../../components/Tooltip';
 import { Colors } from '../../constants';
 import useOverflow from '../../hooks/useOverflow';
+import { useTranslation } from '../../contexts/Translation';
 
 const Container = styled.div`
   display: flex;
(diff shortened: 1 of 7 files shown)


packages/web/app/views/programRegistry/ConditionSection.jsx
```diff
@@ -39,9 +40,52 @@ const ClippedConditionName = styled.span`
...
+// TODO: Translate program registry condition name and category, plus sort by that new display name
+const ConditionMapperFn = (condition) => {
+  const { programRegistryCondition, conditionCategory } = condition;
+  const { name } = programRegistryCondition;
+  const fullText = `${name} (${conditionCategory})`;
+  return (
+    <ConditionalTooltip title={fullText} visible={fullText.length > 65} key={condition.id}>
+      <Condition>
+        <ClippedConditionName>
+          {name} <ConditionCategory>({conditionCategory})</ConditionCategory>
```
tcaiger, line 82:
```suggestion
<TranslatedEnum value={conditionCategory} enumValues={PROGRAM_REGISTRY_CONDITION_CATEGORIES} />
```
As we discussed, we could use the TranslatedEnum here

tcaiger (changes requested):
Looks good overall - just have a couple of questions

packages/web/app/views/programRegistry/ConditionSection.jsx
```diff
@@ -67,17 +113,9 @@ export const ConditionSection = ({ patientProgramRegistration }) => {
...
-          const { name } = programRegistryCondition;
-          return (
-            <Condition key={condition.id}>
-              <ConditionalTooltip title={name} visible={name.length > 30}>
-                <ClippedConditionName>{name}</ClippedConditionName>
-              </ConditionalTooltip>
-            </Condition>
-          );
-        })}
+        {openConditions.map(ConditionMapperFn)}
```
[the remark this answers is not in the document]
NavarroEmilioLuis: By the way this was a great suggestion because it highlights the fact that we're not using a `key` for each condition. I think that in this case we will just have to bypass it anyway, but it's good that it came up to the surface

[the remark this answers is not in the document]
NavarroEmilioLuis: Lol nevermind I was adding the key anyway on the mapper 😅 it still good and I'll still make the change though

packages/web/app/views/programRegistry/ConditionSection.jsx
NavarroEmilioLuis: Okay so this doesn't really work because we're saving the "labels" as values on the database, so the keys do not match the enums 🤔 I think the proper way to go is to follow the same pattern elsewhere, have a constant with the values we'll save in the database, then have a constant that maps labels

NavarroEmilioLuis: Or maybe these were just saved before the change 🤔 I think it might be that by looking at the code hey

NavarroEmilioLuis: Yep, seems to be that. Ran these queries to cleanup the epic branch:

```
app=# UPDATE patient_program_registration_conditions SET condition_category = 'underInvestigation' WHERE condition_category = 'Under investigation';
UPDATE 2
app=# UPDATE patient_program_registration_conditions SET condition_category = 'disproven' WHERE condition_category = 'Disproven';
UPDATE 1
app=# UPDATE patient_program_registration_conditions SET condition_category = 'unknown' WHERE condition_category = 'Unknown';
UPDATE 3
app=# UPDATE patient_program_registration_conditions SET condition_category = 'confirmed' WHERE condition_category = 'Confirmed';
UPDATE 1
```

tcaiger: Thanks 🙏

tcaiger (approved):
Happy for you to make that one change when you are ready :)

tcaiger (approved):
extra changes lookn good

PR #7265, opened 2025-02-25
feat(fakeData): KAM-338: Populate db from log request tally

tcodling:
### Changes

This is as far as i got during my shift as seconded dev on Kamaka. Its not fully fleshed out but this pr is a start on a function that takes a json file of request counts for each model and populates a db instance based on those. I also made a few changes to the helper functions to make them work without supplying any parameters (they just grab random ids if none are explicitly supplied)

I think this has lots of potential for improvement like:
- Possibly matching endpoints rather than model names if we want a more accurate simulation
- Only has POSTs right now. It should also have PUTs
- Should cover all data types not just the few with helpers in the fake data package
- Even more optional parameters + randomness

### Deploys

### Remember to...

- ...write or update tests


packages/fake-data/src/populateDb/parseTally/populateFromLogTally.ts
```diff
@@ -0,0 +1,61 @@
...
+  Patient: { POST: createPatient, PUT: () => null },
+  Encounter: { POST: createEncounter, PUT: () => null },
+  ImagingRequest: { POST: createImagingRequest, PUT: () => null },
+  Invoice: { POST: createInvoice, PUT: () => null },
+  LabRequest: { createLabRequest, PUT: () => null },
+  ProgramRegistry: { POST: createProgramRegistry, PUT: () => null },
+  Survey: { POST: createSurveyResponse, PUT: () => null },
+  Tasking: { POST: createTask, PUT: () => null },
+  Vaccine: { POST: createAdministeredVaccine, PUT: () => null },
+  ReportDefinition: { POST: createDbReport, PUT: () => null },
```
tcodling, line 33: Not 100% sure this is the best structure but i like the idea of a big dictionary of user actions to simulate

passcod (approved):

PR #7259, opened 2025-02-24
feat: KAM-336: add devices table

passcod:
### Changes

This is a partial implementation of KAM-336 and thus SAV-456, which adds a `devices` table but not the login, cryptography, nor quota machinery (even though it pre-emptively documents all that in the dbt docs).

The two internal sync tables `sync_queued_devices` and `sync_device_ticks` are made to use fkeys to the `devices` table.

**This PR as-is breaks sync in non-test environments** as the devices just plain aren't created. However, as all the (jest) tests pass and "[the rest of the owl](https://knowyourmeme.com/memes/how-to-draw-an-owl)" would complicate the PR, it is submitted for review (probably to be merged into an epic).

### Deploys

- [ ] **Deploy to Tamanu Internal** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

PR #7268, opened 2025-02-26
feat(audit): NASS-1598: Attach userid to postgres session for audit logs

dannash100:
### Changes

requiring some work from https://github.com/beyondessential/tamanu/pull/7259/files#diff-bef6833a2ad9868ab6a329f02875458fff364a496d51da73f1e4e304a5889f08

Attach current user id to postgres session - while ensuring that there is no mismatch even with more active sessions that max in pool and random 0-3 second latency in endpoints.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/config/test.json5
+++ b/packages/central-server/config/test.json5
@@ -34,7 +34,7 @@
     "useHardcodedPermissions": false
   },
   "log": {
-    "level": "warn"
+    "consoleLevel": "warn"
   },
   "cors": {
     "allowedOrigin": "https://fake-place-xxx-yyy.com"
--- a/packages/central-server/config/test.json5
+++ b/packages/central-server/config/test.json5
@@ -34,7 +34,7 @@
     "useHardcodedPermissions": false
   },
   "log": {
-    "consoleLevel": "warn"
+    "level": "warn"
   },
   "cors": {
     "allowedOrigin": "https://fake-place-xxx-yyy.com"
--- a/packages/database/src/migrations/1739970132205-addAuditTrigger.ts
+++ b/packages/database/src/migrations/1739970132205-addAuditTrigger.ts
@@ -1,12 +1,12 @@
 import { QueryInterface } from 'sequelize';
 
 export async function up(query: QueryInterface): Promise<void> {
-  await query.createFunction(
-    'logs.record_change',
-    [],
-    'trigger',
-    'plpgsql',
-    `
+    await query.createFunction(
+      'logs.record_change',
+      [],
+      'trigger',
+      'plpgsql',
+      `
       INSERT INTO logs.changes (
         table_oid,
         table_schema,
@@ -20,25 +20,27 @@ export async function up(query: QueryInterface): Promise<void> {
         record_id,
         record_update,
         record_data
-      ) VALUES (
-        TG_RELID,                 -- table_oid
-        TG_TABLE_SCHEMA,          -- table_schema
-        TG_TABLE_NAME,            -- table_name
-        CURRENT_TIMESTAMP,        -- logged_at
-        NEW.created_at,           -- created_at
-        NEW.updated_at,           -- updated_at
-        NEW.deleted_at,           -- deleted_at
-        NEW.updated_at_sync_tick, -- updated_at_sync_tick
-        NEW.updated_by_user_id,   -- updated_by_user_id
-        NEW.id,                   -- record_id
-        TG_OP = 'UPDATE',         -- record_update
-        to_jsonb(NEW.*)           -- record_data
-      );
-      RETURN NEW;
-    `,
-  );
+        ) VALUES (
+          TG_RELID,                 -- table_oid
+          TG_TABLE_SCHEMA,          -- table_schema
+          TG_TABLE_NAME,            -- table_name
+          CURRENT_TIMESTAMP,        -- logged_at
+          NEW.created_at,           -- created_at
+          NEW.updated_at,           -- updated_at
+          NEW.deleted_at,           -- deleted_at
+          NEW.updated_at_sync_tick, -- updated_at_sync_tick
+          coalesce(current_setting('audit.userid', true), uuid_nil()::text), -- updated_by_user_id
+          NEW.id,                   -- record_id
+          TG_OP = 'UPDATE',         -- record_update
+          to_jsonb(NEW.*)           -- record_data
+          );
+          RETURN NEW;
+          `,
+        );
+
 }
 
 export async function down(query: QueryInterface): Promise<void> {
   await query.dropFunction('logs.record_change', []);
 }
+
--- a/packages/database/src/migrations/1739970132205-addAuditTrigger.ts
+++ b/packages/database/src/migrations/1739970132205-addAuditTrigger.ts
@@ -30,7 +30,7 @@ export async function up(query: QueryInterface): Promise<void> {
           NEW.updated_at,           -- updated_at
           NEW.deleted_at,           -- deleted_at
           NEW.updated_at_sync_tick, -- updated_at_sync_tick
-          coalesce(current_setting('${AUDIT_USERID_KEY}', true), uuid_nil()::text), -- updated_by_user_id
+          coalesce(nullif(current_setting('${AUDIT_USERID_KEY}', true), '') uuid_nil()::text), -- updated_by_user_id
           NEW.id,                   -- record_id
           TG_OP = 'UPDATE',         -- record_update
           to_jsonb(NEW.*)           -- record_data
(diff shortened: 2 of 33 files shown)


packages/database/src/migrations/1739970132205-addAuditTrigger.ts
```diff
@@ -20,25 +20,25 @@ export async function up(query: QueryInterface): Promise<void> {
...
+        ) VALUES (
+          TG_RELID,                 -- table_oid
+          TG_TABLE_SCHEMA,          -- table_schema
+          TG_TABLE_NAME,            -- table_name
+          CURRENT_TIMESTAMP,        -- logged_at
+          NEW.created_at,           -- created_at
+          NEW.updated_at,           -- updated_at
+          NEW.deleted_at,           -- deleted_at
+          NEW.updated_at_sync_tick, -- updated_at_sync_tick
+          coalesce(current_setting('audit.userid', true), uuid_nil()::text), -- updated_by_user_id
```
passcod, line 32: let's use `tamanu.audit.userid` so we can make 100% sure we don't have namespacing conflicts

passcod: Also may want to pull that into a constant

packages/database/src/services/database.js
```diff
@@ -17,14 +17,16 @@ createDateTypes();
...
-Sequelize.useCLS({
+export const namespace = {
   bind: () => {}, // compatibility with cls-hooked, not used by sequelize
   get: (id) => asyncLocalStorage.getStore()?.get(id),
   set: (id, value) => asyncLocalStorage.getStore()?.set(id, value),
   run: (callback) => asyncLocalStorage.run(new Map(), callback),
-});
+};
+
+export const asyncLocalStorage = new AsyncLocalStorage();
```
passcod, line 27: is there a reason why the const asyncLocalStorage is now _after_ the const namespace where it is used?

packages/facility-server/app/middleware/audit.js
```diff
@@ -0,0 +1,9 @@
+import { namespace } from '@tamanu/database/services/database';
+
+export const attachAuditUserToDbSession = async (req, _res, next) => {
```
dannash100, line 3: will move to shared and apply on central next

packages/database/src/services/database.js
```diff
@@ -108,6 +110,16 @@ async function connectToDatabase(dbOptions) {
...
   });
+
+  class QueryWithAditConfig extends sequelize.dialect.Query {
+    async run(sql, options) {
+      const userid = namespace.get('audit.userid');
+      if (userid) await super.run(`SELECT set_config('audit.userid', $1, false)`, [userid]);
+      return super.run(sql, options);
+    }
+  }
+  sequelize.dialect.Query = QueryWithAditConfig;
```
passcod, line 121: We should add some kind of comment here to explain what's happening (ie to explain that this is called by sequelize internals with the sql string for every single query done via sequelize).

packages/database/src/services/migrations/constants.ts
```diff
@@ -36,4 +36,7 @@ export const NON_LOGGED_TABLES = [
 
   // internal and also signers.privateKey needs to be hard-deletable
   'signers',
+
+  // internal configuration
+  'local_system_facts'
```
passcod, line 41:
```suggestion
  'local_system_facts',
```

packages/database/src/services/database.js
```diff
@@ -108,6 +110,16 @@ async function connectToDatabase(dbOptions) {
     logging,
     pool,
   });
+
+  class QueryWithAditConfig extends sequelize.dialect.Query {
+    async run(sql, options) {
+      const userid = namespace.get('audit.userid');
+      if (userid) await super.run(`SELECT set_config('audit.userid', $1, false)`, [userid]);
```
dannash100, line 117: Might actually have to set this in every case, as I think the config might stick if no authed user

packages/facility-server/__tests__/setupUtilities.js
```diff
@@ -6,15 +6,27 @@ import { FAKE_UUID_PATTERN } from '@tamanu/utils/generateId';
...
         [Op.like]: FAKE_UUID_PATTERN,
       }),
-    }),
-  );
+    ]
+    if (Model.name === 'User') {
+      where.push({
+        displayName: {
+          [Op.not]: 'System',
+        },
```
passcod, line 21: Hmm I think better to match on the `SYSTEM_USER_UUID` id instead of displayName

packages/facility-server/app/routes/apiv1/index.js
```diff
@@ -108,6 +109,8 @@ apiv1.use(authMiddleware);
 
 apiv1.use(constructPermission);
 
+apiv1.use(attachAuditUserToDbSession)
```
passcod, line 112:
```suggestion
apiv1.use(attachAuditUserToDbSession);
```

passcod (changes requested):
very cool

packages/facility-server/app/routes/apiv1/index.js
dannash100: Ahh i turned of lint of save as VS code destroying computer

packages/facility-server/__tests__/setupUtilities.js
dannash100: Oh yeah ohk

packages/database/src/services/database.js
passcod: Does it work if we set it to null?

packages/database/src/services/database.js
dannash100: moved

packages/database/src/services/database.js
```diff
@@ -108,6 +110,16 @@ async function connectToDatabase(dbOptions) {
     logging,
     pool,
   });
+
+  class QueryWithAditConfig extends sequelize.dialect.Query {
+    async run(sql, options) {
+      const userid = namespace.get('tamanu.audit.userid');
+      if (userid) await super.run(`SELECT set_config('tamanu.audit.userid', $1, false)`, [userid]);
+      return super.run(sql, options);
```
passcod, line 118: Hmm I think this being a separate awaited run will not work. Hope I'm wrong, but can you test by putting a sleep after the if?

dannash100: ahh shoot - ohk i did wonder. You can't put multiple statements in this run. So perhaps we would have to incorporate into query silently sumhow

dannash100: Mabye like 
```
WITH _ AS (SELECT set_config('tamanu.audit.userid', $1, false))
```
but this will goof up if it starts with WITH

mabye 😓 
```
`WITH _ AS (SELECT set_config('tamanu.audit.userid', $1, false)) {sql.replace(/^WITH/, '')}`
```

dannash100: Actully to work fully it would need to look like that

```js
  const auditConfig = `WITH _audit AS (SELECT set_config('${AUDIT_USERID_KEY}', '${userid}', false))`;
  if (sql.startsWith('WITH')) {
    sql = sql.replace(/^WITH/i, `${auditConfig},`);
  } else if (/^(UPDATE|INSERT|DELETE|MERGE)/i.test(sql)) {
    sql = `${auditConfig} ${sql}`;
  }
  return super.run(sql, options)
```

Ill do test with async sleep in between the two statements tho as its pretty bad feels

dannash100: Scrap that it works as you stated: PHEW
```
dialect.Query is created and run on a single acquired connection from the pool, which is then released later. So await is used, but the single connection guarantees that we're running sequentially in the one context.
```

packages/database/src/migrations/1739970132205-addAuditTrigger.ts
```diff
@@ -20,25 +21,25 @@ export async function up(query: QueryInterface): Promise<void> {
...
+        ) VALUES (
+          TG_RELID,                 -- table_oid
+          TG_TABLE_SCHEMA,          -- table_schema
+          TG_TABLE_NAME,            -- table_name
+          CURRENT_TIMESTAMP,        -- logged_at
+          NEW.created_at,           -- created_at
+          NEW.updated_at,           -- updated_at
+          NEW.deleted_at,           -- deleted_at
+          NEW.updated_at_sync_tick, -- updated_at_sync_tick
+          coalesce(current_setting('${AUDIT_USERID_KEY}', true), uuid_nil()::text), -- updated_by_user_id
```
passcod, line 33:
```suggestion
          coalesce(nullif(current_setting('${AUDIT_USERID_KEY}', true), '') uuid_nil()::text), -- updated_by_user_id
```

This ensures that we default to uuid_nil even if the setting's value is not null, but is the empty string... as this is what happens when `set_config('foo', null)` is called.

passcod (changes requested):
one last thing

passcod (approved):

packages/central-server/config/test.json5
```diff
@@ -34,7 +34,7 @@
     "useHardcodedPermissions": false
   },
   "log": {
-    "consoleLevel": "warn"
+    "level": "warn"
```
passcod, line 37: wait, what

dannash100: I got confused never mind

passcod (approved):

passcod (approved):

PR #7278, opened 2025-02-28
fix(vaccines): TAMOC-275: Fix broken vaccine module in mobile (HOTFIX 2.19)

chris-bes:
### Changes
- Fix broken vaccine module in mobile

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/mobile/App/models/Setting.ts
+++ b/packages/mobile/App/models/Setting.ts
@@ -8,6 +8,7 @@ import { SYNC_DIRECTIONS } from './types';
 import { IFacility } from '../types';
 import { SETTINGS_SCOPES } from '~/constants';
 import { readConfig } from '~/services/config';
+import { parseOrKeep } from '~/utils/parseOrKeep';
 
 @Entity('setting')
 export class Setting extends BaseModel {
@@ -78,7 +79,7 @@ export class Setting extends BaseModel {
 
     const settingsObject = {};
     for (const currentSetting of settings) {
-      setAtPath(settingsObject, currentSetting.key, JSON.parse(currentSetting.value ?? null));
+      setAtPath(settingsObject, currentSetting.key, parseOrKeep(currentSetting.value));
     }
 
     if (key === '') {
--- a/packages/mobile/App/utils/parseOrKeep.ts
+++ b/packages/mobile/App/utils/parseOrKeep.ts
@@ -1,0 +1,7 @@
+export function parseOrKeep(value: string): string | object {
+  try {
+    return JSON.parse(value);
+  } catch (e) {
+    return value;
+  }
+}


packages/mobile/App/models/Setting.ts
```diff
@@ -78,7 +79,7 @@ export class Setting extends BaseModel {
 
     const settingsObject = {};
     for (const currentSetting of settings) {
-      setAtPath(settingsObject, currentSetting.key, JSON.parse(currentSetting.value ?? null));
```
chris-bes, line 81: If currentSetting.value is not a json string, this will fail...

NavarroEmilioLuis: Hmmm if `currentSetting.value` is falsy then we will keep the exact value instead of using `null`, which might not be great, but it's not same behavior!

NavarroEmilioLuis (changes requested):
Well, looks good just one thing to maintain the same behavior unless we're cool to break

packages/mobile/App/models/Setting.ts
chris-bes: Hmmm if a setting's value is deliberately set to `false`, I would think that's the value that they intentionally want to set right? Why would we want to use `null` instead in that case?

packages/mobile/App/utils/parseOrKeep.ts
```diff
@@ -0,0 +1,7 @@
+export function parseOrKeep(value: string): string | object {
+  try {
+    return JSON.parse(value);
+  } catch (e) {
+    return value;
```
NavarroEmilioLuis, line 5: Also I think we should be double checking a couple of things here possibly 🤔 do you have examples of what values were causing the error? Should we validate they are a specific type, etc.

packages/mobile/App/models/Setting.ts
NavarroEmilioLuis: Hmmmm no clue really. But this seems to be already what it is on main, so this is backporting hey?

packages/mobile/App/utils/parseOrKeep.ts
chris-bes: Yeah so anything that is not JSON format would cause issue. We have a few in settings like emails (string), feature flags ( (boolean true/false),...

packages/mobile/App/models/Setting.ts
chris-bes: Yeah that's right, this was for fixing another issue before, but now another setting is causing a problem in vaccine modules. Perhaps I should also backport it back to the version that this was introduced which is 2.18

NavarroEmilioLuis (approved):
Yeah okay happy to carry on!

PR #7280, opened 2025-02-28
feat: EPI-1094: Create MAR view and display medications

huytranbes:
### Changes

- Add MAR route and view
- Build a date selector and time windows header
- List medications (grouped by scheduled and PRN) on the MAR that have start date before the selected date

### Deploys

### Remember to...

- ...write or update tests


packages/web/app/forms/MedicationForm.jsx
```diff
@@ -497,6 +498,12 @@ export const MedicationForm = ({ encounterId, onCancel, onSaved }) => {
     setSubmittedMedication(newMedication);
   };
 
+  const onFinalise = async ({ data, isPrinting, submitForm }) => {
+    setAwaitingPrint(isPrinting);
+    await submitForm(data);
+    await queryClient.invalidateQueries(['encounterMedication', encounterId]);
```
tcaiger, line 504: This is an optional suggestion but invalidateQueries could go in the onSuccess callback. It might catch an unexpected edge case or make it less likely someone misses onFinalise in future

packages/web/app/views/patients/panes/EncounterMedicationPane.jsx
```diff
@@ -98,6 +104,7 @@ export const EncounterMedicationPane = React.memo(({ encounter, readonly }) => {
             disabled={readonly}
             variant="outlined"
             color="primary"
+            onClick={onMarClick}
```
tcaiger, line 107:
```suggestion
            onClick={navigateToMar}
```
You could simplify this by passing navigateToMar to onClick

packages/web/app/views/patients/medication/MarTable.jsx
```diff
@@ -0,0 +1,271 @@
...
+  medication,
+  discontinued,
+}) => {
+  const doseAmountDisplay = isPrn ? (
+    <TranslatedText stringId="medication.table.variable" fallback="Variable" />
+  ) : (
+    doseAmount
+  );
+  return (
+    <React.Fragment key={id}>
```
tcaiger, line 149:
```suggestion
    <>
```
You don't need the key here because it is used in the look where the component is called

packages/web/app/views/patients/medication/MarTable.jsx
```diff
@@ -0,0 +1,271 @@
...
+      <MedicationCellContainer discontinued={discontinued}>
+        <Box fontWeight={500}>{medication.name}</Box>
+        <Box>
+          <span>{doseAmountDisplay}</span>
+          <span>{` ${units}, ${frequency}, ${DRUG_ROUTE_LABELS[route]}`}</span>
```
tcaiger, line 154:
```suggestion
          <span> {units}, {frequency}, {DRUG_ROUTE_LABELS[route]</span>
```
This could be simplified

packages/web/app/views/patients/medication/MarTable.jsx
```diff
@@ -0,0 +1,271 @@
...
+
+const MedicationGrid = styled.div`
+  display: grid;
+  grid-template-columns: minmax(100px, 1fr) repeat(12, 48px);
+`;
+
+const CurrentTimeOverlay = styled.div`
+  position: absolute;
+  top: 0;
+  width: 48px;
```
tcaiger, line 121: We should put the 48px in a constant at the top of the file since it is important for the layout

packages/web/app/views/patients/medication/MarTable.jsx
```diff
@@ -0,0 +1,271 @@
...
+        <TimeSlotLabel>{periodLabel || ''}</TimeSlotLabel>
+        <div>{`${formatTime(startTime)} - ${formatTime(endTime)}`}</div>
+      </TimeSlotText>
+    </TimeSlotHeaderContainer>
+  );
+};
+
+export const MarTable = ({ selectedDate }) => {
+  const { encounter } = useEncounter();
+  const [currentTimeSlotIndex, setCurrentTimeSlotIndex] = useState(-1);
```
tcaiger, line 188: The current time slot is derived from existing values so it doesn't need to go into state. We could do something like this which would simplify the code

const currentTimeSlotIndex  = MEDICATION_ADMINISTRATION_TIME_SLOTS.findIndex(({ periodLabel, startTime, endTime }) => (
return startDate <= Date.now() && Date.now() <= endDate;

tcaiger (changes requested):
Great work - this is a complex feature. I have left a couple of suggestions - most important is the currentIndex suggestion.

packages/web/app/views/patients/medication/MarTable.jsx
huytranbes: Thanks for your nice suggestions. I have resolved all of them

tcaiger (approved):
Nice work!

PR #7284, opened 2025-03-02
feat(programRegistries): SAV-868: Program registry put endpoint

tcaiger:
### Changes

- Split out `patientProgramRegistrationConditions` endpoint from `patientProgramRegistration` and into their own file
- add new program registry put endpoint

### Deploys

- [ ] **Deploy to Tamanu Internal** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

PR #7285, opened 2025-03-02
feat(programRegistries): SAV-868: Program registry put endpoint #7284

tcaiger:
### Changes

Split out patientProgramRegistrationConditions endpoint from patientProgramRegistration and into their own file
add new program registry put endpoint
### Deploys

### Remember to...

- ...write or update tests


packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistrationConditions.js
```diff
@@ -0,0 +1,110 @@
+import express from 'express';
```
tcaiger, line 1: All of the endpoints in this file are unchanged

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistration.js
```diff
@@ -111,6 +93,59 @@ patientProgramRegistration.post(
   }),
 );
 
+patientProgramRegistration.put(
```
tcaiger, line 96: This is the new endpoint

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistration.js
```diff
@@ -111,6 +93,59 @@ patientProgramRegistration.post(
...
+      await models.PatientProgramRegistration.findByPk(programRegistrationId);
+
+    if (!existingRegistration) {
+      throw new NotFoundError('PatientProgramRegistration not found');
+    }
+
+    const { patientId, programRegistryId } = existingRegistration;
+
+    const conditionsData = conditions
+      .filter((condition) => condition.id)
```
rohan-bes, line 118: Out curiosity, when would we be passing in conditions without an `id`?

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistration.js
```diff
@@ -111,6 +93,59 @@ patientProgramRegistration.post(
...
+          updateOnDuplicate: ['date', 'conditionCategory', 'reasonForChange'],
+        }),
+      ]);
+    });
+
+    // Convert Sequelize model to use a custom object as response
+    const responseObject = {
+      ...registration.get({ plain: true }),
+      conditions: conditionsRecords,
+    };
```
rohan-bes, line 143: Not entirely sure what the expected behaviour would be here. With the behaviour as it is currently, if the conditions weren't changed we'd return with a `conditions: []` field which feels a little misleading.

Thoughts on just returning the `registration`? Or returning registration and conditions if conditions have been updated?

rohan-bes (approved):
Looks good! Nice tests :)

One comment about the return object of the `PUT` endpoint but nothing major

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistrationConditions.js
rohan-bes: Thanks for breaking these out into a separate file :pray:

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistration.js
tcaiger: There isn't actually a valid scenario when that should happen. My intention was validation so I've updated this to throw an error if there are any conditions without an id

packages/facility-server/app/routes/apiv1/patient/patientprogramRegistration/patientProgramRegistration.js
tcaiger: You are right - I think it should return the registration record with the conditions attached or just the registration record by itself. I decided on the simple record option as we aren't doing anything with this response so it seems unnecessary to have an extra DB query and have extra code to maintain.

PR #7291, opened 2025-03-03
feat(web): support Chromium and MS Edge

passcod:
### Changes

These are both essentially Chrome in another skin, so are no additional support (testing) burden. In fact, we already say we support Chromium in some docs, but didn't actually.

(TBH the site also works perfectly well in Firefox etc, but whatever)

### DIFF
--- a/packages/web/app/App.jsx
+++ b/packages/web/app/App.jsx
@@ -44,8 +44,11 @@ export function App({ sidebar, children }) {
     localStorage.getItem('DISABLE_SINGLE_TAB') || process.env.DISABLE_SINGLE_TAB === 'true';
 
   const browser = Bowser.getParser(window.navigator.userAgent);
-  const isChrome = browser.satisfies({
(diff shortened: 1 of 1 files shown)


dannash100 (approved):
## Chromish

packages/web/app/App.jsx
```diff
@@ -54,7 +57,7 @@ export function App({ sidebar, children }) {
   if (!isDebugMode) {
     // Skip browser/platform check in debug mode
     if (!isDesktop) return <MobileStatusPage platformType={platformType} />;
-    if (!isChrome) return <UnsupportedBrowserStatusPage />;
+    if (!isChromish) return <UnsupportedBrowserStatusPage />;
```
edmofro, line 60: UpsupportedBrowserStatusPage also needs a tweak to add Edge to the copy 🙏

edmofro: @passcod can you throw up a new PR for this?

PR #7296, opened 2025-03-04
feat(audit): NASS-1598: Pause audit log during facility <-> central sync persist

dannash100:
### Changes
- Utils for getting and setting session config var on both postgres and sequelize async storage 
- Sets a variable in transaction context to pause the creation of changes audit logs

issues fixed by [feat(audit): NASS-1599: Add migration audit log](https://github.com/beyondessential/tamanu/pull/7306#top)

Few last things to do: 
- Fix determinism re sequelize meta table

### Deploys

### Remember to...

- ...write or update tests


packages/database/src/utils/audit/pauseAuditForTransaction.ts
```diff
@@ -0,0 +1,6 @@
+import type { Sequelize } from 'sequelize';
+import { AUDIT_PAUSE_KEY } from '@tamanu/constants/audit';
+
+export const pauseAuditForTransaction = async (sequelize: Sequelize): Promise<void> => {
+  await sequelize.query(`SELECT set_config('${AUDIT_PAUSE_KEY}', 'true', true)`);
```
dannash100, line 5: the third param means only applies to transaction. I can document that tho

packages/database/src/migrations/1739970132205-addAuditTrigger.ts
```diff
@@ -8,35 +8,41 @@ export async function up(query: QueryInterface): Promise<void> {
...
-        NEW.updated_at_sync_tick, -- updated_at_sync_tick
-        coalesce(nullif(current_setting('${AUDIT_USERID_KEY}', true), ''), uuid_nil()::text), -- updated_by_user_id
-        NEW.id,                   -- record_id
-        TG_OP = 'UPDATE',         -- record_update
-        to_jsonb(NEW.*)           -- record_data
-      );
-      RETURN NEW;
-    `,
+      BEGIN
+        IF (SELECT coalesce(current_setting('${AUDIT_PAUSE_KEY}', true), 'false')::boolean) THEN
```
passcod, line 12:
```suggestion
        IF (SELECT coalesce(nullif(current_setting('${AUDIT_PAUSE_KEY}', true), ''), 'false')::boolean) THEN
```

This also needs the nullif

packages/database/src/utils/audit/pauseAuditForTransaction.ts
passcod: Hmmm maybe let's do something like:

```ts
sequelize.setSessionVar = (key: string, value: any) => sequelize.query(
	`SELECT set_config($key, $value)`,
	{ bind: { key, value } },
);

sequelize.setTransactionVar = (key: string, value: any) => sequelize.query(
	`SELECT set_config($key, $value, true)`,
	{ bind: { key, value } },
);
```

dannash100: Ahh yes way better ohk

passcod: Also maybe we should add a SQL function that wraps the SQL side of it to avoid writing that complicated nullif coalesce etc thing more than once.

dannash100:
@passcod After adding a migration for the get and set session functions it seems like determinism fails trying to create a change log for creation of SequalizeMeta entry which has none of the fields specified in the insert like created_at etc

This is the bit pretty sure 🤔 
```
  // remove changelog trigger before migrations
  // except from SequelizeMeta, so that migrations are recorded in the changelog
```

passcod:
Oh right. We should instead exclude sequelizemeta and write manually to the log on migrate, I think.

passcod (approved):
very cool, love that we now have a full new well-designed API for this kinda thing

PR #7298, opened 2025-03-04
feat: EPI-1124: Ability to manually enter the price of an additional invoice product

huytranbes:
### Changes

Ability to manually enter the price of an additional invoice product

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/Field/PriceField.jsx
+++ b/packages/web/app/components/Field/PriceField.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import { NumberInput } from './NumberField';
+
+export const PriceField = ({ field, ...props }) => {
+  const handleInput = e => {
+    const value = e.target.value;
+    // If value is negative just return empty
+    if (/^[−-]/.test(value)) {
+      e.target.value = '';
+      return;
+    }
+    // If the value is a decinal number, set it to 2 decimal places
+    if (value.includes('.')) {
+      const decimalPlaces = value.split('.')[1].length;
+      if (decimalPlaces > 2) {
+        e.target.value = parseFloat(value).toFixed(2);
+      }
+    }
+  };
+  return (
+    <NumberInput
+      value={field.value}
+      name={field.name}
+      onChange={field.onChange}
+      onInput={handleInput}
+      step="0.01"
+      min={0}
+      max={999999}
+      size="small"
+      {...props}
+    />
+  );
+};
--- a/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
+++ b/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
@@ -260,9 +260,9 @@ export const InvoiceItemRow = ({
       ...item,
       productName: value.productName,
       productCode: value.code,
-      productPrice: value.price,
       productDiscountable: value.discountable,
       product: { ...item.product, price: value.price },
+      ...(value.price !== null && { productPrice: value.price }),
     });
   };
 
--- a/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
+++ b/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import { Box } from '@material-ui/core';
 import { TranslatedText } from '../../Translation';
--- a/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
+++ b/packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
@@ -119,7 +119,8 @@ export const InvoiceItemRow = ({
   const hidePriceInput =
     item?.product?.price ||
     item?.product?.price === 0 ||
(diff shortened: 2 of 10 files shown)


packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
```diff
@@ -115,6 +117,7 @@ export const InvoiceItemRow = ({
       casing: 'lower',
     },
   );
+  const hidePriceInput = item.product?.price || item.product?.price === 0;
```
tcaiger, line 120: I wonder if this would be better to be `item.product?.price !== undefined`. I'll let you decide

packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
```diff
@@ -256,9 +260,17 @@ export const InvoiceItemRow = ({
...
+      product: { ...item.product, price: value.price },
     });
   };
 
+  useEffect(() => {
+    formArrayMethods.replace(index, {
+      ...item,
+      productPrice: defaultItem?.productPrice,
+    });
+  }, []);
```
tcaiger, line 272:
```suggestion
  }, [defaultItem?.productPrice]);
```
Should this reset everytime the default item changes?

packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
```diff
@@ -345,17 +357,34 @@ export const InvoiceItemRow = ({
...
+                <Field
+                  name={`invoiceItems.${index}.productPrice`}
+                  component={NumberField}
+                  min={0}
+                  max={999999}
+                  onInput={validateDecimalPlaces}
+                  size="small"
+                  required
+                  style={{ width: '100%' }}
+                />
```
tcaiger, line 386:
```suggestion

const PriceField = ({ field, ...props }) => {
  const handleChange = e => {
    let { value } = e.target;
    // If value is empty string and negative sign just return
    if (value === '' || /^[−-]/.test(value)) {
      field.onChange(value);
      return;
    }

    // IF the value is a number, set it to 2 decimal places
    let num = parseFloat(value);
    if (!isNaN(num)) {
      field.onChange(num.toFixed(2));
    }
  };

  return <NumberField value={field.value} onInput={handleChange} step="0.01" {...props} />;
};

                <Field
                  name={`invoiceItems.${index}.productPrice`}
                  component={PriceField}
                  min={0}
                  max={999999}
                  onInput={validateDecimalPlaces}
                  size="small"
                  required
                  style={{ width: '100%' }}
                />
```

It might be better to make a custom field time that extends Number field. Something like this (I haven't tested this code) because I think that setting e.target.value could have unexpected consequences.

tcaiger (changes requested):
Nice work - I left a couple of suggestions. I think that a PriceField would make the intention of the input much clearer and could be useful in other places in the Invoicing section of Tamanu

packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
huytranbes: No. The data from back-end actually returns null

packages/web/app/components/Invoice/EditInvoiceModal/InvoiceItem.jsx
huytranbes: I've tried this, but `field.onChange` didn't directly prevent typing invalid number when input
I've implement the `PriceField` component btw

packages/web/app/components/Field/PriceField.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { NumberInput } from './NumberField';
+
+export const PriceField = ({ field, ...props }) => {
```
tcaiger, line 4: Sorry I am not that experienced with Formik so I had to look up the exact syntax. It turns out you can either use `const { setFieldValue } = useFormikContext();` or onChange with the the event object `field.onChange({ target: { value: parseFloat(value).toFixed(2), name: field.name } });`. I think that setFieldValue is a bit cleaner eg.

```
export const PriceField = ({ field, ...props }) => {
  const { setFieldValue } = useFormikContext();

  const handleChange = e => {
    const value = e.target.value;

    // If value is negative just return empty
    if (/^[−-]/.test(value)) {
      setFieldValue(value);
      return;
    }
    // If the value is a decimal number, set it to 2 decimal places
    if (value.includes('.')) {
      const decimalPlaces = value.split('.')[1].length;
      if (decimalPlaces > 2) {
        setFieldValue(field.name, parseFloat(value).toFixed(2));
      }
    }
  };

  return (
    <NumberInput
      {...props}
      value={field.value}
      name={field.name}
      onChange={handleChange}
      step="0.01"
      min={0}
      max={999999}
      size="small"
    />
  );
};

tcaiger:
Also it looks like there are some tests failing so you will need to run the `dbt-generate-docs` script to get the tests to pass.

tcaiger (approved):

packages/web/app/components/Field/PriceField.jsx
huytranbes: I'm sorry, this didn't work either. Seems like Formik's method can not directly interfere with the DOM to prevent invalid input

PR #7306, opened 2025-03-05
feat(audit): NASS-1599: Add migration audit log

dannash100:
### Changes
- Its too different for changes trigger so now theres a seperate flow for migrations.
- We have to check if table exists because what if you have migrated down beyond the deletion of table 😱 noticed this in some tests that do up and then down and check some things
- Fixes the issues in underlying pr

### Deploys

### Remember to...

- ...write or update tests


passcod (approved):
✔️ with sync tick

PR #7310, opened 2025-03-06
fix(cd): Bestool walg thing

NavarroEmilioLuis:
### Changes

Checking if it's possible this unblocks the toolbox step that is failing everywhere

### Deploys


NavarroEmilioLuis:
Deleting, this should be the proper fix: https://github.com/beyondessential/tamanu/pull/7311

PR #7317, opened 2025-03-06
feat(programRegistry): SAV-090: Reference data translations

tcaiger:
### Changes

- Set up program registry models for translations on the importer module
- Add program registry models to reference type list in constants so that suggesters include them in translations
- Update all references to program registry models in the front end templates.

Note that I didn't add translations in the template for the AddConditionForm modal as that modal is about to be deleted in an upcoming PR and there is some custom behaviour in there that would require a custom translation solution.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/app/admin/importer/importRows.js
+++ b/packages/central-server/app/admin/importer/importRows.js
@@ -66,15 +66,14 @@ export async function importRows(
   validationContext = {},
 ) {
   const stats = { ...previousStats };
-  console.log('sheetName', sheetName);
 
-  console.log('Importing rows to database', { count: rows.length });
+  log.debug('Importing rows to database', { count: rows.length });
   if (rows.length === 0) {
-    console.log('Nothing to do, skipping');
+    log.debug('Nothing to do, skipping');
     return stats;
   }
 
-  console.log('Building reverse lookup table');
+  log.debug('Building reverse lookup table');
   const lookup = new Map();
   for (const {
     model,
@@ -86,7 +85,7 @@ export async function importRows(
     if (name) lookup.set(`kind.${kind}-name.${name.toLowerCase()}`, id);
   }
 
-  console.log('Resolving foreign keys', { rows: rows.length });
+  log.debug('Resolving foreign keys', { rows: rows.length });
   const resolvedRows = [];
   for (const { model, sheetRow, values } of rows) {
     try {
@@ -151,11 +150,11 @@ export async function importRows(
   }
 
   if (resolvedRows.length === 0) {
-    console.log('Nothing left, skipping');
+    log.debug('Nothing left, skipping');
     return stats;
   }
 
-  console.log('Validating data', { rows: resolvedRows.length });
+  log.debug('Validating data', { rows: resolvedRows.length });
   const validRows = [];
   for (const { model, sheetRow, values } of resolvedRows) {
     try {
@@ -204,7 +203,7 @@ export async function importRows(
   }
 
   if (validRows.length === 0) {
-    console.log('Nothing left, skipping');
+    log.debug('Nothing left, skipping');
     return stats;
   }
 
@@ -215,8 +214,7 @@ export async function importRows(
   };
   await validateTableRows(models, validRows, pushErrorFn);
 
-  console.log('Upserting database rows', { rows: validRows.length });
-  console.log('SKIP EXISTING: ', skipExisting);
+  log.debug('Upserting database rows', { rows: validRows.length });
   const translationRecordsForSheet = [];
   for (const { model, sheetRow, values } of validRows) {
     const Model = models[model];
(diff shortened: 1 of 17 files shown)


packages/central-server/app/admin/importer/importRows.js
```diff
@@ -115,15 +115,16 @@ export async function importRows(
                   })
                 : await models[fkSchema.model].count({ where: { id: fkFieldValue } })) > 0;
 
-            const idByRemoteName = (fkSchema.model === 'ReferenceData'
```
tcaiger, line 118: this file is just formatting changes. Sorry about that.

packages/central-server/app/admin/importer/importerEndpoint.js
```diff
@@ -5,31 +5,39 @@ import { singularize } from 'inflection';
...
+export function normaliseSheetName(name, modelName) {
   const norm = camelCase(
     lowerCase(name)
       .split(/\s+/)
-      .map(word => singularize(word))
+      .map((word) => singularize(word))
       .join(' '),
   );
 
-  // singularize transforms 'reference data' to 'reference datum', which is not what we want
```
tcaiger, line 23: I refactored these conditionals into a dictionary before adding the new options

packages/central-server/app/admin/programImporter/programImporter.js
```diff
@@ -21,7 +21,7 @@ export async function programImporter({
   checkPermission('create', 'Program');
   checkPermission('write', 'Program');
 
-  const createContext = sheetName => ({
+  const createContext = (sheetName) => ({
```
tcaiger, line 24: this file is just auto formatting too

packages/web/app/views/programRegistry/ConditionSection.jsx
```diff
@@ -98,6 +96,21 @@ export const ConditionSection = ({ patientProgramRegistration, programRegistryCo
...
 
+  const translatedData = conditions?.data?.map(condition => {
+    const { programRegistryCondition } = condition;
+    const { id, name } = programRegistryCondition;
+    const stringId = getReferenceDataStringId(id, 'programRegistryCondition');
+
+    const translatedName = getTranslation(
+      getReferenceDataStringId(id, 'programRegistryCondition'),
+      name,
+    );
```
NavarroEmilioLuis, line 107: Probably we want to use `stringId` in line 105 instead? Or remove that altogether

packages/web/app/views/programRegistry/ConditionSection.jsx
```diff
@@ -108,36 +121,32 @@ export const ConditionSection = ({ patientProgramRegistration, programRegistryCo
...
-              visible={x.programRegistryCondition?.name?.length > 30}
+      {sortedData.map(condition => (
+        <ConditionContainer key={condition.id}>
+          <ConditionalTooltip
+            title={condition.translatedName}
+            visible={condition.translatedName?.length > 30}
+          >
+            <ClippedConditionName>{condition.translatedName}</ClippedConditionName>
+          </ConditionalTooltip>
+          <ConditionalTooltip title="Patient must be active" visible={isRemoved}>
```
NavarroEmilioLuis, line 132: This is out of scope, but we probably need to schedule a meeting to say something like: tooltips need some sort of refactor to avoid clogging components with translations (although that's already a thing). Anyway, this bit wouldn't be translated

packages/web/app/views/programRegistry/DeleteProgramRegistryFormModal.jsx
```diff
@@ -54,7 +59,13 @@ export const DeleteProgramRegistryFormModal = ({ patientProgramRegistration, onC
...
         <p className="header">Confirm patient registry deletion</p>
         <p className="desc">
-          {`Are you sure you would like to delete the patient from the ${patientProgramRegistration?.programRegistry?.name}? This will delete associated patient registry records. This action is irreversible.`}
+          Are you sure you would like to delete the patient from the{' '}
+          <TranslatedReferenceData
+            fallback={patientProgramRegistration?.programRegistry?.name}
+            value={patientProgramRegistration?.programRegistry?.id}
+            category="programRegistry"
+          />
+          ? This will delete associated patient registry records. This action is irreversible.
```
NavarroEmilioLuis, line 68: I definitely don't love how this ends up looking in the code (having to add spaces like `{' '}` looks clunky!). The broader question is: this modal content won't be translated, and I think it's kinda unmanageable that we set this to be three different translations (once being the reference data) - languages won't map 1:1 - so seconding that notion that we might need to tech design a way to cover for these translation strings mixtures. If that makes sense?

packages/central-server/app/admin/importer/importerEndpoint.js
```diff
@@ -5,31 +5,43 @@ import { singularize } from 'inflection';
...
-  if (norm === 'referenceDatumRelation') return 'referenceDataRelation';
-
-  if (norm === 'vaccineSchedule') return 'scheduledVaccine';
-  if (norm === 'procedure') return 'procedureType';
-
-  // This is needed to handle the way we are exporting that data
```
NavarroEmilioLuis, line 29: Makes me wonder about how the exporter is affected by these changes? I would think that it would include the program registry tables on that output which wouldn't be right? (messy ref data)

NavarroEmilioLuis (changes requested):
Welp, some change requests, some questions and some ramblings. I think we could move on with this one but have to flag that we're starting to apply a pattern that doesn't really cover translations to a full extent. While we could implement something here, we might need a bigger effort to pull the whole thing off

packages/central-server/app/admin/programImporter/importProgramRegistry.js
```diff
@@ -11,10 +11,8 @@ import { readTwoModelTypeSheet } from './readMetadata';
 
 function readProgramRegistryData(workbook) {
   log.debug('Reading Registry data');
-  const {
```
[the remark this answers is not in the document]
tcaiger: No this change was done by prettier. I do have that turned on to auto run on save which is why this change came through. I thought that we adopt prettier across the board.

packages/web/app/views/programRegistry/ConditionSection.jsx
tcaiger: woops thanks

packages/web/app/views/programRegistry/DeleteProgramRegistryFormModal.jsx
tcaiger: I think that there is already a solution for this scenario which is "replacements". So it could look like this once complete.

```js
<TranslatedText
            stringId="programRegistry.deleteModal.warning"
            fallback="Are you sure you would like to delete the patient from the :programRegistry? This will delete associated patient registry records. This action is irreversible."
            replacements={{
              programRegistry: (
                <TranslatedReferenceData
                  fallback={patientProgramRegistration?.programRegistry?.name}
                  value={patientProgramRegistration?.programRegistry?.id}
                  category="programRegistry"
                />
              ),
            }}
          />
```

I was planning on doing the static string translations at once in this ticket https://linear.app/bes/issue/SAV-911/program-registry-static-string-translations

packages/central-server/app/admin/importer/importerEndpoint.js
tcaiger: The export of the program registry tables is not affected by these changes. It already exports kind of funny because it exports with the Programs data and is included in the Registry and RegistryCondition sheet names but everyone expects that now so we can't change it.

NavarroEmilioLuis (approved):
LGTM

PR #7253, opened 2025-02-21
feat(timesync): NASS-1474: Store time offset between central and facility servers

dannash100:
### Changes

Very wip just was confirming what the actual requirements where.
- Use timesync to store an offset +/- in ms of time between facility and central server to be used in auditing.

one question tho do yall rekon we should do 
```
timesync.destroy()
```
On SIGTERM or not important 🤔 

### Deploys

- [x] **Deploy to Tamanu Internal** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

PR #7318, opened 2025-03-06
feat: Audit logs (EPIC)

passcod:
### Changes

This is a common branch to merge Audit Project PRs into.

### Deploys

- [x] **eploy to Tamanu Internal**  %fakedata=1

### Status tracking

- [ ] Change Logs
  - [ ] ~~Use the last updated by column name introduced by nassau~~ (obsoleted by tech design)
  - [x] Make sure we're defaulting to the system user (#7268)
  - [x] Make sure the system user exists (#7268)
  - [x] Indexes (#7239)
  - [x] ~~Change primary key to bigserial for insertion speed and btree indexing (#7239)~~
  - [x] Use synced time offset (see #7253) for the logged_at timestamp
  - [ ] Sync it
    - [x] figure out what happens on sync with the bigserial, do we need a device field as part of the pkey
    - [x] figure out how to sync it... do we have a second update_at_sync_tick just for syncing the changelogs...
  - [ ] Tests (for insertion)
  - [ ] Add relevant fields from access log (is_mobile, version)
- [ ] Mobile
  - [ ] Doesn't record its own logs but instead that happens on central sync
  - [ ] Override the last updated by session var
- [ ] Access Logs
	- [x] Schema
	- [ ] Insertions
	- [ ] Use the synced time offset
	- [ ] Fix schema id default (**stop using `uuid_generate_v4`!**)
	- [ ] Indexes
	- [ ] Tests
- [ ] Manual release step upgrade command which copies all the current state of the tables into `logs.changes`. (The framework for this is coming in the Week 12-13 seconding sprint)


passcod:
On occasien need to bump timesimp in central-server too to match facility-server — currently there's two versions specified, though they're compatible and should resolve

Harnesh787 (approved):

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -581,6 +581,7 @@ export class CentralSyncManager {
     try {
       // commit the changes to the db
       const persistedAtSyncTick = await sequelize.transaction(async () => {
+        await sequelize.setTransactionVar(AUDIT_PAUSE_KEY, true);
```
edmofro, line 584: This is random enough amongst the rest of the code in her to deserve a comment imo, adds coupling to a completely different system that you haven't otherwise had to think about while reading through CentralSyncManager

packages/database/src/migrations/1739970132204-ensureSystemUserPresent.ts
```diff
@@ -0,0 +1,20 @@
...
+export async function up(query: QueryInterface) {
+  // The previous iteration of that migration, 1685403132663-systemUser, was too
+  // clever and created situations where the system user isn't present in real
+  // deployments and also in tests. As we're increasingly relying on the system
+  // user existing with a Nil UUID, this migration is here to make certain.
+  await query.sequelize.query(`
+    INSERT INTO "users"
+    (id, email, display_name, role)
+    VALUES
+    (uuid_nil(), 'system', 'System', 'system')
```
edmofro, line 12: Didn't know about the `uuid_nil` function, there's all sorts in postgres huh

edmofro (approved):
Pre-approving, I would like one more comment in there but otherwise sync changes look good to me

packages/central-server/app/sync/CentralSyncManager.js
tcodling: @passcod Are you able to add a comment in here and then we can go ahead with the merge?

edmofro: @tcodling can you please add the comment? I know @passcod is trying to get us as much assurance on sync testing of auditability today as possible, so be great to leave them focused on that

tcodling: Ill try my best just I am totally out of the loop what is going on here 😆

edmofro: Something like this, although I don't know if the comment about mobile is true @dannash100??

```suggestion
        // Don't produce change logs for changes made through sync
        // Auditable change logs are generated on facility servers and will sync in to central
        // For mobile, we generate change logs on central through a separate mechanism
        await sequelize.setTransactionVar(AUDIT_PAUSE_KEY, true);
```

dannash100: Yeah thats true re mobile just not currently implemented

dannash100: Comment looks good to me @tcodling

PR #7327, opened 2025-03-10
fix(translations): Allow translated text import to handle a number like in repeats

dannash100:
### Changes
Just noticed this when trying to import scraped translations It would fail on `medication.property.repeats.0` - `medication.property.repeats.12` 

See spreadsheet that now imports correctly
[problematic-translations.xlsx](https://github.com/user-attachments/files/19155037/problematic-translations.xlsx)

XLSX package must have read them as numbers and so they would error on trim, cooerce into string works great

### Deploys

### Remember to...

- ...write or update tests


tcodling (approved):
👂 pic

PR #7333, opened 2025-03-10
feat(programRegistry): SAV-911: Program registry string translations

tcaiger:
### Changes

- Setup translation components for missing translations in program registries

### Deploys

### Remember to...

- ...write or update tests


packages/web/app/views/programRegistry/ActivatePatientProgramRegistry.jsx
```diff
@@ -236,29 +235,13 @@ export const ActivatePatientProgramRegistry = ({ onClose, patientProgramRegistra
...
-            .required()
-            .translatedLabel(
-              <TranslatedText
-                stringId="patientProgramRegistry.registeringFacility.label"
-                fallback="Registering facility"
-              />,
-            ),
+            .required(<TranslatedText stringId="general.required" fallback="Required" />),
+          clinicianId: foreignKey().required(
+            <TranslatedText stringId="general.required" fallback="Required" />,
```
tcaiger, line 240: Direction from Flic was to use the generic required message everywhere

packages/web/app/views/programRegistry/ConditionSection.jsx
```diff
@@ -112,10 +112,26 @@ export const ConditionSection = ({ patientProgramRegistration, programRegistryCo
   return (
     <Container>
       <HeadingContainer>
-        <Heading5>Related conditions</Heading5>
-        <ConditionalTooltip title="Patient must be active" visible={isRemoved}>
+        <Heading5>
+          <TranslatedText
+            stringId="patientProgramRegistry.general.relatedConditions.label"
```
tcaiger, line 117: I tried to re-use as many string ids as possible and prefixed the re-usable ones in the `patientProgramRegistry.general` key

packages/web/app/views/programRegistry/DeleteProgramRegistryFormModal.jsx
```diff
@@ -55,17 +50,31 @@ export const DeleteProgramRegistryFormModal = ({ patientProgramRegistration, onC
...
+        </p>
         <p className="desc">
-          Are you sure you would like to delete the patient from the{' '}
-          <TranslatedReferenceData
-            fallback={patientProgramRegistration?.programRegistry?.name}
-            value={patientProgramRegistration?.programRegistry?.id}
-            category="programRegistry"
+          <TranslatedText
+            stringId="patientProgramRegistry.modal.deleteRegistry.description"
+            fallback="Are you sure you would like to delete the patient from the :name? This will delete associated patient registry records. This action is irreversible."
```
NavarroEmilioLuis, line 75: Nitpicking but possibly

```suggestion
            fallback="Are you sure you would like to delete the patient from the :programRegistry? This will delete associated patient registry records. This action is irreversible."
```

packages/web/app/views/programRegistry/PatientProgramRegistryView.jsx
```diff
@@ -62,7 +63,15 @@ export const PatientProgramRegistryView = () => {
   }
 
   if (isError) {
-    return <p>Program registry &apos;{title || 'Unknown'}&apos; not found.</p>;
+    return (
+      <p>
+        <TranslatedText
+          stringId="patientProgramRegistry.registeringNotFoundMessage"
+          fallback="Program registry ':title' not found."
+          replacements={{ title: title || 'Unknown' }}
```
NavarroEmilioLuis, line 71: A bit more involved replacement... I would say we need to double check whether `title` is already properly translated. As to the fallback, either using it in the title itself and translating it there, or just translating it here. Probably the first one better.

packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
```diff
@@ -63,7 +63,7 @@ export const ProgramRegistryTable = ({ searchParameters }) => {
         key: 'homeVillage',
         title: (
           <TranslatedText
-            stringId="programRegistry.table.column.homeVillage"
+            stringId="patientProgramRegistry.homeVillage.label"
```
NavarroEmilioLuis, line 66: This one is interesting cause it's not really a patient program registry column. It's read from reference data via the patients table.

However, I think if we wanted to make a distinction with whatever we call that field, we might want to specify this being a `registryTable.column`?

packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
```diff
@@ -212,7 +217,12 @@ export const ProgramRegistryTable = ({ searchParameters }) => {
         refreshCount={refreshCount}
         endpoint={`programRegistry/${params.programRegistryId}/registrations`}
         columns={columns}
-        noDataMessage="No Program registry found"
+        noDataMessage={
+          <TranslatedText
+            stringId="patientProgramRegistry.registrytable.column.noDataMessage"
+            fallback="No Program registry found"
```
NavarroEmilioLuis, line 223: Nitpicking

```suggestion
            fallback="No program registry found"
```

packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
```diff
@@ -212,7 +217,12 @@ export const ProgramRegistryTable = ({ searchParameters }) => {
         refreshCount={refreshCount}
         endpoint={`programRegistry/${params.programRegistryId}/registrations`}
         columns={columns}
-        noDataMessage="No Program registry found"
+        noDataMessage={
+          <TranslatedText
+            stringId="patientProgramRegistry.registrytable.column.noDataMessage"
```
NavarroEmilioLuis, line 222: Nitpicking

```suggestion
            stringId="patientProgramRegistry.registrytable.noDataMessage"
```

NavarroEmilioLuis (commented):
Nothing jumped out of me, but lots of comments we might want to address!

packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
NavarroEmilioLuis: Actually 🤔 I wonder if we want `table` instead of `registrytable` in all of these. I also think it's weird it's not camelCase?

packages/web/app/views/programRegistry/ActivatePatientProgramRegistry.jsx
tcaiger: yes I agree with you - good suggestion

packages/web/app/views/programRegistry/DeleteProgramRegistryFormModal.jsx
```diff
@@ -55,17 +50,31 @@ export const DeleteProgramRegistryFormModal = ({ patientProgramRegistration, onC
...
+            replacements={{ name: patientProgramRegistration?.programRegistry?.name }}
```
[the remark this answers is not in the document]
tcaiger: yep that's right - good catch

packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
tcaiger: It's not a database table column at all because it is just the title of the front end table which is why I called it a label so I think it doesn't make sense to give it a table string id

packages/web/app/views/programRegistry/AddConditionFormModal.jsx
```diff
@@ -35,6 +36,7 @@ export const AddConditionFormModal = ({
 }) => {
   const api = useApi();
   const queryClient = useQueryClient();
+  const { getTranslation } = useTranslation();
```
NavarroEmilioLuis, line 39: Uhhh for some reason the import is missing `useTranslation`

NavarroEmilioLuis (approved):
LGTM

PR #7341, opened 2025-03-11
fix(pm2): SAV-918: Correctly read config file from pm2.config.cjs

rohan-bes:
### Changes

By setting `NODE_CONFIG_DIR` we can ensure that the config is read from the correct directory.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/pm2.config.cjs
+++ b/packages/central-server/pm2.config.cjs
@@ -1,15 +1,20 @@
 const os = require('node:os');
+
+const cwd = '.'; // IMPORTANT: Leave this as-is, for production build
+
+process.env.NODE_CONFIG_DIR = cwd + '/config/';
 const config = require('config');
 
-const totalMemoryMB = Math.round(os.totalmem() / (1024**2));
+const totalMemoryMB = Math.round(os.totalmem() / 1024 ** 2);
 const memory = process.env.TAMANU_MEMORY_ALLOCATION || (totalMemoryMB * 0.6).toFixed(0);
 
 const availableThreads = os.availableParallelism();
 const minimumApiScale = totalMemoryMB > 3000 ? 2 : 1;
 const maximumApiScale = 4; // more requires custom caddy config
-const defaultApiScale = Math.min(maximumApiScale, Math.max(minimumApiScale, Math.floor(availableThreads / 2)));
-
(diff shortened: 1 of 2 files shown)


packages/central-server/pm2.config.cjs
```diff
@@ -1,15 +1,20 @@
 const os = require('node:os');
+
+const cwd = '.'; // IMPORTANT: Leave this as-is, for production build
```
rohan-bes, line 3: @passcod I haven't been able to work out exactly how this value is changed on production. It's automatically done by the `bestool`?

packages/central-server/pm2.config.cjs
```diff
@@ -1,15 +1,20 @@
...
+
+process.env.NODE_CONFIG_DIR = cwd + '/config/';
```
passcod, line 5: This will break production.

packages/central-server/pm2.config.cjs
passcod: https://github.com/beyondessential/tamanu/blob/main/.github/workflows/cd-package-servers.yml#L44-L52

packages/central-server/pm2.config.cjs
passcod: https://github.com/beyondessential/tamanu/blob/main/Dockerfile#L59-L60

passcod (approved):
This will work, but needs a comment both in this file and [on the Dockerfile](https://github.com/beyondessential/tamanu/blob/main/Dockerfile#L59-L60) to explain why their usages don't conflict.

rohan-bes:
> This will work, but needs a comment both in this file and [on the Dockerfile](https://github.com/beyondessential/tamanu/blob/main/Dockerfile#L59-L60) to explain why their usages don't conflict.

@passcod Do we still need the line to set `NODE_CONFIG_DIR` in the Dockerfile? I guess if we're running things other than pm2, like migrations?

passcod:
If that line isn't there, the containers will not work period. It's how the containers get their configuration.

passcod:
Specifically the one in the dockerfile has `/config:` prefixed, this is the critical bit alongside the normal packages/whatever-server/config value.

passcod:
And finally there's no pm2 in containers.

PR #7345, opened 2025-03-11
refactor: constantise all the LocalSystemFact keys

passcod:
### Changes

- Gathers all the LocalSystemFact keys into a single file in constants, and replaces all usage.
- At the Typescript level, enforces that `LocalSystemFact.{get,set,incrementValue}` have to be called with known fact keys
- Renames `CURRENT_SYNC_TIME` to `CURRENT_SYNC_TICK` to reflect modern terminology -- as all usage is constantised, this was trivial

### Deploys

### Remember to...

- ...write or update tests


packages/database/src/sync/constants.ts
```diff
@@ -1,13 +1,15 @@
+export {
```
dannash100, line 1: Might not need to repeat exports here as not used 🤔

dannash100 (approved):
Looks great - just a tiny thing

packages/database/src/sync/constants.ts
passcod: Okay I had to convince myself of that (that I hadn't left any old import) but this is right, so will remove.

```
rg 'import[\n\s{]+FACT_[\w,\n\s}]+from' --multiline
```

edmofro (approved):
Good call

PR #7346, opened 2025-03-12
deps: upgrade pdfjs

passcod:
### Changes

Just a dependency bump.


NavarroEmilioLuis (approved):
LGTM, doesn't seem like the API changed so it should be fine

passcod:
Apparently this did break things, I'll leave it to the next maintenance deps sprint.

PR #7351, opened 2025-03-12
feat(programRegistry): SAV-912: Mobile translations pt 1

NavarroEmilioLuis:
### Changes

Splitting the enum bit into a separate PR and maybe then will have to do a second pass

### Deploys

### DIFF
--- a/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
+++ b/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
@@ -54,7 +54,7 @@ const PatientProgramRegistrySummary_ = ({ selectedPatient }): ReactElement => {
       <Row>
         <SectionHeader h1 fontSize={14} fontWeight={500} color={theme.colors.TEXT_SUPER_DARK}>
           <TranslatedText
-            stringId="TBD"
+            stringId="general.localisedField.programRegistry.label"
             fallback="Program registry"
           />
         </SectionHeader>
--- a/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
+++ b/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
@@ -54,7 +54,7 @@ const PatientProgramRegistrySummary_ = ({ selectedPatient }): ReactElement => {
       <Row>
         <SectionHeader h1 fontSize={14} fontWeight={500} color={theme.colors.TEXT_SUPER_DARK}>
           <TranslatedText
-            stringId="general.localisedField.programRegistry.label"
+            stringId="patientProgramRegistry.programRegistry.label"
             fallback="Program registry"
           />
         </SectionHeader>
--- a/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
+++ b/packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
@@ -16,6 +16,7 @@ import { useBackendEffect } from '~/ui/hooks/index';
 import { LoadingScreen } from '~/ui/components/LoadingScreen';
 import { ErrorScreen } from '~/ui/components/ErrorScreen';
 import { useAuth } from '~/ui/contexts/AuthContext';
+import { TranslatedText } from '~/ui/components/Translations/TranslatedText';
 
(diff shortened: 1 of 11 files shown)


packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
```diff
@@ -52,7 +53,10 @@ const PatientProgramRegistrySummary_ = ({ selectedPatient }): ReactElement => {
     <StyledView margin={20} borderRadius={5}>
       <Row>
         <SectionHeader h1 fontSize={14} fontWeight={500} color={theme.colors.TEXT_SUPER_DARK}>
-          Program registry
+          <TranslatedText
+            stringId="general.localisedField.programRegistry.label"
```
tcaiger, line 57: I think that the localised field key is just for legacy localised strings. For these I used `patientProgramRegistry.general.programRegistry.label` but open to suggestions for a better one

packages/mobile/App/ui/navigation/screens/patientProgramRegistration/form/PatientProgramRegistrationDetailsForm.tsx
```diff
@@ -158,20 +179,25 @@ export const PatientProgramRegistrationDetailsForm = ({ navigation, route }: Bas
...
                     name="registeringFacilityId"
                   />
                 </StyledView>
                 <StyledView marginLeft={20} marginRight={20}>
                   <LocalisedField
-                    label={<TranslatedText stringId="general.status.label" fallback="Status" />}
+                    label={
+                      <TranslatedText
+                        stringId="patientProgramRegistry.clinicalStatus.label"
+                        fallback="Status"
```
tcaiger, line 193: hmmm this is an interesting one. On desktop I did `<TranslatedText stringId="general.status.label" fallback="Status" />` for these. @dannash100 What would you recommend here?

packages/mobile/App/ui/navigation/stacks/PatientSummaryStack.tsx
```diff
@@ -31,7 +32,12 @@ const PatientSummary = ({
   return (
     <ErrorBoundary>
       <StackHeader
-        title="Program registries"
+        title={
+          <TranslatedText
+            stringId="patient.action.viewProgramRegistries"
```
tcaiger, line 37: My understanding of `action` was that it was for buttons or form actions. Maybe this should be more list `patientProgramRegistry.tab.title`?

tcaiger (changes requested):
Looks great overall. I left a few suggestions - the localised field one is the only one that I feel more strongly about

packages/mobile/App/ui/navigation/screens/patientProgramRegistration/form/PatientProgramRegistrationDetailsForm.tsx
NavarroEmilioLuis: Yeah that's what I did initially, but it's not a general status. Might be the same copy, but something else entirely

packages/mobile/App/ui/navigation/stacks/PatientSummaryStack.tsx
NavarroEmilioLuis: Might, though it's also not a tab, maybe a header?

packages/mobile/App/ui/navigation/screens/patientProgramRegistration/PatientProgramRegistrySummary.tsx
NavarroEmilioLuis: Oh okay yeah I'll do that one

packages/mobile/App/ui/navigation/stacks/PatientSummaryStack.tsx
tcaiger: yeah header sounds good

packages/mobile/App/ui/navigation/screens/patientProgramRegistration/form/PatientProgramRegistrationDetailsForm.tsx
tcaiger: yeah I think you are right. I will change mine to match yours

rohan-bes (approved):
Looks good :+1:

NavarroEmilioLuis:
Just FYI, I asked Rohan if he could approve this PR cause Tom isn't in today - we did discuss and the changes should be already in place so only bypassing to speed things up 🙏 should anything else come up, will resolve in yet another PR

PR #7352, opened 2025-03-12
tweak(logs): omit timestamps when running as a systemd service

passcod:
### Changes

When running as a systemd service (e.g. in Linux), the logs have two timestamps:

```
Mar 13 07:48:54 facility.example.tamanu.app systemd-tamanu-facility[71159]: 2025-03-12T22:48:54.953Z info: Sync:
```

This PR adds a `log.timeless` config boolean, and also auto-detects systemd and switches on timeless mode accordingly.

### DIFF
--- a/packages/central-server/config/default.json5
+++ b/packages/central-server/config/default.json5
@@ -42,7 +42,8 @@
   "log": {
     "path": "",
     "consoleLevel": "http",
-    "color": true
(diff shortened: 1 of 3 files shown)


NavarroEmilioLuis (approved):
Awesome!

anilvkaran (approved):

PR #7353, opened 2025-03-12
fix: fix session restore

tcaiger:
### Changes
Fix the restore session functionality that is used to keep a user logged in during local dev. The issue was that the `ability` method was unintentionally getting set from persisted redux state rather than from the session restore function.

Example of bug before fix:

Screenshot 2025-03-13 at 12 23 41 PM

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/api-client/src/TamanuApi.js
+++ b/packages/api-client/src/TamanuApi.js
@@ -80,7 +80,7 @@ export class TamanuApi {
     this.setToken(token);
 
     const { user, ability } = await this.fetchUserData(permissions);
-    return { user, token, localisation, server, availableFacilities, ability, role };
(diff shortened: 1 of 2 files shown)


packages/api-client/src/TamanuApi.js
```diff
@@ -80,7 +80,7 @@ export class TamanuApi {
     this.setToken(token);
 
     const { user, ability } = await this.fetchUserData(permissions);
-    return { user, token, localisation, server, availableFacilities, ability, role };
+    return { user, token, localisation, server, availableFacilities, ability, role, permissions };
```
tcaiger, line 83: This is a related bug but was not the main issue. Downstream we use this permissions list to build the user ability.

passcod (approved):

PR #7355, opened 2025-03-13
fix(mobile): Tidy up mobile MultiSelectModalField component

NavarroEmilioLuis:
### Changes

So this had a couple of issues and I decided to fix it up. It will also be needed to display a fully translatable section on the program registry form. However I'm targeting this to main cause it's also pretty standalone and minor, and we can always update the branch when we are doing the test there.

### DIFF
--- a/packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
+++ b/packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
@@ -24,7 +24,7 @@ interface MultiSelectModalFieldProps {
   searchPlaceholder?: string;
 }
 
-const appendLabel = (items: OptionType[]) => {
+const extractLabel = (items: OptionType[]) => {
   return items.map(x => ` ${x.label}`).toString();
 };
 
@@ -47,7 +47,7 @@ export const MultiSelectModalField = ({
 
   const handleSaveCallback = (selectedItems: OptionType[]): void => {
     onChange(selectedItems);
-    setLabel(appendLabel(selectedItems));
+    setLabel(extractLabel(selectedItems));
   };
 
   const openModal = (): void =>
@@ -71,7 +71,7 @@ export const MultiSelectModalField = ({
       selectedValues.push(data);
     }
 
-    const updatedLabel = appendLabel(selectedValues);
+    const updatedLabel = extractLabel(selectedValues);
     setLabel(updatedLabel);
   }, [suggester]);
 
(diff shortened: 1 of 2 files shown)


packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
```diff
@@ -41,35 +45,42 @@ export const MultiSelectModalField = ({
...
       modalTitle,
       value,
       searchPlaceholder,
     });
-  const loadDefaultValues = async (values: string[]) => {
-    const _values: OptionType[] = [];
-    values.forEach(async x => {
+
+  // This function is not in use, however, it's meant to be used when
+  // initial values are set. Pay extra care to the value format it expects.
```
tcaiger, line 63: I'm a bit confused about this comment as this function is being used below right?

packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
```diff
@@ -41,35 +45,42 @@ export const MultiSelectModalField = ({
...
+      return;
+    }
+    const selectedValues: OptionType[] = [];
+    for (const x of values) {
       const data = await suggester.fetchCurrentOption(x);
-      _values.push(data);
-    });
-    const _label = appendLabel(_values);
-    setLabel(_values.length > 0 ? _label : setLabel(placeholder));
-  };
```
tcaiger, line 68: Nice clean up - much better after

packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
```diff
@@ -41,35 +45,42 @@ export const MultiSelectModalField = ({
...
+  }, [suggester]);
 
   useEffect(() => {
-    if (!suggester || !value) return;
-    loadDefaultValues(value);
+    if (!value) return;
+    loadInitialLabel(value);
+  // Disabling the linter because we only want to run this on mount
+  // if the component has an initial value and ignore value updates.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
tcaiger, line 83: thanks for the comment - it's quite an annoying lint rule as running on mount is one of the main use cases for useEffect!

tcaiger (approved):
Looks good - I just have a question about the comment - maybe that's meant to be deleted

packages/mobile/App/ui/components/MultiSelectModal/MultiSelectModalField.tsx
tcaiger: ok that sounds good to me

PR #7356, opened 2025-03-13
feat(testing): NASS-1618: Set up Playwright

flynncs:
### Changes

- Creates a new `e2e-tests` package
- Adds Playwright as a dependency
- Implements basic test run scripts
- POC test

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/e2e-tests/playwright.config.ts
+++ b/packages/e2e-tests/playwright.config.ts
@@ -39,15 +39,15 @@ export default defineConfig({
       use: { ...devices['Desktop Chrome'] },
     },
 
-    {
-      name: 'firefox',
-      use: { ...devices['Desktop Firefox'] },
-    },
+    // {
+    //   name: 'firefox',
+    //   use: { ...devices['Desktop Firefox'] },
+    // },
 
-    {
-      name: 'webkit',
-      use: { ...devices['Desktop Safari'] },
-    },
+    // {
+    //   name: 'webkit',
+    //   use: { ...devices['Desktop Safari'] },
+    // },
 
     /* Test against mobile viewports. */
     // {
--- a/packages/e2e-tests/playwright.config.ts
+++ b/packages/e2e-tests/playwright.config.ts
@@ -1,0 +1,97 @@
+import { defineConfig, devices } from '@playwright/test';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// import dotenv from 'dotenv';
+// import path from 'path';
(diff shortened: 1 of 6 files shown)


packages/e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,94 @@
...
+  testDir: './tests',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
```
edmofro, line 24: These will need revisiting for appropriateness in the next ticket

packages/e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
```
edmofro, line 50: We don't support firefox or safari

packages/e2e-tests/playwright.config.ts
```diff
@@ -0,0 +1,94 @@
...
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
```
edmofro, line 64: We do support edge

edmofro (changes requested):
Only required change would be to comment out firefox and safari, otherwise lgtm! I like that you've chosen `e2e-tests` as the package name, rather than `playwright`

packages/e2e-tests/playwright.config.ts
flynncs: Should we test against brand specific chromium browsers, i.e. Edge & Chrome, or leave it as just the generic chromium browser being tested?

edmofro: Just using chromium is fine, especially if it adds to the time taken for the suite to run. This one was optional :+1:

edmofro (approved):

PR #7363, opened 2025-03-13
doc: create SECURITY.md

passcod:
### Changes

Short and sweet is how those typically are. In fact, ours is on the long side!


dannash100 (approved):
Looks good to me! :shipit:

PR #7364, opened 2025-03-13
fix(lint): eslint config and migration template

passcod:
### Changes

This came from the audit epic but is better applied to main directly.

We deliberately exclude the migration template for mobile from linting... this also excludes the server one, and thus gets rid of the "delete this line" instruction.

Exclude the eslint no-unused-vars from all TS files, instead relying on the TS rule; this prevents invalid "errors" like this where eslint doesn't understand TS function signatures:

```typescript
interface Foo {
  bar: (qux: string) => Promise<void>;
//      ^^^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ qux isn't used
}
```

Move support for the `_underscorePrefixed` allowed unused vars to apply to all typescript, not just mobile, especially now that we have mobile merged into the workspace and want to start sharing code more effectively.

Removed mentions of the gone package `qr-tester`.

Fixes for the web package and its storybook files... that I'm not sure how they ever worked. I think something in the cascade of rules made it so the web package was considered both node and browser, when that wasn't really the intent.

Still, as we've used a couple node-ish things that are polyfilled by Vite, include those explicitly (this should discourage relying on more polyfills).

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -76,15 +76,6 @@ export default [
     // mobile rule exclusions (as warnings, until we fix them)
     files: [`packages/mobile/**/*.${TS_EXTS}`],
     rules: {
-      'no-unused-vars': 'off',
-      "@typescript-eslint/no-unused-vars": [
-        'error',
-        {
-          "argsIgnorePattern": "^_",
-          "varsIgnorePattern": "^_",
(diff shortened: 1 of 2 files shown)


dannash100 (approved):
🙏

PR #7367, opened 2025-03-13
refactor(audit): replace timesync implementation

passcod:
### Changes

Update: using the [timesimp](https://passcod.github.io/timesimp/js/classes/Timesimp.html) package instead.

<details>
<summary>Original PR description</summary>

Something kept bothering me about the timesync package, and so I read the paper it's based on and the implementation side-by-side... the package is wrong! The [paper](https://web.archive.org/web/20160310125700/http://mine-control.com/zack/timesync/timesync.html) specifies:

> 1. Client stamps **current local time** on a "time request" packet and sends to server
> 2. Upon receipt by server, server stamps **server-time** and returns
> 3. Upon receipt by client, client subtracts **current time** from **sent time** and divides by two to compute latency. It subtracts **current time** from **server time** to determine client-server time delta and adds in the half-latency to get the correct clock delta.
> 4. The first result should immediately be used to update the clock since it will get the local clock into at least the right ballpark (at least the right timezone!)
> 5. The client repeats steps 1 through 3 five or more times, pausing a few seconds each time. Other traffic may be allowed in the interim, but should be minimized for best results
> 6. The results of the packet receipts are accumulated and sorted in lowest-latency to highest-latency order. The median latency is determined by picking the mid-point sample from this ordered list.
> 7. All samples above approximately 1 standard-deviation from the median are discarded and the remaining samples are averaged using an arithmetic mean.

(emphasis mine)

Granted, this is a bit confusing, but 1. really should read:

> 1. Client stamps current local time **as sent-time** on a "time request" packet and sends to server

This is indeed (a simple version of) [how NTP works](https://en.wikipedia.org/wiki/Network_Time_Protocol#Clock_synchronization_algorithm). Notably, this excludes stamping the reception/transmit time, because at our level, there really isn't much of a difference or opportunity to do this. (At the NTP level, in a low-level programming model, the server can effectively stamp the received package _extremely early_, and then stamp it out _just before it goes to the network interface_, and thus the client can remove the NTP server processing time from the calculations. In Node.js, by the time _we_ get access to the incoming packet, _so much work_ has already gone into processing it by the Node runtime and event loop etc that it doesn't make sense to stamp the packet, only to immediately resend back into the event loop machinery, where _even more time_ will be spent that we can't account for.)

Compared to NTP, we also don't stamp the reference time, as we simply don't have that information (it's the last time that the server reference clock was updated).

Also, as Daniel originally found, the package is really old and hasn't kept up with advances in Node and JS runtimes, and was using polyfilled Promises, which we already know from past (and this) experience just doesn't play well with our codebase anymore (and also plain haven't been needed for years). Finally, it's written in a very old style of Javascript.

This refactor is less a rewrite of the timesync package, and more a write-from-scratch of the original algorithm. Additionally:
- it uses a simpler envelope (just JSON, instead of JSONAPI), and is flexible on the transport (so if in the future we want to do timesync over a websocket or QUIC connection, or to add authentication to the endpoint, we easily can).
- it provides `ctx.timestamp` throughout the application, so you can get adjusted dates using `ctx.timestamp.now()`.
- it has a detailed configuration in `settings` instead of the config file, and is written in such a way that it can be enabled and disabled at runtime.
- it has a read only mode, so it can be used in multi-process environments safely.

Finally, instead of using millisecond offsets, it uses and stores microsecond offsets, just to really future proof.

</details>

The sync codeowners bit is that this refactor moves the file out of the sync folder, and into services, where it really belongs.

Tests fail because this is part of a set of PRs that all come together; this [this earlier run](https://github.com/beyondessential/tamanu/actions/runs/13830763740/job/38694338909?pr=7344) which includes all the parts together.


packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+    // not enough for confidence
+    if (responses.length < 3) return;
+
+    responses.sort((a, b) => a.latency - b.latency);
+    const deltas = responses.map(({ delta }) => delta);
+
+    const medianIdx = Math.floor(deltas.length / 2);
+    const median = deltas[medianIdx]!;
+
+    const mean = deltas.reduce((acc, d) => acc + d, 0);
```
edmofro, line 224: Isn't this the sum?

packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+    const variance =
+      deltas.map((d) => Math.pow(d - mean, 2)).reduce((acc, d) => acc + d, 0) / (deltas.length - 1);
+    const std = Math.sqrt(variance);
+
+    const inliers = deltas.filter((d) => d <= median + std || d >= median - std);
+    const offset = Math.round(deltas.reduce((acc, d) => acc + d, 0) / inliers.length);
```
edmofro, line 230: You're getting the sum of all the deltas, but only dividing by the number of inliers. Seems weird?

packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    const medianIdx = Math.floor(deltas.length / 2);
+    const median = deltas[medianIdx]!;
+
+    const mean = deltas.reduce((acc, d) => acc + d, 0);
+    const variance =
+      deltas.map((d) => Math.pow(d - mean, 2)).reduce((acc, d) => acc + d, 0) / (deltas.length - 1);
+    const std = Math.sqrt(variance);
+
+    const inliers = deltas.filter((d) => d <= median + std || d >= median - std);
```
edmofro, line 229: I think this should be an &?
```suggestion
    const inliers = deltas.filter((d) => d <= median + std && d >= median - std);
```

FWIW I find the other order more of a "visual" representation
```suggestion
    const inliers = deltas.filter((d) => d >= median - std && d <= median + std);
```

packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+      );
+    }
+
+    return state;
+  }
+
+  /** Use init() instead. */
+  constructor(noAccess: Symbol, { models, settings, log, query }: CreateArgs) {
+    if (noAccess !== NO_CONSTRUCTOR_ACCESS) {
+      throw new Error('Do not construct TimesyncState using new, use init() instead');
```
edmofro, line 99:
```suggestion
      throw new Error('Do not construct Timesync using new, use init() instead');
```

edmofro (changes requested):
I'm pretty tired today so I might be thinking wacky, but I think I found a number of bugs? Perhaps tests are a good idea.

packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+ */
+export type QueryFn = (request: TimeRequest, timeoutMs: number) => Promise<TimeResponse | null>;
+
+export interface CreateArgs {
+  models: Models;
+  settings: ReadSettings;
+  log: Logger;
+
+  /** A callback that performs a query to a timesync server and returns the response.
+   * False to disable querying; in that case the offset will be reloaded at regular intervals.
```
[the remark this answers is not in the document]
passcod: from the database into the runtime

packages/database/src/services/timesync.ts
```diff
@@ -0,0 +1,238 @@
...
+            this.#log.error(`Timesync save error: ${err}`);
+          });
+        }
+      } catch (err) {
+        this.#log.error(`Timesync processing error: ${err}`);
+      }
+    }
+
+    if (responses.length % 2 === 0) {
+      // the first response is most likely to be an outlier, if we have to discard it
```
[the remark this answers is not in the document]
passcod: if we have an even number, we need to discard one. we don't unilaterally discard the first

packages/facility-server/app/ApplicationContext.js
```diff
@@ -51,6 +51,7 @@ export class ApplicationContext {
       acc[facilityId] = new ReadSettings(this.models, facilityId);
       return acc;
     }, {});
+    this.settings.global = new ReadSettings(this.models);
```
dannash100, line 54: Global settings should be included in facility settings reader, but I can see this being useful for ease when don't care about facility

dannash100 (approved):
Beautiful work! Tiny bit of duplication with the route when setting timesync in context but its nothing

PR #7370, opened 2025-03-14
tweak(tamanuWeb): Update global disabled button style

tcaiger:
### Changes

- Update global disabled button style

This change was requested by Flic in SAV-868 but doing it in a seperate PR so that everyone can be aware of it

### Deploys

### Remember to...

- ...write or update tests


packages/web/app/constants/styles.js
```diff
@@ -4,6 +4,11 @@ export const MUI_SPACING_UNIT = 8;
 // (with the exception of searchTintColor)
 export const Colors = {
   primary: '#326699',
+  primaryBlue: {
+    '100': '#326699',
+    '30': '#C2D2E1',
+    '10': '#EBF0F5',
+  },
```
tcodling, line 11: Is this referring to opacity? Not very straightforward on first glance what is going on here to be honest 🤔 When would we ever use '100' when we have `primary` already

tcaiger: yes fair questions - I meant to (forgot) to put a question on that idea. My thinking was to try to organise the colors to match the colors in figma in [Flic's new library](https://www.figma.com/design/0zuHw8wmEMhEL5IJ0apjBd/Tamanu-Design-System?node-id=27-685).

I think the way she has done it is a color system rather than opacity per say. Material UI does the same thing. https://mui.com/material-ui/customization/color/?srsltid=AfmBOor5eBOt7xGLyHQ3Xc-wAv93VfyQHXDTBAhVnpznRLXzzO0wG7gh

tcodling: Ah ok! Might feel a bit weird to people that haven't seen it before initially (like me!) but i like the system! Only thing that feels a little weird to me is that we have both `primary` and `primaryBlue` but I would understand if you don't want to standardise that in this pr as feels a bit out of scope

tcodling (approved):
Approved. I feel that there could be some more standardisation/cleanup of the colors constant to reflect our new colour organisation but feels out of scope for this

PR #7372, opened 2025-03-14
tweak(audit): use uuid primary key for changelog table

passcod:
### Changes

Thinking more about this, we really need a UUID primary key to be able to sync the table without collisions.

Ran the formatter so you probably want to review with whitespace changes omitted.


dannash100 (approved):
Looks gud, seems like there sum lint issue with timesync fix i did and dbt out of date from this change.

PR #7373, opened 2025-03-14
feat(programRegistry): SAV-912: Translated enum in mobile

NavarroEmilioLuis:
### Changes

Welp

So doing a full feature on this seems unfeasible with the time we have at hand. Translated enums do not exist on mobile and reconciliating everything within the two packages is overkill. Also a lot of these enums aren't really part of mobile at all and the ones that are used are called differently so yeah hard to find. I think we will have to start adding them case by case instead of all in one go


tcaiger (approved):
LGTM - I am happy with the update as we go approach

PR #7381, opened 2025-03-16
tweak(audit): add more fields to log tables to harmonise them

passcod:


passcod:
This needs to be rebased, it's really just this one commit: https://github.com/beyondessential/tamanu/pull/7381/commits/4c8792d8a670f7d96a2a6f7f1bfd9e587f959187

passcod:
Really it's just adding `device_id` and `tamanu_version` columns to both migration logs and changelogs, and adding this function:

```ts
import type { QueryInterface } from 'sequelize';
 


 

export async function up(query: QueryInterface) {
 

  await query.sequelize.query(`
 

    CREATE OR REPLACE FUNCTION local_system_fact(the_key text, the_default text)
 

    RETURNS NULL ON NULL INPUT
 

    RETURNS text AS $$
 

      DECLARE
 

        the_value text;
 

      BEGIN
 

        SELECT value INTO the_value
 

        FROM local_system_facts WHERE key = the_key;
 

        IF NOT FOUND THEN
 

          the_value = the_default;
 

        END IF;
 

        RETURN the_value;
 

      END;
 

    $$ LANGUAGE plpgsql STABLE PARALLEL SAFE;
 

  `);
 

}
 


 

export async function down(query: QueryInterface) {
 

  await query.sequelize.query('DROP FUNCTION local_system_fact CASCADE');
}
```

-- it might be faster to just redo just that from scratch instead rebasing or merging or whahtever

PR #7382, opened 2025-03-17
ci: KAM-303: dbt deploy

passcod:
- The central/facility dichotomy in database/models is useless, as there's basically no true difference between facility and central. Some tables don't exist (or are used) in facility that exist in central, but not the other way around (FHIR). So this PR deletes all the facility db models and removes the functionality around that; this makes things a bit faster and also removes a source of confusion and inconsistency.
- Then we generate and deploy DBT docs to be publicly hosted on a proper website (Daniel).
- Daniel has also figured out how to replace the frontpage text, so that's now the contents of the `database/README.md` file.
- Finally, on that same file, we add a notice to point to the right URL for people coming from slab.

We deploy:
- the `main` branch to <https://data.bes.au/tamanu/main/>
- the `release/*` branches to e.g. <https://data.bes.au/tamanu/release-2.27/>
- the released versions to e.g. <https://data.bes.au/tamanu/v2.27.3/>


NavarroEmilioLuis (approved):
LGTM

PR #7384, opened 2025-03-17
feat(scripts): Add translations scraper directory option

NavarroEmilioLuis:
### Changes

Okay so this is a pretty weird one and possibly experimental🙏motivation is that aside from the master translation list, I think it's useful to only get the translations updated in a specific branch. This should allow us to get said list!

### DIFF
--- a/.github/scripts/scrape-translations.sh
+++ b/.github/scripts/scrape-translations.sh
@@ -10,8 +10,12 @@ gtregex="getTranslation\(\s*?[\"'](.*?)[\"'],.*?[\"'](.*?)[\"'].*?\)"
 # running git diff main..<branch> --unified=0 | grep '^+ ' > branch-diffs.txt
 directory=${1-./packages}
 
-# Get all translated string data from registered enums, TranslatedText and getTranslatedString.
-teoutput=$(npx tsx packages/constants/scripts/printTranslatedEnums.ts)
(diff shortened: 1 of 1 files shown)


edmofro:
What's the motivation for this @NavarroEmilioLuis?

Note that the admin panel has the ability to only import new rows, and ignore those that are already imported

NavarroEmilioLuis:
@edmofro the motivation is to have a list of the new ones for "auditability" purposes? So instead of having a massive list, only the ones who were modified can be easily found here. The purpose of this is not really to export as a csv and import to the server, but rather to get the list of strings independently?

It could also work for the second one, as it's a smaller, easier list to modify 🤔

edmofro:
But where do you want to see that smaller list?

NavarroEmilioLuis:
@edmofro The actual output? .txt or .csv would work just fine. The reason is really to get all the string IDs we are using on the program registry translations epic, cause those are the one we need to test, this can be done as one off but I thought it was worth keeping around for the same reason

edmofro:
Oh yep it was this I was hunting for, the original motivation for doing this

> The reason is really to get all the string IDs we are using on the program registry translations epic, cause those are the one we need to test

In which case, really nice util ;-)

edmofro (approved):

PR #7387, opened 2025-03-17
Hacky hack

NavarroEmilioLuis:
### Changes

Will it blend?


NavarroEmilioLuis:
So it was a circular dependency causing the issue in file packages/web/app/views/programRegistry/ProgramRegistrySidebarItem.jsx. Tom caught it already and I didn't realize. This was done in commit: https://github.com/beyondessential/tamanu/pull/7333/commits/7190747eb70308a5df34155ed0c4d42425b96a19

PR #7392, opened 2025-03-18
fix(exporter): sort role columns for permissions export

passcod:
### Changes

Specifically for the permissions exporter, sort the role columns in lexical order, so it's easier to compare two deployments.

### Deploys

### DIFF
--- a/packages/central-server/app/admin/exporter/modelExporters/PermissionExporter.js
+++ b/packages/central-server/app/admin/exporter/modelExporters/PermissionExporter.js
@@ -33,6 +33,9 @@ export class PermissionExporter extends ModelExporter {
   }
 
   getHeadersFromData(data) {
-    return Object.keys(data[0]);
+    const heads = ['verb', 'noun', 'objectId'];
+    const cols = Object.keys(data[0]).filter((col) => !heads.includes(col));
+    cols.sort();
+    return heads.concat(cols);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/central-server/app/admin/exporter/modelExporters/PermissionExporter.js
```diff
@@ -1,38 +1,45 @@
...
+        item.set('objectId', permission.objectId);
+        for (const role in this.#roles) {
+          item.set(role, '');
+        }
+        data.set(key, item);
       }
-    });
 
-    return Object.values(data);
+      item.set(permission.roleId, permission.deletedAt ? 'n' : 'y');
```
NavarroEmilioLuis, line 36: Well, a bit elegant in that we don't have to handle two different cases but yeah, reads different than just doing one thing if item exists, another one if item doesn't and creating the object in one go (not sure if the map helps us here with ordered props?)

NavarroEmilioLuis (approved):
I'd say unusual approach on writing but should work,. I wonder if sorting by role columns is enough or if we would want to actually sort the whole thing by the `key`? Seems like only ordering by role could potentially make it so that verbs and nouns are not necessarily consistent?

Also left two style/preference comments that can obviously be ignored.

NavarroEmilioLuis (approved):
Nice

passcod:
Tested in the auto-deploy: passed.

PR #7395, opened 2025-03-18
fix(suggestions): SAV-898: Inability to search by key word

NavarroEmilioLuis:
### Changes

Sneaky issue because there is a lot of misdirection. However the problem is that we were overwriting the `suggestedIds` query key (`id`) when destructuring the `omit(whereQuery, 'name')` object.

There might be a potential gain to simply check whether that object contains an `id` key, and only then include the `AND` block, at the cost of obfuscating things further. I decided not to do it, but could make a pretty function for it, as I think inline would be too obnoxious to read

### Deploys

### DIFF
--- a/packages/facility-server/app/routes/apiv1/suggestions.js
+++ b/packages/facility-server/app/routes/apiv1/suggestions.js
@@ -84,7 +84,10 @@ function createSuggesterRoute(
         [Op.or]: [
           whereQuery,
           {
-            id: { [Op.in]: suggestedIds },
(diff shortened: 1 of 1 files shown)


tcaiger (approved):
Change looks good to me. It would be ideal to add a test to Suggestions.test.js if time permits

NavarroEmilioLuis:
> Change looks good to me. It would be ideal to add a test to Suggestions.test.js if time permits

Yeah sounds like a plan. Will omit from here and just work it out on a different PR tomorrow

PR #7403, opened 2025-03-19
fix(programRegistry): SAV-912: Update string ID prefixes

NavarroEmilioLuis:
### Changes

Mobile was missing the update


dannash100 (approved):
:shipit:

PR #7404, opened 2025-03-19
fix(programRegistry): Merge program registries epic branches

NavarroEmilioLuis:
### Changes

Just a chore


NavarroEmilioLuis:
This is the contrary diff. If you try to edit the 'into' branch to `epic-program-registries-2` it errors with:

There are no new commits between base branch 'epic-program-registries-2' and head branch 'fix/merge-program-registries-epic-branches-20250319

Sooo they are up to date, I will remove the former. Note that the reason we have to is that the first one wasn't able to deploy for some weird reason!

PR #7413, opened 2025-03-20
fix(mobile): Unblock sync for deleted records

NavarroEmilioLuis:
### Changes

So we upgraded typeorm version on v2.26 - in this version the findByIds method does not accept options so we didn't get soft deleted records... Meaning that didn't realize they needed to be updated instead of created!

See: https://github.com/typeorm/typeorm/blame/master/src/repository/Repository.ts#L588-L589

### Deploys

### DIFF
--- a/packages/mobile/App/models/Patient.ts
+++ b/packages/mobile/App/models/Patient.ts
@@ -1,4 +1,4 @@
-import { Column, Entity, Index, OneToMany } from 'typeorm';
+import { Column, Entity, Index, OneToMany, In } from 'typeorm';
 import { getUniqueId } from 'react-native-device-info';
 import { addHours, parseISO, startOfDay, subYears } from 'date-fns';
 import { groupBy } from 'lodash';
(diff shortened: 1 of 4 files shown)


edmofro (changes requested):
Man good catch. How did you come across it?

Btw there are a few other spots.

NavarroEmilioLuis:
@edmofro Priya found an issue when testing patient contacts and had to dig quite a lot!!

NavarroEmilioLuis:
@edmofro also I think this is the only one where we use `findByIds` in mobile that requires deleted records (or use any options). Unless I'm missing something? I think the rest of the methods were catch by Thomas on the update one, this one is super sneaky

edmofro:
Oh I see, the others don't require options! Perhaps we should discontinue any use across the board, lest someone follow the pattern somewhere that soft deletion is required without realising it's not supported?

edmofro:
And amazing work by Priya! Sheesh this one would have been _gnarly_ in prod

edmofro (approved):

PR #7422, opened 2025-03-20
fix(ci): no-issue: Fix storybook __dirname unresolved errors

rohan-bes:
### Changes

Now that we use modules for this file we need to switch to manually define `__dirname`: https://nodejs.org/docs/latest-v15.x/api/esm.html#esm_no_require_exports_or_module_exports

When I run storybook locally it seems to use cjs, but in CI it uses ESM... not sure what's going on there and perhaps there's a higher order fix that could be done here.

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/.storybook/main.mjs
+++ b/packages/web/.storybook/main.mjs
@@ -1,5 +1,9 @@
+import { fileURLToPath } from 'url';
 import { dirname, join, resolve } from 'path';
 import { mergeConfig } from 'vite';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
 
 function getAbsolutePath(packageName) {
   return dirname(resolve(__dirname, join('../..', packageName, 'package.json')));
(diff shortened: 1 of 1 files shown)


packages/web/.storybook/main.mjs
```diff
@@ -1,13 +1,29 @@
...
+function getNodeModulePath(workspace, packageName) {
+  return dirname(resolve(dir, join(workspace, 'node_modules/', packageName, 'package.json')));
+}
+
+function getRootNodeModulePath(packageName) {
+  return getNodeModulePath('../../../', packageName);
+}
+
+function getWorkspaceNodeModulePath(packageName) {
+  return getNodeModulePath('../', packageName);
```
rohan-bes, line 21: Having to manually resolve these `node_module` paths, and having to track where in the monorepo they get installed feels very clunky. Open to any idea on how to make this more stable and simple :pray:

passcod (approved):
This looks incredibly cursed but yay for greening the build!

PR #7425, opened 2025-03-20
fix(programRegistry): SAV-912: Mobile translations

NavarroEmilioLuis:
### Changes

A couple of spots difficult to see

### DIFF
--- a/packages/database/src/models/TranslatedString.ts
+++ b/packages/database/src/models/TranslatedString.ts
@@ -127,7 +127,7 @@ export class TranslatedString extends Model {
       where: {
         language,
         stringId: {
-          [Op.startsWith]: `${REFERENCE_DATA_TRANSLATION_PREFIX}.${refDataType}`,
+          [Op.startsWith]: `${REFERENCE_DATA_TRANSLATION_PREFIX}.${refDataType}.`,
         },
         ...(queryString ? { text: { [Op.iLike]: `%${queryString}%` } } : {}),
       },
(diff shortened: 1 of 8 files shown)


packages/database/src/models/TranslatedString.ts
```diff
@@ -127,7 +127,7 @@ export class TranslatedString extends Model {
       where: {
         language,
         stringId: {
-          [Op.startsWith]: `${REFERENCE_DATA_TRANSLATION_PREFIX}.${refDataType}`,
+          [Op.startsWith]: `${REFERENCE_DATA_TRANSLATION_PREFIX}.${refDataType}.`,
```
NavarroEmilioLuis, line 130: The reason for this change is that there are some `refDataType` that include others as prefix, for example:

programRegistry
programRegistryClinicalStatus
programRegistryConditions

There might be some others outside of program registries, but the general problem is that, given enough string IDs, some might get buried on the lookup. This also should improve performance as we will discard unnecessary records - that is unless not all string IDs have a dot right after the `refDataType`

packages/mobile/App/models/TranslatedString.ts
```diff
@@ -66,7 +66,7 @@ export class TranslatedString extends BaseModel {
     const referenceDataTranslations = await this.getRepository().find({
       where: {
         language,
-        stringId: Like(`refData.${refDataType}%`),
+        stringId: Like(`refData.${refDataType}.%`),
```
NavarroEmilioLuis, line 69: Same here

tcodling (approved):
lgtm! edit: Actually its failing the mobile test! looks like easy fix though

PR #7450, opened 2025-03-24
tweak(central): default the mobile update URLs to the right thing

passcod:
### Changes

From an idea by @tcodling in [NASS-1266](https://linear.app/bes/issue/NASS-1266/smooth-updates-of-s3-app-builds#comment-153639c1).

We add a placeholder string replacement to the `updateUrls.mobile` with the minimum required version `{minVersion}`, and change the default to be `https://meta.tamanu.app/versions/^{minVersion}/mobile`. Since the meta server supports npm-style version ranges, this then resolves to the latest version in the release branch.

Then a manual release step that's needed is to drop the `updateUrls` block from the config on central (to let the default apply), and 🎉 mobile updates will "just work" from now on, no further config changes needed on upgrade.

### Deploys

### DIFF
--- a/packages/central-server/config/default.json5
+++ b/packages/central-server/config/default.json5
@@ -37,7 +37,7 @@
     }
   },
   "updateUrls": {
-    "mobile": ""
+    "mobile": "https://meta.tamanu.app/versions/^{minVersion}/mobile",
   },
   "log": {
     "path": "",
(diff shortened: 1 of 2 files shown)


tcodling (approved):
clean!

anilvkaran (approved):

PR #7453, opened 2025-03-24
feat(reports): SAV-908: Create new field for sensitive lab test categories in reports

NavarroEmilioLuis:
### Changes

Create a new field and updates the old one to exclude sensitive ones

### Deploys

### DIFF
--- a/packages/facility-server/app/routes/apiv1/suggestions.js
+++ b/packages/facility-server/app/routes/apiv1/suggestions.js
@@ -592,6 +592,50 @@ createSuggester(
   },
 );
 
+createSuggester(
+  'nonSensitiveLabTestCategory',
+  'ReferenceData',
+  (search) => {
+    const baseWhere = DEFAULT_WHERE_BUILDER(search);
+    return {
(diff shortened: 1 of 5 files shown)


packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -592,6 +592,48 @@ createSuggester(
   },
 );
 
+createSuggester(
```
tcaiger, line 595: Is this new endpoint necessary? It seems like it's not used.

tcaiger (approved):
Nice one

packages/facility-server/app/routes/apiv1/suggestions.js
NavarroEmilioLuis: Yep, it's used in: https://github.com/beyondessential/tamanu/pull/7453/files#diff-a13eff1751e2c4e50edd0fb141f5b4b2e6e55c7f9637600ddf96510b121da25fR17

PR #7467, opened 2025-03-26
fix(programRegistry): SAV-868: Program registry exports

tcaiger:
### Changes
If a table cell component or accessor references another component and that component has a translation then downloads will break for that table. The core issue is that normalizeRecursively function in DownloadDataButton doesn’t take into account nested components.

This PR just patches the issue for the Program Registry Table but the issue might come up again as more table cells get translated so I've created a ticket to look at a more sustainable fix.

https://linear.app/bes/issue/NOTAM-457/table-cellcomponents-break-datadownload

### Deploys

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
+++ b/packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
@@ -11,7 +11,7 @@ import { ChangeStatusFormModal } from './ChangeStatusFormModal';
 import { Colors } from '../../constants';
 import { LimitedLinesCell } from '../../components/FormattedTableCell';
 import { RegistrationStatusIndicator } from './RegistrationStatusIndicator';
-import { ClinicalStatusDisplay } from './ClinicalStatusDisplay';
+import { ClinicalStatusCell } from './ClinicalStatusDisplay';
 import { useRefreshCount } from '../../hooks/useRefreshCount';
 import { ActivatePatientProgramRegistry } from './ActivatePatientProgramRegistry';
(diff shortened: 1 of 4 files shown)


packages/web/app/views/programRegistry/ProgramRegistryTable.jsx
```diff
@@ -125,9 +125,7 @@ export const ProgramRegistryTable = ({ searchParameters }) => {
       {
         key: 'clinicalStatus',
         title: <TranslatedText stringId="programRegistry.clinicalStatus.label" fallback="Status" />,
-        accessor: row => {
```
tcaiger, line 128: converting this from an accessor didn't make a different in terms of the bug but seemed more consistent with how CellComponents are used everywhere else.

rohan-bes (approved):
Thanks for creating the ticket for the deeper fix :pray:

I'm mildly curious if there's anything wrong with just wrapping all the components in the translation context regardless of if they need it? Trying to track down if an element has nested translation stuff seems tricky...

Approving!

tcaiger:
> Thanks for creating the ticket for the deeper fix 🙏
> 
> I'm mildly curious if there's anything wrong with just wrapping all the components in the translation context regardless of if they need it? Trying to track down if an element has nested translation stuff seems tricky...
> 
> Approving!

Yes I agree that wrapping all the components is probably a pretty good idea. I did a quick test of that and it seemed to work well without performance issues. I didn't want to rush that fix into a release late in the cycle though which is why I just did this but I'll add a note to the ticket.

PR #7493, opened 2025-03-30
tweak(testing): no-issue: Make e2e test environment configurable

avaek:
### Changes

- Introduces a `.env` file to allow developers to run tests locally or against an external deployed environment
- Removes `baseURL` from `playwright.config.ts` so tests can be run against either the facility frontend or admin frontend
- If the `LAUNCH_LOCAL_SERVERS_WHEN_RUNNING_TESTS` environment variable is set to `true` then Playwright will automatically start your local environments for you when the tests are run
- The admin panel frontend has been added to the list of environments to run locally so tests can be built against it

This is my first PR to Tamanu so if there are any best practices I've missed or any corrections at all that you'd recommend please let me know 🙏 If my entire approach here is wrong, please let me know that too 😅


packages/e2e-tests/.env.example
tcodling, line 1: Great comments and named variables!

packages/e2e-tests/playwright.config.ts
```diff
@@ -70,28 +72,37 @@ export default defineConfig({
...
-    {
-      command: 'npm run client-start-dev --workspace=@tamanu/web-frontend',
-      port: 5173,
-      reuseExistingServer: !process.env.CI,
-      timeout: 240 * 1000,
-      stdout: 'pipe',
-    },
-  ],
+  /* Automatically run your local servers and frontends before starting the tests if running tests against local environment */
+  ...(process.env.LAUNCH_LOCAL_SERVERS_WHEN_RUNNING_TESTS === 'true' ? {
```
tcodling, line 76: Clean! You could do something like `Boolean(process.env.LAUNCH_LOCAL_SERVERS_WHEN_RUNNING_TESTS)` to resolve the string and run the logic like its a boolean but honestly this is clearer to read so i wouldn't change it :))

packages/e2e-tests/utils/urls.ts
tcodling, line 1: not sure about the name of this file 🤔 I feel like it would make more sense to me if it was called something like `navigation.ts`

This is totally just personal preference though so you name it whatever makes most sense to you as youll be using it the most. But if you were imagining this file was eventually going to contain a bunch of functions called `goToXXX` that take you to different places in the app, I think navigation makes sense but otherwise something like `urls` or `serverUrls` would be fine

tcodling (approved):
Amazing first pr!!

packages/e2e-tests/utils/urls.ts
avaek: Thanks for the review @tcodling! I agree, your suggestion for the new name for this file makes way more sense - I've updated that file now 🙏

tcodling (approved):
lgtm!

PR #7528, opened 2025-04-07
fix(fhir): SAV-945: Handle undefined upstreamIds when building FHIR References (hotfix for v2.26)

rohan-bes:
### Changes

This fix won't change the underlying issue of why we're seeing `undefined` upstreamIds in Nauru. However since that's the only deployment where this issue is occurring I'd like to just change the code to log and ignore `undefined` upstreamIds and not block the resolver task. Then we can investigate further which resources seem to cause this behaviour.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/models/fhir/Resource.ts
+++ b/packages/database/src/models/fhir/Resource.ts
@@ -200,7 +200,20 @@ export class FhirResource extends Model {
     });
 
     for (const unresolvedResource of unresolvedResources) {
-      await this.materialiseFromUpstream(unresolvedResource.upstreamId);
+      try {
+        await this.materialiseFromUpstream(unresolvedResource.upstreamId);
+      } catch (error) {
+        if (error instanceof Error) {
+          // Rethrowing like this to preserve stacktrace while logging which resource failed to resolve
+          const rethrownError = new Error(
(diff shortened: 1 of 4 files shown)


packages/database/src/models/fhir/Resource.ts
```diff
@@ -200,7 +200,20 @@ export class FhirResource extends Model {
...
+        if (error instanceof Error) {
+          // Rethrowing like this to preserve stacktrace while logging which resource failed to resolve
+          const rethrownError = new Error(
+            `Error resolving upstreams for ${this.fhirName}/${unresolvedResource.id}: ${error.stack ?? error.message ?? error.toString() ?? ''}`,
+          );
+          rethrownError.stack = error.stack;
+          throw rethrownError;
+        } else {
+          throw error;
+        }
```
rohan-bes, line 215: Apparently there's a much cleaner way of doing this as of Node 16, using the `cause` option: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

However our typescript compiler didn't seem happy with that, and I didn't want to play around with our TS config on a hotfix. I think I'll try and get it working before merging to main though :+1:

dannash100 (approved):
Looks good :) thank you for all your work on this stuff

PR #7532, opened 2025-04-10
fix: Fix localised field string ids

flynncs:
### Changes

- removed the 's' from a few localised field string ids

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


passcod:
Is there any possibility of production locales needing a change due to this?

flynncs:
I don't think so, as the localised field translations should already exist under the correct string id's & the incorrect id's don't exist in the translations spreadsheets, but I'll have to double check

flynncs:
yup confirmed with Daniel it should be fine 🙏

dannash100 (approved):
good catch

PR #7537, opened 2025-04-10
feat(programRegistry): SAV-861: Update patient program registry schema

NavarroEmilioLuis:
### Changes

A couple of tricky steps because of how the query needs to change depending on the db status (if it's from scratch or not)

### DIFF
--- a/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
+++ b/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
@@ -15,7 +15,7 @@ export async function up(query: QueryInterface): Promise<void> {
   }
 
   // Save previously set time zone
-  const previousTimeZoneQuery = await query.sequelize.query('show timezone');
+  const previousTimeZoneQuery: any = await query.sequelize.query('show timezone');
   const previousTimeZone = previousTimeZoneQuery[0].TimeZone;
 
   // Set time zone defined in config
--- a/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
+++ b/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
@@ -1,4 +1,6 @@
 import { FACT_CURRENT_SYNC_TICK, SYSTEM_USER_UUID } from '@tamanu/constants';
+import { selectFacilityIds } from '@tamanu/utils/selectFacilityIds';
+import config from 'config';
 import { QueryInterface, QueryTypes } from 'sequelize';
 
 interface tableOid {
@@ -12,6 +14,25 @@ export async function up(query: QueryInterface): Promise<void> {
   );
   const tableOid = tableOidQuery.oid;
 
+  // Check if updated_at_sync_tick exists in logs.changes
+  const [changesUpdatedAtSyncTickQuery]: any = await query.sequelize.query(`
+    SELECT EXISTS (SELECT TRUE
+    FROM information_schema.columns 
+    WHERE table_schema = 'logs' AND table_name = 'changes' AND column_name = 'updated_at_sync_tick');
+  `);
+  const changesHasUpdatedAtSyncTick = changesUpdatedAtSyncTickQuery?.[0]?.exists;
+  const updatedAtSyncTickSelect = `(SELECT value FROM local_system_facts WHERE key = '${FACT_CURRENT_SYNC_TICK}')::bigint,`;
+
+  // Check if updated_at_sync_tick exists in patient_program_registrations
+  const [pprUpdatedAtSyncTickQuery]: any = await query.sequelize.query(`
+    SELECT EXISTS (SELECT TRUE
+    FROM information_schema.columns 
+    WHERE table_schema = 'public' AND table_name = 'patient_program_registrations' AND column_name = 'updated_at_sync_tick');
+  `);
+  const pprHasUpdatedAtSyncTick = pprUpdatedAtSyncTickQuery?.[0]?.exists;
+  const isFacilityServer = !!selectFacilityIds(config);
+  const syncTickInitialValue = isFacilityServer ? '-999,' : '0,';
+
   // Migrate historical changes to audit table
   await query.sequelize.query(`
     INSERT INTO logs.changes (
@@ -22,10 +43,14 @@ export async function up(query: QueryInterface): Promise<void> {
       created_at,
       updated_at,
       deleted_at,
-      updated_at_sync_tick,
+      ${changesHasUpdatedAtSyncTick ? 'updated_at_sync_tick,' : ''}
       updated_by_user_id,
       record_id,
       record_update,
+      record_created_at,
+      record_updated_at,
+      record_deleted_at,
+      record_sync_tick,
       record_data
     )
     SELECT
@@ -36,10 +61,14 @@ export async function up(query: QueryInterface): Promise<void> {
       now(),
       now(),
       CASE WHEN ppr.deleted_at IS NOT NULL THEN now() ELSE NULL END,
-      (SELECT value FROM local_system_facts WHERE key = '${FACT_CURRENT_SYNC_TICK}')::bigint,
+      ${changesHasUpdatedAtSyncTick ? updatedAtSyncTickSelect : ''}
       COALESCE(ppr.clinician_id::text, '${SYSTEM_USER_UUID}'),
       ppr.id,
       registration_summary.is_insert,
+      ppr.created_at,
+      ppr.updated_at,
+      ppr.deleted_at,
+      ${pprHasUpdatedAtSyncTick ? 'ppr.updated_at_sync_tick,' : syncTickInitialValue}
       to_jsonb(ppr.*)
     FROM patient_program_registrations ppr
     JOIN (
--- a/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
+++ b/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
@@ -1,0 +1,89 @@
+import { FACT_CURRENT_SYNC_TICK, SYSTEM_USER_UUID } from '@tamanu/constants';
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+interface tableOid {
+  oid: number;
+}
+
+export async function up(query: QueryInterface): Promise<void> {
+  const [tableOidQuery]: any = await query.sequelize.query<tableOid>(
+    `SELECT oid FROM pg_class WHERE relname = 'patient_program_registrations';`,
+    { type: QueryTypes.SELECT },
+  );
+  const tableOid = tableOidQuery.oid;
+
+  // Migrate historical changes to audit table
+  await query.sequelize.query(`
+    INSERT INTO logs.changes (
+      table_oid,
+      table_schema,
+      table_name,
+      logged_at,
+      created_at,
+      updated_at,
+      deleted_at,
+      updated_at_sync_tick,
+      updated_by_user_id,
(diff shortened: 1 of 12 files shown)


packages/database/src/models/PatientProgramRegistration.ts
```diff
@@ -102,42 +96,9 @@ export class PatientProgramRegistration extends Model {
     });
   }
 
-  static async create(values: any): Promise<any> {
```
NavarroEmilioLuis, line 105: No need to replace the `create` method as this should not be append only now

packages/database/src/models/PatientProgramRegistration.ts
```diff
@@ -102,42 +96,9 @@ export class PatientProgramRegistration extends Model {
...
-    };
-
-    // Ensure a new id is generated, rather than using the one from existingRegistration
-    delete newRegistrationValues.id;
-
-    return super.create(newRegistrationValues);
-  }
-
-  static async getMostRecentRegistrationsForPatient(patientId: string) {
+  static async getRegistrationsForPatient(patientId: string) {
```
NavarroEmilioLuis, line 99: Use a more sensible name? Other occurrences have also been replaced except for mobile which is another card

packages/facility-server/__tests__/apiv1/PatientProgramRegistration.test.js
```diff
@@ -193,50 +158,6 @@ describe('PatientProgramRegistration', () => {
         programRegistryConditionId: programRegistryCondition.id,
       });
     });
-
-    it('appends a new entry to the history for a program registration', async () => {
```
NavarroEmilioLuis, line 197: Test removed cause there is no point on testing this: auditability work should already have their unit tests (plus that work isn't ready here yet)

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -2,7 +2,7 @@ import { trace } from '@opentelemetry/api';
 import { Op, QueryTypes } from 'sequelize';
 import _config from 'config';
 
-import { SYNC_DIRECTIONS, DEBUG_LOG_TYPES, SETTINGS_SCOPES, AUDIT_PAUSE_KEY } from '@tamanu/constants';
+import { DEBUG_LOG_TYPES, SETTINGS_SCOPES, AUDIT_PAUSE_KEY } from '@tamanu/constants';
```
NavarroEmilioLuis, line 5: The only change for sync owners I think. This is just an import that wasn't used that the linter was complaining. Uhhh I presume it was caused by a faulty merge conflict elsewhere

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+    `SELECT oid FROM pg_class WHERE relname = 'patient_program_registrations';`,
+    { type: QueryTypes.SELECT },
+  );
+  const tableOid = tableOidQuery.oid;
+
+  // Check if updated_at_sync_tick exists in logs.changes
+  const [changesUpdatedAtSyncTickQuery]: any = await query.sequelize.query(`
+    SELECT EXISTS (SELECT TRUE
+    FROM information_schema.columns 
+    WHERE table_schema = 'logs' AND table_name = 'changes' AND column_name = 'updated_at_sync_tick');
```
dannash100, line 21: Should these other queries have type SELECT or no 🤔

dannash100, line 21: I think we are just using record_sync_tick in my upcoming sync work. So no need for updated_at_sync_tick. Or is this backwards compatibility

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // Check if updated_at_sync_tick exists in patient_program_registrations
+  const [pprUpdatedAtSyncTickQuery]: any = await query.sequelize.query(`
+    SELECT EXISTS (SELECT TRUE
+    FROM information_schema.columns 
+    WHERE table_schema = 'public' AND table_name = 'patient_program_registrations' AND column_name = 'updated_at_sync_tick');
+  `);
+  const pprHasUpdatedAtSyncTick = pprUpdatedAtSyncTickQuery?.[0]?.exists;
+  const isFacilityServer = !!selectFacilityIds(config);
+  const syncTickInitialValue = isFacilityServer ? '-999,' : '0,';
```
dannash100, line 34: intentional?
```suggestion
  const syncTickInitialValue = isFacilityServer ? '-999,' : '0';
```

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+  `);
+
+  // Update schema
+  await query.removeColumn('patient_program_registrations', 'is_most_recent');
+}
+
+export async function down(query: QueryInterface): Promise<void> {
+  // Add the is_most_recent column back
+  await query.addColumn('patient_program_registrations', 'is_most_recent', {
+    type: 'BOOLEAN',
```
dannash100, line 102:
```suggestion
    type: DataTypes.BOOLEAN,
```

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  // Set all to be true
+  await query.sequelize.query(`
+    UPDATE patient_program_registrations
+    SET is_most_recent = TRUE;
+  `);
+
+  // Remove historical records from audit table
+  await query.sequelize.query(`
```
dannash100, line 114: I wonder if this should just include the migrated ones, I dunno if thats possible tho

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+      record_updated_at,
+      record_deleted_at,
+      record_sync_tick,
+      record_data
+    )
+    SELECT
+      ${tableOid},
+      'public',
+      'patient_program_registrations',
+      to_timestamp(ppr.date, 'yyyy-mm-dd hh24:mi:ss'),
```
dannash100, line 60: Do we need to set timezone before this conversion - see other migrations. Ignore if not

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -0,0 +1,118 @@
...
+      CASE WHEN ppr.deleted_at IS NOT NULL THEN now() ELSE NULL END,
+      ${changesHasUpdatedAtSyncTick ? updatedAtSyncTickSelect : ''}
+      COALESCE(ppr.clinician_id::text, '${SYSTEM_USER_UUID}'),
+      ppr.id,
+      registration_summary.is_insert,
+      ppr.created_at,
+      ppr.updated_at,
+      ppr.deleted_at,
+      ${pprHasUpdatedAtSyncTick ? 'ppr.updated_at_sync_tick,' : syncTickInitialValue}
+      to_jsonb(ppr.*)
```
dannash100, line 72: Should we strip this too
```suggestion
   to_jsonb(ppr.*) - 'is_most_recent',
```

dannash100 (changes requested):
Some thoughts! cool to see tho!

dannash100 (changes requested):
Some thoughts! cool to see tho!

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
dannash100: Oh i see the comment, if from scratch! does updated_at_sync_tick removal on changes table make this more simple?

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
dannash100: I think this is too tricky tho

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Interesting, I'll try again and confirm but this was working fine. Does it make a difference using the select type query?

NavarroEmilioLuis: I don't understand the question? Are you guys planning on making further changes to the `logs.changes` table?

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Yeah we probably have to good point. Would've been better with the other migration file

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: I don't see why: I wonder if there is a way to remove that is not a hassle plus migrations will happen in the future and we're not going to modify previous logs on it. It just so happens that we are also ditching that field here.

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Not possible, there is no way of telling which one is from the migration and which isn't. In any case, if we down migrate this is because we are moving to an earlier version that won't have audit tables etc.

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Yeah okay I'll add this

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Hmm okay so without select/with select:

`[ { exists: false } ] { exists: false }`

So we don't really need it, but it makes it less of a hassle to access. Though I already did it this way, I'd rather stick with it!

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
dannash100: oh!

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
dannash100: Nah leave it its all good

dannash100 (approved):
Love it!

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
dannash100: Yeah i often do select for that simplicity of results - but its chill i agree

PR #7540, opened 2025-04-11
fix(sync): SAV-947: Take advisory lock -before- updating sync tick

edmofro:
### Changes

See comment inline

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/migrations/1744340076240-fixRaceConditionInSettingUpdateSyncTick.ts
+++ b/packages/database/src/migrations/1744340076240-fixRaceConditionInSettingUpdateSyncTick.ts
@@ -1,0 +1,71 @@
+import { QueryInterface } from 'sequelize';
+
+// copied from 1692710205000-allowDisablingSyncTrigger with the order of operations fixed to
+// first get the sync tick, then lock it, then use it
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    CREATE OR REPLACE FUNCTION set_updated_at_sync_tick()
+      RETURNS trigger
+      LANGUAGE plpgsql AS
+      $func$
+      DECLARE
+        current_tick bigint;
+      BEGIN
+        IF ((SELECT value FROM local_system_facts WHERE key = 'syncTrigger') = 'disabled') THEN
+            RETURN NEW;
+        END IF;
+        -- If setting to "-1" representing "not marked as updated on this client", we actually use
+        -- a different number, "-999", to represent that in the db, so that we can identify the
+        -- difference between explicitly setting it to not marked as updated (when NEW contains -1),
+        -- and having other fields updated without the updated_at_sync_tick being altered (in which
+        -- case NEW will contain -999, easily distinguished from -1)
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/1744340076240-fixRaceConditionInSettingUpdateSyncTick.ts
```diff
@@ -0,0 +1,67 @@
...
+          NEW.updated_at_sync_tick := -999;
+        ELSE
+          -- take an advisory lock on the current sync tick (if one doesn't already exist), to
+          -- record that an active transaction is using this sync tick
+          -- see waitForPendingEditsUsingSyncTick for more details
+          PERFORM pg_try_advisory_xact_lock_shared(NEW.updated_at_sync_tick);
+
+          -- now that the advisory lock is acquired, we can safely select the current sync tick and
+          -- know that no sync session will skip over this change until the transaction is complete
+          SELECT value FROM local_system_facts WHERE key = 'currentSyncTick' INTO NEW.updated_at_sync_tick;
```
edmofro, line 30: The only difference is that these two are flipped.
Previously the advisory look was taken after the sync tick was selected, meaning there was a potential race condition where:
1. A record was changed with sync tick x
2. A sync snapshot including this sync tick started 
3. The pg advisory lock was taken
Steps 1 and 3 are the two lines of SQL visible above (in their old order). If step 2 managed to slip in between them, it could think that the snapshot had all records for sync tick x, but actually there was still an in-flight transaction with another record, so that record would be skipped by that sync session. 

In reality, this would (I think) never ever happen, because within step 2. there would be a few lines of javascript, with awaits, and an SQL query, which would be unlikely to fit between 1. and 3.

This is where we do the wait
```
      const { tick: currentTick } = await this.tickTockGlobalClock();

      await this.waitForPendingEdits(currentTick);
```
So the record save would need to happen before the tick tock first line (if it happened afterwards, it would use the "tock" of the tick tock, which would not be considered part of the range of this snapshot), but the advisory lock would need to be taken after the tick tock. 

FWIW, the tick tock code looks like this:
```
  static async incrementValue(key: FactName, amount: number = 1) {
    const [rowsAffected] = await this.sequelize.query(
      `
        UPDATE
          local_system_facts
        SET
          value = value::integer + :amount,
          updated_at = NOW()
        WHERE
          key = :key
        RETURNING
          value;
      `,
      { replacements: { key, amount } },
    );
    if (rowsAffected.length === 0) {
      throw new Error(`The local system fact table does not include the fact ${key}`);
    }
    const fact = rowsAffected[0] as LocalSystemFact;
    return fact.value;
  }
```
So it's a bit of work.

edmofro:
This PR is closed, but can be reopened and tested as part of the sync maintenance sprint

dannash100 (approved):
Thanks for re-grabbing this!

dannash100 (approved):
Thanks !

edmofro:
Rebased to the epic and will now merge

PR #7562, opened 2025-04-21
feat(sync): NASS-1671: Sync behavior for changelog auditing using sync_lookup_ticks

dannash100:
### Changes
Started as pairing with @edmofro 

- Sync changelogs using sync_lookup_tick to reduce expense of process - we can do this because changelogs are immutable
- Only `patient_program_registrations` change logs are pulled to facility, but can update with requirements
- All change logs are pushed up to central
- Lots of tests!

Only outstanding stuff now is more tests on CentralSyncManager

Its deploying now ✅ 

Existing failures can be ignored 
- Storybook fix is in main
- Determinism checks will resolve self on epic merge

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **New Features**
  - Changelog records are now included and synced alongside snapshot records during data synchronization between facility and central servers, improving auditability and traceability.
  - Outgoing and incoming sync operations now handle changelog records explicitly, ensuring relevant changes are tracked and persisted.

- **Bug Fixes**
  - Improved test reliability by ensuring necessary database tables are properly cleared before each test run.

- **Database**
  - Changed the primary key for change logs from an auto-incremented number to a UUID for improved uniqueness.
  - Introduced a new sync lookup ticks table to support enhanced synchronization tracking.

- **Documentation**
  - Clarified that change log IDs are UUIDs, not auto-incremented values.

- **Tests**
  - Added comprehensive test suites for changelog extraction, attachment, and insertion utilities to ensure correct behavior.


packages/database/__tests__/utils/audit/insertChangelogRecords.test.ts
```diff
@@ -0,0 +1,159 @@
...
+import { Sequelize, Op } from 'sequelize';
+
+import { selectFacilityIds } from '@tamanu/utils/selectFacilityIds';
+import { ENCOUNTER_TYPES } from '@tamanu/constants';
+
+import { insertChangelogRecords } from '../../../src/utils/audit/insertChangelogRecords';
+import type { ChangelogRecord } from '../../../src/utils/audit/attachChangelogToSnapshotRecords';
+
+// Mock dependencies
+vi.mock('config', () => ({
```
dannash100, line 11: this shud be e2e i think

packages/database/src/utils/audit/insertChangelogRecords.ts
```diff
@@ -0,0 +1,53 @@
...
+import { selectFacilityIds } from '@tamanu/utils/selectFacilityIds';
+
+import type { ChangelogRecord } from './attachChangelogToSnapshotRecords';
+
+export const insertChangelogRecords = async (
+  sequelize: Sequelize,
+  changelogRecords: ChangelogRecord[],
+) => {
+  // TODO AUDIT: what to do here do we need this
+  const isFacility = !!selectFacilityIds(config);
```
dannash100, line 13: This bit biggest todo really, this feel a little weird,
but the code is same for both ends expect for that setting of sync tick to -999

packages/central-server/app/sync/addChangelogRecords.js
```diff
@@ -0,0 +1,42 @@
...
+  const lookupTicks = await models.SyncLookupTick.findAll({
+    where: {
+      lookupEndTick: {
+        [Op.between]: [pullSince, pullUntil],
+      },
+    },
+    order: [['lookupEndTick', 'ASC']],
+  });
+
+  if (!lookupTicks.length) {
```
dannash100, line 30: Also what do we want to do if we don't have two available ticks to form the range 🤔

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -433,6 +432,70 @@ describe('CentralSyncManager', () => {
...
+        sync: {
+          lookupTable: {
+            enabled: true,
+          },
+          maxRecordsPerSnapshotChunk: DEFAULT_MAX_RECORDS_PER_SNAPSHOT_CHUNKS,
+        },
+      });
+
+      // The lookup table needs to be built twice to create a proper range of SyncLookupTick records
+      await centralSyncManager.updateLookupTable();
```
dannash100, line 477: Kinda a weird thing - but currently it relies on the two runs on lookup table to create a range of LookupSyncTicks 🤔

packages/central-server/app/sync/addChangelogRecords.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const addChangelogRecords = async (models, pullSince, pullUntil, snapshotRecords) => {
+  const { SyncLookupTick } = models;
+  const { sequelize } = SyncLookupTick;
+
+  // find the range of source sync ticks matching the lookup table ticks being synced
+  const lookupTicks = await models.SyncLookupTick.findAll({
+    where: {
+      lookupEndTick: {
+        [Op.between]: [pullSince, pullUntil],
```
dannash100, line 24: Last thing - should this be exclusive too? like other tick comparisons

packages/central-server/app/sync/addChangelogRecords.js
```diff
@@ -0,0 +1,51 @@
+import { Op } from 'sequelize';
+
+import { attachChangelogToSnapshotRecords } from '@tamanu/database/utils/audit';
+
+// After the regular sync snapshotting process is complete, we need to grab any changelog records
+// that:
+// - are associated with the records inside the snapshot; and
+// - have an updated_at_sync_tick matching the range of lookup table ticks being synced; and
```
edmofro, line 8:
```suggestion
// - have an updated_at_sync_tick matching the range of lookup table source ticks being synced; and
```

packages/central-server/app/sync/addChangelogRecords.js
edmofro: Throw an error I think? Do we expect that to ever happen? We should probably early return at the top if there are no snapshot records, but if there are snapshot records there should be lookup ticks (because they made it into the lookup table somehow). Are you worried about the first time we deploy, when we don't have anything populated into the lookup ticks table? Perhaps we need to initialise it with something.

packages/central-server/app/sync/addChangelogRecords.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    order: [['lookupEndTick', 'ASC']],
+  });
+
+  if (!lookupTicks.length) {
+    return snapshotRecords;
+  }
+
+  const minSourceTick = lookupTicks.at(0).sourceStartTick;
+  const maxSourceTick = lookupTicks.at(-1).sourceStartTick;
```
edmofro, line 38:
```suggestion
  const maxSourceTick = lookupTicks.at(-1).lookupEndTick;
```
Is that right?

packages/database/src/models/SyncLookupTick.ts
```diff
@@ -0,0 +1,43 @@
...
+  declare lookupEndTick: number;
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static initModel({ primaryKey, ...options }: InitOptions) {
+    super.init(
+      {
+        id: {
+          // Sequelize always requires an id column
+          // so this is to enforce using lookup_end_tick as the id.
+          // even tho we don't really need an id in this table.
```
edmofro, line 18:
```suggestion
          // even though we don't really need an id in this table.
```

packages/database/src/types/sync.ts
```diff
@@ -38,6 +38,15 @@ export interface SyncSnapshotAttributes {
   requiresRepull?: boolean;
 }
 
+// TODO use model on merge with other sync pr
```
edmofro, line 41: Remember this TODO

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -194,15 +197,18 @@ export class FacilitySyncManager {
     log.info('FacilitySyncManager.snapshottingOutgoingChanges', { pushSince });
     const modelsForPush = getModelsForPush(this.models);
     const outgoingChanges = await snapshotOutgoingChanges(this.sequelize, modelsForPush, pushSince);
-    if (outgoingChanges.length > 0) {
+    const outgoingChangesWithChangelogs = await attachChangelogToSnapshotRecords(this.sequelize, outgoingChanges, {
```
edmofro, line 200: Could be lower. Is it needed for that test spy? If not, would put it right above `pushOutgoingChanges`. Otherwise, at least within the .length check

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -194,15 +197,18 @@ export class FacilitySyncManager {
...
+        totalPushing: outgoingChangesWithChangelogs.length,
       });
       if (this.__testSpyEnabled) {
-        this.__testOnlyPushChangesSpy.push({ sessionId, outgoingChanges });
+        this.__testOnlyPushChangesSpy.push({ sessionId, outgoingChanges: outgoingChangesWithChangelogs });
       }
-      await pushOutgoingChanges(this.centralServer, sessionId, outgoingChanges);
-      await deleteRedundantLocalCopies(modelsForPush, outgoingChanges);
+      await pushOutgoingChanges(this.centralServer, sessionId, outgoingChangesWithChangelogs);
+      await deleteRedundantLocalCopies(modelsForPush, outgoingChangesWithChangelogs);
```
edmofro, line 211:
```suggestion
      await deleteRedundantLocalCopies(modelsForPush, outgoingChanges);
```
Don't think it needs changelogs right?

edmofro (commented):
There's an empty file packages/database/__tests__/utils/audit/attachChangelogToSnapshotRecords.ts

You've done a really good job of pulling out the common bits

Will review more later - haven't touched tests or a couple of important files yet

packages/database/src/utils/audit/insertChangelogRecords.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  const queryInterface = sequelize.getQueryInterface();
+
+  const existingRecords = (await queryInterface.select(
+    null,
+    { tableName: 'changes', schema: 'logs' },
+    {
+      where: {
+        [Op.or]: changelogRecords.map(({ table_name, record_id }) => ({
+          table_name,
```
dannash100, line 23: @edmofro I might be tripping here - but I think we don't want to filter out by table_name, record_id right? - as we expect there to be many immutable change logs for singular records.

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
```diff
@@ -0,0 +1,57 @@
...
+): Promise<SyncSnapshotAttributesWithChangelog[]> => {
+  if (!snapshotRecords.length) {
+    return snapshotRecords;
+  }
+  const changelogRecords = (await sequelize.query(
+    `
+    SELECT * FROM logs.changes
+    WHERE updated_at_sync_tick > :minSourceTick
+    ${maxSourceTick ? 'AND updated_at_sync_tick < :maxSourceTick' : ''}
+    ${tableWhitelist ? `AND table_name IN (:tableWhitelist)` : ''}
```
edmofro, line 28: We don't technically need a whitelist here if we can filter the `recordTypeAndIds` array instead, but I also think it keeps things clear and probably does no harm, maybe even a little good to the query performance

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
```diff
@@ -0,0 +1,57 @@
...
+    AND CONCAT(table_name, '-', record_id) IN (:recordTypeAndIds)
+    `,
+    {
+      type: QueryTypes.SELECT,
+      replacements: {
+        minSourceTick,
+        maxSourceTick,
+        tableWhitelist,
+        recordTypeAndIds: snapshotRecords.map((r) => `${r.recordType}-${r.recordId}`),
```
edmofro, line 37: Suggest we filter this by whitelisted tables

packages/database/src/utils/audit/insertChangelogRecords.ts
edmofro: That's right, this should use normal logic of the changelog primary key (uuid) to manage upsert behaviour

packages/database/src/utils/audit/extractChangelogFromSnapshotRecords.ts
```diff
@@ -0,0 +1,28 @@
+import type {
+  ChangelogRecord,
+  SyncSnapshotAttributes,
+  SyncSnapshotAttributesWithChangelog,
+} from 'types/sync';
+
+export const extractChangelogFromSnapshotRecords = (
```
edmofro, line 7: Gosh it seems a little crazy to be doing attach on the way out and then extract on the way in, just to have a "nice" data format huh. You've done the work now but if I'd thought about the fact that we're spending quite a few extra cpu cycles on all of that, I might of suggested something different...

packages/database/src/utils/audit/insertChangelogRecords.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const existingKeys = existingRecords.map((r) => `${r.table_name}-${r.record_id}`);
+  const recordsToInsert = changelogRecords
+    .filter((r) => !existingKeys.includes(`${r.table_name}-${r.record_id}`))
+    .map(({ record_data, updated_at_sync_tick, ...changelogRecord }) => {
+      return {
+        ...changelogRecord,
+        // TODO Should we he have to do this ?
+        updated_at_sync_tick: isFacility ? -999 : Number(updated_at_sync_tick),
+        record_data: JSON.stringify(record_data),
```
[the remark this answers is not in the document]
dannash100: Yeah seemingly!

packages/database/src/utils/audit/extractChangelogFromSnapshotRecords.ts
dannash100: Yeah i kept thinking that, i tested for performative way to do these iterations tho. hopefully not a biggie. I am open to changing if you think important.

packages/database/src/utils/audit/insertChangelogRecords.ts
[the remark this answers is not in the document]
dannash100: Also without a model bigint is parsed as a string using raw query, but I think as a number if a model. 
Felix's pr includes a model so I might use that

packages/central-server/app/sync/addChangelogRecords.js
dannash100: Yeah it was made obvious by test suite - as it spins up new environment each time and failed.

packages/central-server/app/sync/addChangelogRecords.js
dannash100: Hmm i think you did that bit 🤔 

it was in the original pairing
```
  const minSourceTick = lookupTicks.at(0).sourceStartTick;
  const maxSourceTick = lookupTicks.at(-1).sourceStartTick;
```
Is that right tho 😖

dannash100:
@coderabbitai full review

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -647,7 +666,10 @@ export class CentralSyncManager {
       incomingSnapshotRecordsCount: incomingSnapshotRecords.length,
       sessionId,
     });
-    await insertSnapshotRecords(sequelize, sessionId, incomingSnapshotRecords);
+
+    const { snapshotRecords, changelogRecords } = extractChangelogFromSnapshotRecords(incomingSnapshotRecords);
+    await insertSnapshotRecords(sequelize, sessionId, snapshotRecords);
+    await insertChangelogRecords(sequelize, changelogRecords);
   }
```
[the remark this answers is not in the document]
dannash100: @edmofro keen for look over this one when you rereview - re what kinda transaction this should be?

packages/database/__tests__/utils/audit/insertChangelogRecords.test.ts
```diff
@@ -0,0 +1,222 @@
...
+        created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+        updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
+        deleted_at TIMESTAMP WITH TIME ZONE,
+        updated_at_sync_tick BIGINT NOT NULL,
+        updated_by_user_id TEXT NOT NULL,
+        record_id TEXT NOT NULL,
+        record_update BOOLEAN NOT NULL,
+        record_data JSONB NOT NULL
+      );
+    `);
```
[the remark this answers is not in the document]
dannash100: Ohk ive worked out why this happens - its skipping all .ts migrations

dannash100:
Ohk cool all tests passing and now using models and new data structure 🙏.
Storybook and determinism failures are accounted for - storybook fix is in main, and determinism will fix itself on epic merge - its due to existing migration changes, much like the last 3 merged branches.

packages/database/__tests__/utils/extractChangelogFromSnapshotRecords.test.ts
```diff
@@ -0,0 +1,164 @@
...
+import type { SyncSnapshotAttributesWithChangelog } from '../../src/types/sync';
+import { Attributes } from 'sequelize';
+
+describe('extractChangelogFromSnapshotRecords', () => {
+  it('should extract changelog records from snapshot records', () => {
+    // Arrange
+    const changelogRecords: Attributes<ChangeLog>[] = [
+      {
+        recordId: 'record1',
+        id: 'dogman-1',
```
dannash100, line 14: small note that changelog should be added to fake package and used here

packages/central-server/__tests__/sync/CentralSyncManager.test.js
edmofro: I don't understand why? A single run creates a range where the lower bound is the `sourceStartTick` and the upper bound is `lookupEndTick`

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
```diff
@@ -0,0 +1,58 @@
...
+    : snapshotRecords;
+
+  if (!relevantRecords.length) {
+    return snapshotRecords;
+  }
+
+  const changelogRecords = await models.ChangeLog.findAll({
+    where: {
+      recordSyncTick: {
+        [Op.gt]: minSourceTick,
```
edmofro, line 28: I've realised we actually do want this to be inclusive, not exclusive! Because in other places where we're exclusive, we're saying "we know we've got up to here, get stuff since", whereas here we're saying "we want any changes that happened at the same time as any of these ticks that the source records were saved at"

edmofro: You'll have to remind me where the other place was that we were looking at, in case it's the same scenario

packages/central-server/__tests__/sync/CentralSyncManager.test.js
edmofro: Maybe because of the inclusive/exclusive thing? See my other comment, and once you change that see if you can drop this weirdness

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -433,6 +432,87 @@ describe('CentralSyncManager', () => {
...
+            tableName: 'patient_program_registrations',
+            recordId: patientProgramRegistrationChange.recordId,
+          }),
+          expect.objectContaining({
+            tableName: 'patient_program_registrations',
+            recordId: patientProgramRegistrationChange.recordId,
+          }),
+        ]),
+      );
+    });
```
edmofro, line 515: This is a really nice golden path test...but it's pretty minimal in the scheme of all the combinations of things that can happen I reckon.
- Multiple tables
- Some changes to the same record that will have already been picked up by previous syncs and shouldn't be included
- Some changes to the same record that are later than the lookup table has got up to and shouldn't be included (this is an important case we considered heavily, so should test to make sure we got it right)
- Changes with sync ticks on the very border of needing to be included vs. excluded (I think this is already happening above actually, which is why we're catching that weirdness of needing two lookup table cycles)

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -555,14 +567,21 @@ export class CentralSyncManager {
...
-    return findSyncSnapshotRecords(
+    const session = await this.connectToSession(sessionId);
+    const snapshotRecords = await findSyncSnapshotRecords(
       this.store.sequelize,
       sessionId,
       SYNC_SESSION_DIRECTION.OUTGOING,
       fromId,
       limit,
     );
+    const recordsForPull = await addChangelogRecords(
```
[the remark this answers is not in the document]
edmofro: This creates a chain of code smells I'd be keen to talk about!

packages/database/__tests__/utils/extractChangelogFromSnapshotRecords.test.ts
```diff
@@ -0,0 +1,164 @@
+import { type ChangeLog } from '../../src/models/ChangeLog';
```
edmofro, line 1: Should this test file be under `__tests__/utils/audit` to match folder structure completely?

packages/facility-server/__tests__/audit/attachChangelogToSnapshotRecords.test.js
```diff
@@ -0,0 +1,165 @@
...
+  afterAll(() => ctx.close());
+
+  afterEach(async () => {
+    const { ChangeLog } = models;
+    await ChangeLog.destroy({ where: {} });
+  });
+
+  it('should attach changelog records to snapshot records within the specified tick range', async () => {
+    const { ChangeLog } = models;
+    await Promise.all([
```
edmofro, line 22: This needs to have more negative cases covered, notably including changelog records for record id / table combos that aren't included in the snapshot records (I suspect this isn't catching the `or` that should be an `and`, if I'm right about that)

packages/facility-server/__tests__/audit/insertChangelogRecords.test.js
```diff
@@ -0,0 +1,193 @@
...
+    // Act
+    await insertChangelogRecords(models, changelogRecords);
+
+    // Assert
+    const results = await models.ChangeLog.findAll({
+      order: [['recordId', 'ASC']],
+    });
+    expect(results).toHaveLength(3); // Should have 3 records (existing + 2 new)
+
+    // Should ignore the existing record as changelog records are immutable
```
edmofro, line 104:
```suggestion
    // Should ignore the change to the existing record as changelog records are immutable
```

packages/facility-server/__tests__/audit/insertChangelogRecords.test.js
```diff
@@ -0,0 +1,193 @@
...
+    // Act
+    await insertChangelogRecords(models, changelogRecords, true);
+
+    // Assert
+    const result = await models.ChangeLog.findAll();
+    expect(result).toHaveLength(1);
+    expect(result[0].recordSyncTick).toBe("-999");
+  });
+
+  it('should preserve recordSyncTick for non-facility records', async () => {
```
edmofro, line 140:
```suggestion
  it('should use the current sync tick for non-facility records', async () => {
```

edmofro: Oh right - maybe this isn't true, if there's no trigger on the changelog record table itself. What sync tick do records get when they sync in to the central server? Do they get the one they had on the facility server, or do they get the current sync tick on the central server at the time they sync in? I can't recall.

edmofro (changes requested):
Got some more comments for ya

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
```diff
@@ -0,0 +1,58 @@
...
+        ...(maxSourceTick && { [Op.lt]: maxSourceTick }),
+      },
+      ...(tableWhitelist && {
+        tableName: {
+          [Op.in]: tableWhitelist,
+        },
+      }),
+      [Op.or]: relevantRecords.map(({ recordType, recordId }) =>({
```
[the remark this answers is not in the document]
dannash100: this bit is an or actully!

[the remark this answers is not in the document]
edmofro: Oh really?? That confuses me! Can you explain?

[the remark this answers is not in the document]
dannash100: So its saying - first snapshot records tableName and recordId, or second snapshot records tableName and recordId, or third snapshot records.... and so on and on.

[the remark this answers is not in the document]
dannash100: Plus include all the other stuff

packages/facility-server/__tests__/audit/insertChangelogRecords.test.js
dannash100: Yeah got a bit confused here - might hit u up in office on the last of the tick stuff.

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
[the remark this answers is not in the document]
edmofro: aaaah sorry I get it, the or applies to the array that comes after it, it's not an or against the wider query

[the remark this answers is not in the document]
edmofro: feels like there's a clever way to do this - like make a composite dynamic column (or just index??) on `table_name + '_' + record_id` and then query that with an `IN`

packages/facility-server/__tests__/audit/insertChangelogRecords.test.js
edmofro: Yep all good!

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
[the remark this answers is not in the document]
dannash100: Yeah true that ill take a look at it today!

packages/central-server/app/sync/CentralSyncManager.js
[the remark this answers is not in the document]
dannash100: Sorted this as discussed 🙏

packages/database/src/utils/audit/attachChangelogToSnapshotRecords.ts
dannash100: Oh the getting of the range

```
  const lookupTicks = await SyncLookupTick.findAll({
    where: {
      lookupEndTick: {
        [Op.gt]: pullSince,
        [Op.lt]: pullUntil,
      },
    },
    order: [['lookupEndTick', 'ASC']],
  });
```

edmofro: Great, that one is correct as it is matching the query that we use to find the relevant records in the lookup table when doing the normal snapshot (which is exclusive)

The inclusive and exclusive places could do with comments, considering how easy it has been to forget and get confused!

packages/database/src/utils/audit/insertChangelogRecords.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  const existingIds = existingRecords.map(({ id }) => id);
+  const recordsToInsert = changelogRecords
+    .filter(({ id }) => !existingIds.includes(id))
+    .map(({ recordData, recordSyncTick, ...changelogRecord }) => {
+      return {
+        ...changelogRecord,
+        // TODO Should we he have to do this ?
+        recordSyncTick: isFacility ? -999 : Number(recordSyncTick),
```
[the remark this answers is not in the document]
dannash100: just realised tho - there is no trigger on changes to update the recordSyncTick if set 🤔 Might need sum pairing on last of this stuff

packages/database/src/types/sync.ts
dannash100: Yeah its resolved weird tho it show up still ay

dannash100: Ohh my bad nah its still in there ill remove

packages/central-server/__tests__/sync/CentralSyncManager.test.js
dannash100: Doing these test cases now

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -169,6 +167,11 @@ describe('CentralSyncManager', () => {
       displayName: 'System',
       role: 'system',
     });
+    // Add these lines to ensure proper cleanup
```
edmofro, line 170: "Add these lines" sounds like an AI agent comment right? Prob unnecessary

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -433,6 +437,384 @@ describe('CentralSyncManager', () => {
...
+            recordId: patientProgramRegistrationChange.recordId,
+          }),
+          expect.objectContaining({
+            tableName: 'patient_program_registrations',
+            recordId: patientProgramRegistrationChange.recordId,
+          }),
+        ]),
+      );
+    });
+    // TODO: I don't think this setup is correct, make sure testing this is correct
```
edmofro, line 517: 👀

packages/central-server/__tests__/sync/CentralSyncManager.test.js
edmofro: Yeah I think this is a good thing to add a cursor rule around ay, it's just a dumb thing for it to do
Tbh I am still not really confident that cursor rules are consistently adhered to though, but try

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -433,6 +437,384 @@ describe('CentralSyncManager', () => {
...
+        },
+        () => true,
+      );
+
+      const outgoingChanges = await centralSyncManager.getOutgoingChanges(sessionId, {});
+      const patientProgramRegistrationChange = outgoingChanges.find(
+        (c) => c.recordType === 'patient_program_registrations',
+      );
+
+      // Verify that the change at the boundary tick is included
```
edmofro, line 783: Oh man it's so hard to keep track of when it should be inclusive vs. exclusive, but I would have expected this to be exclusive!

packages/central-server/__tests__/sync/CentralSyncManager.test.js
edmofro: I think the setup is correct from what I can tell! Could probably go harder but this is already really good, I'm happy 😊

packages/database/src/migrations/123_createTriggerFunctionForUpdatedAtSyncTick.js
```diff
@@ -7,13 +9,13 @@ export async function up(query) {
...
+        -- If setting to "${SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER}" representing "not marked as updated on this client", we actually use
+        -- a different number, "${SYNC_TICK_FLAGS.LAST_UPDATED_ELSEWHERE}", to represent that in the db, so that we can identify the
+        -- difference between explicitly setting it to not marked as updated (when NEW contains ${SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER}),
         -- and having other fields updated without the updated_at_sync_tick being altered (in which
-        -- case NEW will contain -999, easily distinguished from -1)
-        IF NEW.updated_at_sync_tick = -1 THEN
-          NEW.updated_at_sync_tick := -999;
+        -- case NEW will contain ${SYNC_TICK_FLAGS.LAST_UPDATED_ELSEWHERE}, easily distinguished from ${SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER})
+        IF NEW.updated_at_sync_tick = ${SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER} THEN
+          NEW.updated_at_sync_tick := ${SYNC_TICK_FLAGS.LAST_UPDATED_ELSEWHERE};
```
edmofro, line 18: We shouldn't edit migrations after they've been deployed

packages/database/src/migrations/125_createFacilityPatientSyncTable.js
```diff
@@ -1,6 +1,7 @@
 import config from 'config';
 import Sequelize from 'sequelize';
 import { selectFacilityIds } from '@tamanu/utils/selectFacilityIds';
+import { SYNC_TICK_FLAGS } from '@tamanu/database/sync';
```
edmofro, line 4: Same goes throughout migrations

packages/database/src/services/migrations/migrationHooks.ts
```diff
@@ -131,11 +132,11 @@ export async function runPreMigration(log: Logger, sequelize: Sequelize) {
...
 
 export async function runPostMigration(log: Logger, sequelize: Sequelize) {
-  // add column: holds last update tick, default to -999 (not marked for sync) on facility,
-  // and 0 (will be caught in any initial sync) on central server
+  // add column: holds last update tick, default to 0 (will be caught in any initial sync) on central server
+  // and SYNC_TICK_FLAGS.UPDATED_ELSEWHERE (not marked for sync) on facility
   // triggers will overwrite the default for future data, but this works for existing data
   const isFacilityServer = !!selectFacilityIds(config);
-  const initialValue = isFacilityServer ? -999 : 0; // -999 on facility, 0 on central server
+  const initialValue = isFacilityServer ? SYNC_TICK_FLAGS.LAST_UPDATED_ELSEWHERE : SYNC_TICK_FLAGS.OVERWRITE_WITH_CURRENT_TICK;
```
edmofro, line 139: I don't think this is a case of it getting overwritten, I think it literally takes 0 because in migrations updated at sync tick setting triggers are turned off

packages/database/src/sync/constants.ts
```diff
@@ -11,3 +11,12 @@ export const COLUMNS_EXCLUDED_FROM_SYNC = [
...
 export const SYNC_LOOKUP_PENDING_UPDATE_FLAG = -1;
+
+export const SYNC_TICK_FLAGS = {
+  INCOMING_FROM_CENTRAL_SERVER: -1,
+  LAST_UPDATED_ELSEWHERE: -999,
+  LOOKUP_PENDING_UPDATE: -2,
+  OVERWRITE_WITH_CURRENT_TICK: 0,
+};
+
+export const SYNC_CHANGELOG_TO_FACILITY_FOR_THESE_TABLES = ['patient_program_registrations'];
```
edmofro, line 22: Oh woah, we've already restricted it to PPR syncing down only? 

I didn't realise that. But also, imo this is a good place for inversion of control. Throughout sync we define rules at the model level, rather than having centralised sets of rules.

edmofro (changes requested):

packages/central-server/app/sync/getLookupSourceTickRange.js
```diff
@@ -0,0 +1,33 @@
...
+export const getLookupSourceTickRange = async (
+  { models },
+  pullSince,
+  pullUntil,
+) => {
+  const { SyncLookupTick } = models;
+  const lookupTicks = await SyncLookupTick.findAll({
+    where: {
+      lookupEndTick: {
+        [Op.gt]: pullSince,
```
edmofro, line 18: @passcod potential place here, alternatives include
- sourceStartTick > pullSince
- lookupEndTick >= pullSince
- sourceStartTick >= pullSince
Just listing out all the possibilities so I can think them through

packages/central-server/app/sync/getLookupSourceTickRange.js
```diff
@@ -0,0 +1,33 @@
...
+        [Op.lt]: pullUntil,
```
edmofro, line 19: Another potential @passcod (tagging you so I can find them easily as much as anything)
```suggestion
        [Op.lte]: pullUntil,
```

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -262,13 +265,22 @@ export class CentralSyncManager {
       await repeatableReadTransaction(store.sequelize, async (transaction) => {
         // do not need to update pending records when it is initial build
         // because it uses ticks from the actual tables for updated_at_sync_tick
-        if (!isInitialBuildOfLookupTable) {
+        if (isInitialBuildOfLookupTable) {
+          await this.store.models.SyncLookupTick.create({
+            sourceStartTick: previouslyUpToTick,
+            lookupEndTick: currentTick,
+          });
```
edmofro, line 272: @passcod can you make sure `updateLookupTable` has been called once before you start any data generation? Just in case it's an initial build difference catching some of the data in a way we haven't dealt with.

passcod: I think the scheduler has it running every 20 seconds? If so it was definitely run. I can do extra ones manually tho

edmofro: You mean it definitely was run prior to starting any of the data generation right? Because you wouldn't have kicked off the data generation / testing stuff within 20 seconds of the deployment coming up anyway

packages/central-server/app/sync/getLookupSourceTickRange.js
edmofro: @dannash100 the most generous of these would be `sourceStartTick >= pullSince`, chuck that in your testing and see if things get better. If not, we know pretty confidently the issue is not in this function.

packages/database/src/migrations/123_createTriggerFunctionForUpdatedAtSyncTick.js
dannash100: true that ill roll these back

packages/database/src/services/migrations/migrationHooks.ts
dannash100: ahh yeah true ohk! weird that i changed this 🤔

dannash100: Ohh because its 0 also i see, ill just leave as 0 here

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -555,14 +567,31 @@ export class CentralSyncManager {
...
       sessionId,
       SYNC_SESSION_DIRECTION.OUTGOING,
       fromId,
       limit,
     );
+    const sourceTickRange = await getLookupSourceTickRange(
+      this.store,
+      session.pullSince,
+      session.pullUntil
+    );
```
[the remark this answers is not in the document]
dannash100: yeah this is a great call

PR #7567, opened 2025-04-22
fix(translations): SAV-953: Fix enum translations

tcaiger:
### Changes

- Update translated enums to be camel case everywhere that they are used.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/constants/src/enumRegistry.ts
```diff
@@ -173,9 +173,9 @@ export const prefixMap = new Map(
 export const enumTranslations = (Object.entries(registeredEnums) as EnumEntries).flatMap(
   ([key, value]) =>
     Object.entries(value).map(([enumKey, enumValue]) => [
-      `${translationPrefixes[key]}.${enumKey
-        .toLowerCase()
```
tcaiger, line 177: this is just removing `.toLowerCase()`. This causes an issue where keys are saved in the database in a different case to the values from the enums. eg. This is saved in the database `programRegistry.property.registrationStatus.recordedinerror` as a string_id but the value from the enum is `recordedInError` so the translation is not found when using the value.

tcodling (approved):
lgtm!

packages/constants/src/enumRegistry.ts
dannash100: Oh yes no this was not entirely intentional,
I wanted to convert
```COMPLEX_CHART_CORE``` to `complexChartCore` Ideally

Comment (tcaiger) [reply packages/constants/src/enumRegistry.ts:177]: I'm not sure that my fix is actually correct - I will test a bit more and have tweak it.

Comment (dannash100) [reply packages/constants/src/enumRegistry.ts:177]: Ohk sick thank u! We need to do without loads if possible to keep constants from needing to be installed yarn to scrape

Comment (tcaiger) [reply packages/constants/src/enumRegistry.ts:177]: oh yeah that makes sense . The transform was actually fine in the end, the issue was just that the keys in TranslatedEnum were not being translated to match. It is still true that the keys don't always come out as camel case but it actually doesn't matter as long as the transform is consistent in the scraper and the front end.

File: packages/constants/src/enumRegistry.ts [hunk 2]
```diff
@@ -169,13 +169,21 @@ export const prefixMap = new Map(
   ]),
 );
 
+/**
+ * Converts a string from formats like SNAKE_CASE to camelCase
+ * Keep in sync with packages/shared/src/utils/enumRegistry.js
+ * @param {string} value - The string to convert
+ * @returns {string} The converted string in camelCase
+ */
+const toCamelCase = (value: string): string => {
```
tcaiger, line 178: It was a bit of a shame that I had to redefine this function again here as well as in utils but the constants package doesn't have utils installed as a dependency so I had to do it.

PR #7583, opened 2025-04-28
feat(charting): SAV-698: Charts and graphs updates

NavarroEmilioLuis:
### Changes

Leftover stuff that wasn't really working plus tweaks

## Summary by CodeRabbit

- **New Features**
  - Introduced unified endpoints and logic for retrieving both vital and chart data, enabling more flexible data visualization options.
  - Added support for context-aware chart rendering, allowing components to distinguish between vital signs and other chart data for tailored display and messaging.

- **Refactor**
  - Consolidated and generalized data fetching hooks and chart component logic to streamline support for multiple graph types.
  - Updated chart and table components to use new props and context values for improved modularity and clarity.

- **Bug Fixes**
  - Adjusted multi-chart view button to appear only when multiple chart options are available.

- **Documentation**
  - Improved no-data messages to reflect whether vitals or other chart data are being displayed.


packages/web/app/views/charts/ChartLineChart.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <LineChart
+      chartData={chartData}
+      visualisationConfig={visualisationConfig}
+      isLoading={isLoading}
+      chartProps={chartProps}
+    />
+  );
+};
```
NavarroEmilioLuis, line 25: This component is almost a dupe of `VitalLineChart`, but we have to know what argument to pass to the `useGraphDataQuery` function (whether it's a vital or a chart). I decided this was better rather than doing like a factory component that injects the query and create two queries that are identical

rohan-bes (approved):
Tbh it does feel a little awkward passing around the `isVital` parameter to handle the conditional logic all over the place. But I can see how things are similar enough that it's not worth breaking out the abstraction.

Maybe I'd suggest we could have two separate API endpoints on the facility server, which just check the permissions and then pass in the relevant `dateDataElementId` to some shared function? Just so that we could keep the `isVital` concept to the frontend.

But I don't feel too strongly

packages/web/app/views/charts/ChartLineChart.jsx
NavarroEmilioLuis: How so?

packages/facility-server/app/routes/apiv1/encounter.js
```diff
@@ -567,6 +572,19 @@ encounterRelations.get(
...
+  asyncHandler(async (req, res) => {
+    req.checkPermission('list', 'Charting');
+    const data = await getGraphData(req, CHARTING_DATA_ELEMENT_IDS.dateRecorded);
+
+    res.send({
+      count: data.length,
+      data,
+    });
+  }),
+);
```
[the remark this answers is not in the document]
NavarroEmilioLuis: Not true: directory is set depending on the `isVital` function parameter

PR #7599, opened 2025-05-01
fix(storybook): Fix storybook

NavarroEmilioLuis:
### Changes

So in #7575 this log was introduced and it started to fail CI. It's very weird and am not sure why it's happening but I talked to Rohan and he's okay to ditch it anyway. So we get this out of the way sooner than later.

## Summary by CodeRabbit

- **Refactor**
	- Removed internal warning log messages when certain reference data is missing. No changes to visible functionality.


dannash100 (approved):
I found this too but no idea why its in scope on storybook tests

NavarroEmilioLuis:
> I found this too but no idea why its in scope on storybook tests

God only knows

PR #7610, opened 2025-05-05
feat(api): SAV-938: Allow any client name in the X-Tamanu-Client header

rohan-bes:
### Changes

Opting to just remove the check to ensure that the `X-Tamanu-Client` is a known client globally. The reasoning for doing this is that it's never been a secure mechanism and we notify clients what the supported names were anyway.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **Refactor**
  - Updated version compatibility checks to allow unrecognized client types to proceed without version constraints, instead of returning an error.
  - Removed support for several legacy client types.
- **Documentation**
  - Updated mobile app README to reflect the renamed version control constant for better clarity.

### DIFF
--- a/packages/central-server/app/middleware/versionCompatibility.js
+++ b/packages/central-server/app/middleware/versionCompatibility.js
@@ -25,7 +25,7 @@ export const MIN_CLIENT_VERSION = MIN_CLIENT_OVERRIDE ?? `${major}.${minor}.0`;
 export const MAX_CLIENT_VERSION = `${major}.${minor}.999`;
 // Note that .999 is only for clarity; higher patch versions will always be allowed
 
-export const SUPPORTED_CLIENT_VERSIONS = {
+export const VERSION_CONTROLLED_CLIENTS = {
   [SERVER_TYPES.FACILITY]: {
     min: MIN_CLIENT_VERSION,
     max: MAX_CLIENT_VERSION,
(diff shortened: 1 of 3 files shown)


packages/central-server/app/middleware/versionCompatibility.js
```diff
@@ -70,19 +49,14 @@ export const versionCompatibility = (req, res, next) => {
...
-  if (!clientTypes.includes(clientType)) {
-    next(
-      new InvalidClientHeadersError(
-        `The only supported X-Tamanu-Client values are ${clientTypes.join(', ')}`,
-      ),
-    );
+  const clientInfo = VERSION_CONTROLLED_CLIENTS[clientType];
+  if (!clientInfo) {
+    // a non version controlled client; ignore version checking
+    next();
```
rohan-bes, line 55: I think is more in the spirit of what this check is doing. There are only specific clients where we apply version controlling to ensure we're not accidentally connecting to the wrong version. These are just internal really, facility/mobile. So we shouldn't be doing version checking on other clients.

edmofro: Yeah agreed, beautiful

edmofro (approved):
Love it

packages/central-server/app/middleware/versionCompatibility.js
edmofro: Really like that variable name too

PR #7617, opened 2025-05-05
fix(translations): HOTFIX 2.30: Fix broken translated suggester logic

tcodling:
### Changes

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/__tests__/apiv1/Suggestions.test.js
+++ b/packages/facility-server/__tests__/apiv1/Suggestions.test.js
@@ -644,6 +644,37 @@ describe('Suggestions', () => {
     expect(khmerRecord.name).toEqual(KHMER_LABEL);
   });
 
+  it('should only search against translated names if they exist', async () => {
+    const { TranslatedString, ReferenceData } = models;
+
+    await ReferenceData.create({
+      id: 'test-drug',
+      type: 'drug',
+      name: 'banana',
+    });
+
+    await TranslatedString.create({
+      stringId: `${REFERENCE_DATA_TRANSLATION_PREFIX}.drug.test-drug`,
+      text: 'apple',
+      language: 'en',
+    });
+
+    // Check that the translated label can be matched through search
+    const result1 = await userApp.get(`/api/suggestions/drug?language=en&q=apple`);
+    expect(result1.body.length).toEqual(1);
+    expect(result1.body[0].name).toEqual('apple');
+
+    // Check that the original label is not matched since a translated label exists
+    const result2 = await userApp.get(`/api/suggestions/drug?language=en&q=banana`);
+    expect(result2.body.length).toEqual(0);
+
+    // Drop the translation and check that the original label is matched
+    await TranslatedString.truncate({ cascade: true, force: true });
+    const result3 = await userApp.get(`/api/suggestions/drug?language=en&q=banana`);
+    expect(result3.body.length).toEqual(1);
+    expect(result3.body[0].name).toEqual('banana');
+  });
+
   it.skip('should return translated labels for the correct facility', async () => {
     const { TranslatedString } = models;
 
--- a/packages/facility-server/__tests__/apiv1/Suggestions.test.js
+++ b/packages/facility-server/__tests__/apiv1/Suggestions.test.js
@@ -651,6 +651,7 @@ describe('Suggestions', () => {
       id: 'test-drug',
       type: 'drug',
       name: 'banana',
+      code: 'test-drug',
     });
 
     await TranslatedString.create({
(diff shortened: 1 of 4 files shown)


packages/facility-server/app/routes/apiv1/suggestions.js
tcodling, line 1: Here i have changed up the translation flow quite a bit for the 3 suggester routes that were using it

Previously it would fetch the translations and records seperately and then map over the records assigning the proper translated name, now we do all this within one query that includes a join.

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -65,34 +60,44 @@ function createSuggesterRoute(
...
+      const where =
+        isTranslatable && hasTranslations
+          ? Sequelize.literal(`EXISTS (
+            SELECT 1 
+            FROM translated_strings 
+            WHERE language = '${language}'
+            AND string_id = '${translationPrefix}' || "${modelName}"."id"
+            AND text ILIKE '%${searchQuery}%'
+          )`)
+          : whereBuilder(`%${searchQuery}%`, query, req);
```
tcodling, line 99: These are the two significant blocks of new code. 

Attributes joins to the translated string table and attaches a translated_name property that is sent to the front end instead of the usual "name" associated with the reference data

The where we match the users inputted searchQuery to the text of the translated strings.

tcodling: This is definitley a weakness of mine and I found it very difficult to get this looking clean and in a way that I am happy with. Totally open to new ideas/pairing with reviewer 🙏

NavarroEmilioLuis (commented):
Changes are hard to follow, I'll keep on reading but I wanted to ask what was broken and how is this meant to solve the issue?

Right off the bat I can see some imports not needed on some files and using template strings instead of replacements to pass to queries which isn't great. Also I don't think translations only applied to three routes, maybe just three suggester types?

PS: Also, some unit tests broke because of this 🤔

tcodling:
@NavarroEmilioLuis no huge rush on this if you are done for today 🙏

packages/facility-server/__tests__/apiv1/Suggestions.test.js
```diff
@@ -644,6 +644,37 @@ describe('Suggestions', () => {
     expect(khmerRecord.name).toEqual(KHMER_LABEL);
   });
 
+  it('should only search against translated names if they exist', async () => {
+    const { TranslatedString, ReferenceData } = models;
+
+    await ReferenceData.create({
+      id: 'test-drug',
+      type: 'drug',
+      name: 'banana',
```
NavarroEmilioLuis, line 653: Seems we need to add `code` too?

packages/facility-server/__tests__/apiv1/Suggestions.test.js
```diff
@@ -644,6 +644,37 @@ describe('Suggestions', () => {
...
+
+    // Check that the original label is not matched since a translated label exists
+    const result2 = await userApp.get(`/api/suggestions/drug?language=en&q=banana`);
+    expect(result2.body.length).toEqual(0);
+
+    // Drop the translation and check that the original label is matched
+    await TranslatedString.truncate({ cascade: true, force: true });
+    const result3 = await userApp.get(`/api/suggestions/drug?language=en&q=banana`);
+    expect(result3.body.length).toEqual(1);
+    expect(result3.body[0].name).toEqual('banana');
```
NavarroEmilioLuis, line 675: Oh interesting!

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -65,34 +63,44 @@ function createSuggesterRoute(
...
+              SELECT "text" 
+              FROM "translated_strings" 
+              WHERE "language" = :language
+              AND "string_id" = '${translationPrefix}' || "${modelName}"."id"
+              LIMIT 1
+            )`),
+            'translated_name',
+          ],
         ],
       };
```
NavarroEmilioLuis, line 91: Well given that we reuse this bit exactly maybe we can make it a common return from a function. Then specify in the query: `attributes: getTranslationAttributes();`

Maybe go even further to pass in the endpoint and return the `attributes` and the `translationPrefix` as well. Essentially just draw from one source as much as possible, should a change happen at any point

NavarroEmilioLuis (changes requested):
Thanks for checking that out. I don't really know much about the implications of the sequelize literal inside attributes, I'm guessing it's possibly a subquery at the select level 🤔 

Aside from that, things check out, the test is still breaking but might just need that `code` added and that's it. I also left one suggestion and happy to move on after that 🙏

packages/facility-server/app/routes/apiv1/suggestions.js
tcodling: yep great call actually. I was feeling like there were just so many of these helper functions in the file i was finding it hard to follow so I didn't want to add another 😆 But agreed this would be way better actually

NavarroEmilioLuis: Yeah it's obnoxious but for our sake better to keep at one place tidy

NavarroEmilioLuis (approved):
LGTM!

PR #7619, opened 2025-05-06
fix(tests): no-issue: Fix intermittent test failures due to invalid fake AppointmentSchedule fields

rohan-bes:
### Changes

Started noticing intermittent failures with this test. I think we've been initialising fake appointment schedules in a way that might not always be valid.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured that generated fake data for appointment schedules now explicitly includes the frequency field.

### DIFF
--- a/packages/fake-data/src/fake/fake.ts
+++ b/packages/fake-data/src/fake/fake.ts
@@ -409,6 +409,7 @@ const MODEL_SPECIFIC_OVERRIDES = {
     const frequency = chance.pickone(REPEAT_FREQUENCY_VALUES);
     const endsMode = chance.pickone(['on', 'after']);
     return {
+      frequency,
       daysOfWeek: [chance.pickone(DAYS_OF_WEEK)],
       nthWeekday:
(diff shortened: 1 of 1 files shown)


packages/fake-data/src/fake/fake.ts
```diff
@@ -409,6 +409,7 @@ const MODEL_SPECIFIC_OVERRIDES = {
     const frequency = chance.pickone(REPEAT_FREQUENCY_VALUES);
     const endsMode = chance.pickone(['on', 'after']);
     return {
+      frequency,
```
rohan-bes, line 412: I think the issue is that depending on the frequency we need to define `nthWeekday`. We do that below however we weren't setting the frequency explicitly so there's a chance the two would be compatible.

dannash100 (approved):
yayer thank you i wondered why this started up

PR #7632, opened 2025-05-12
fix(patientMerge): SAV-866: Update program registry on patient merge

NavarroEmilioLuis:
### Changes

The new design asks for the ID change (to have it be a unique patient and program registry ID). Because of this, when merging patients we will get an issue because of duplicate IDs. I think the easiest way to circumvent this is simply to hard delete the unwanted records.

Previously on the patient merge fix it was decided that all data from the unwanted patient (status history and conditions) should be soft deleted. Now, because of the logs table, we should be confident hard deleting because we do have a trail anyway so we're not losing data.

Note that the first commit was cherry-picking the fix changes with solved conflicts.


packages/scripts/src/testDeterminism.ts
```diff
@@ -58,7 +61,7 @@ async function hashTables(sequelize: Sequelize, tables: string[]): Promise<Table
 
   const hashes: TableHashes = new Map();
   for (const table of tables) {
-    await sequelize.query(`CREATE TEMPORARY TABLE determinism_check_table AS TABLE "${table}"`);
+    await sequelize.query(`CREATE TEMPORARY TABLE determinism_check_table AS TABLE ${table}`);
```
passcod, line 64: What? why?

passcod: Oh I see

NavarroEmilioLuis: Just FYI I've put this into a separate PR to see if it unblocks the error on epic branch:

https://github.com/beyondessential/tamanu/pull/7655

error: https://github.com/beyondessential/tamanu/actions/runs/15031453481/job/42244640058

rohan-bes (approved):
I might've had some questions about why we're permanently enforcing that the id's are the combination of patient_id and registry_id but I feel that decision has probably been well made.

It's looking good to me :)

packages/database/src/migrations/1747097003208-changePatientProgramRegistrationIdColumn.ts
```diff
@@ -0,0 +1,93 @@
...
+  await query.sequelize.query(`
+    ALTER TABLE patient_program_registrations RENAME COLUMN new_id TO id;
+    ALTER TABLE patient_program_registration_conditions RENAME COLUMN patient_program_registration_new_id TO patient_program_registration_id;
+  `);
+
+  // Add constraints to id column, set new primary key and bring back foreign key constraint
+  await query.sequelize.query(`
+    ALTER TABLE patient_program_registrations ADD CONSTRAINT patient_program_registrations_id_key UNIQUE (id);
+    ALTER TABLE patient_program_registrations ALTER COLUMN id SET NOT NULL;
+    ALTER TABLE patient_program_registrations ADD PRIMARY KEY (patient_id, program_registry_id);
```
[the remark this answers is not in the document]
NavarroEmilioLuis: Ah yep so on both topics:

So the decision of enforcing to be a composite ID from patient_id and program_registry_id comes from the design that their combination should be unique, allowing it to be deterministic to avoid duplication, instead, an indirect merge (overwrite) will happen if it happens, with the benefit that the registration doesn't store critical information (plus, if it's duplicate, it should be more or less the same - which is what happens with similar records with generated columns anyway).

Now the reason we're not using `id` as the primary key is to mimic the `patient_facilities` table. It's just a pattern that already exists in our codebase. Originally I was using it as a primary key, but Edwin suggested that we should instead go this route, as we don't want to have a generated column as the primary key itself.

Both are fairly gray area, but the bigger picture is that it should be the safest pattern route as we already have precedents!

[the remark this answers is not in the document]
rohan-bes: Ah thanks for clearing those up :pray:

PR #7652, opened 2025-05-14
KAM-377: audit sync testing

passcod:


passcod:
@coderabbitai pause

packages/central-server/config/default.json5
```diff
@@ -252,6 +252,7 @@
     },
     "syncLookupRefresher": {
       "enabled": true,
+      // every 3 seconds?
```
edmofro: Good idea from @passcod to see if this lowers the rate of data loss @dannash100

passcod: no no, this was me misreading the crontab expression, not a suggestion

edmofro: oh right - would be interesting to see if it changed with a faster lookup table population period, so I still think it is a good suggestion haha

dannash100:
We can close this now 🙏

PR #7654, opened 2025-05-14
fix(programRegistry): SAV-861: Ensure determinism in migration

NavarroEmilioLuis:
### Changes

Found out about this change when working on another card. However it's weird that CI passed on the PR this migration was created, see #7537. While the last commit didn't run migrations steps, previous ones did and passed. I'm not completely sure as to why this wasn't picked up.

The thing here would be that the migration is going to run on both servers, and if they don't have the same ID, we will have duplicate logs for all inserted records


tcaiger (approved):
Good find - the fix makes sense and yeah a bit confusing why it passed the first times. 🤔

PR #7655, opened 2025-05-14
fix(programRegistry): Unblock deterministic migration issue

NavarroEmilioLuis:
### Changes

Yeah okay I'm not completely sure as to why, but I think this doesn't hurt

### DIFF
--- a/packages/scripts/src/testDeterminism.ts
+++ b/packages/scripts/src/testDeterminism.ts
@@ -34,7 +34,7 @@ async function listTables(sequelize: Sequelize): Promise<string[]> {
       .filter((model) => model.syncDirection !== SYNC_DIRECTIONS.DO_NOT_SYNC)
       .map((model) => {
         const schema = (model as any).options?.schema || 'public';
-        return `${schema}.${model.tableName}`;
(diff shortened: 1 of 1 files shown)


packages/scripts/src/testDeterminism.ts
```diff
@@ -58,7 +61,7 @@ async function hashTables(sequelize: Sequelize, tables: string[]): Promise<Table
 
   const hashes: TableHashes = new Map();
   for (const table of tables) {
-    await sequelize.query(`CREATE TEMPORARY TABLE determinism_check_table AS TABLE "${table}"`);
+    await sequelize.query(`CREATE TEMPORARY TABLE determinism_check_table AS TABLE ${table}`);
```
rohan-bes, line 64: Is removing the double quotes here fixing the issue? In general I've felt using double quotes to be safer than not in postgres

rohan-bes: Oh wait now I read the code properly I see it gets wrapped earlier, all good :+1:

rohan-bes (approved):

PR #7659, opened 2025-05-15
fix(programRegistry): SAV-861: Fix migration value

NavarroEmilioLuis:
### Changes

Sneaky. So essentially I got it backwards on the other one. One possible review would be to flip the subquery to be `is_update` instead of `is_insert` and flip the cases, but this feels safer at this point


tcaiger (approved):
Ah yeah this makes more sense. Shouldn't have a material impact on the data but good to correct

PR #7660, opened 2025-05-15
fix(programRegistry): SAV-861: Save latest registration as record_id

NavarroEmilioLuis:
### Changes

changes... lots of changes... logs.changes...

Okay so we realized that previous registrations have all different IDs, however, the way we would read registration history now would be to check for the `record_id` from `logs.changes` table - so when inserting these records we need to grab the latest registration ID.

### DIFF
--- a/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
+++ b/packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
@@ -78,7 +78,7 @@ export async function up(query: QueryInterface): Promise<void> {
       CASE WHEN ppr.deleted_at IS NOT NULL THEN now() ELSE NULL END,
       ${changesHasUpdatedAtSyncTick ? updatedAtSyncTickSelect : ''}
       COALESCE(ppr.clinician_id::text, '${SYSTEM_USER_UUID}'),
-      ppr.id,
+      latest_registrations.latest_registration_id,
       NOT registration_summary.is_insert,
       ppr.created_at,
       ppr.updated_at,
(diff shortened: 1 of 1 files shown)


tcaiger (approved):
Nice! (Just saw Rohan caught this too so happy to have a double review on this one)

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
```diff
@@ -98,7 +98,16 @@ export async function up(query: QueryInterface): Promise<void> {
...
+    ) registration_summary ON ppr.id = registration_summary.id
+    JOIN (
+      SELECT DISTINCT ON (patient_id, program_registry_id)
+        patient_id,
+        program_registry_id,
+        id as latest_registration_id
+      FROM patient_program_registrations
+      WHERE is_most_recent = TRUE
+      ORDER BY patient_id, program_registry_id, date DESC
+    ) latest_registrations ON ppr.patient_id = latest_registrations.patient_id AND ppr.program_registry_id = latest_registrations.program_registry_id
```
rohan-bes, line 110: Just confirming that the `SELECT DISTINCT ON()` is just being defensive here? Like there should only be one record with `is_most_recent = TRUE` per patient/program registry pair. And we should just be getting one row per pair out of this so that we don't expand the JOIN.

rohan-bes (approved):
Took me a little while to wrap my head around this but looks good I think!

packages/database/src/migrations/1744234388450-movePatientProgramRegistrationsToAuditTable.ts
NavarroEmilioLuis: Yep exactly that! The defensiveness is because there might be lingering unresolved merge conflicts here and there, and that is the strategy that we use essentially (is_most_recent true and latest date)

PR #7665, opened 2025-05-15
refactor(audit): Only calculate audit chages source tick range once per sync session

dannash100:
### Changes

- Use modal for audit sync behavior - I dont love this - mabye just flag?
- Only calculate source tick range once and add it to session

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -359,8 +365,11 @@ export class CentralSyncManager {
 
       await this.waitForPendingEdits(tick);
 
+      const { minSourceTick, maxSourceTick } =
+        (await getLookupSourceTickRange(this.store, since, tick)) || {};
```
edmofro, line 369: For some reason I really don't love a bracketed await. Is there a time when `getLookupSourceTickRange` returns something other than an object? Can we fix it there instead of having this ternary?

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -574,19 +583,15 @@ export class CentralSyncManager {
...
       limit,
     );
-    const sourceTickRange = await getLookupSourceTickRange(
-      this.store,
-      session.pullSince,
-      session.pullUntil
-    );
-    if (!sourceTickRange) {
+    const { minSourceTick, maxSourceTick } = session;
+    if (!minSourceTick && !maxSourceTick) {
```
edmofro, line 587:
```suggestion
    if (!minSourceTick || !maxSourceTick) {
```

edmofro (changes requested):
Good change imo, also means that we can see what source ticks it's using from the sync sessions table during a sync if we ever need

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -359,6 +365,8 @@ export class CentralSyncManager {
 
       await this.waitForPendingEdits(tick);
 
+      const lookupTickRange = await getLookupSourceTickRange(this.store, since, tick)
```
edmofro, line 368: Real nit but because there's just two things here, I kinda like getting it in full rather than having the opaque spread below
```suggestion
      const { minSourceTick, maxSourceTick } = await getLookupSourceTickRange(this.store, since, tick)
```

_or_ (and definitely equally happy with this) having it stored under the key `lookupTickRange`. Just don't love the spread of what is an invisible set of properties below.

edmofro (approved):
Just one nit but it's pretty subjective so will approve and leave it up to you

packages/central-server/app/sync/CentralSyncManager.js
dannash100: Yeah true it was to get around that inline await due to having to default it sumway or another might be a cleaner way tho

dannash100: I could return {} early in the getLookupSourceTickRange thats the other option - but wans't sure

PR #7674, opened 2025-05-18
fix(auth): Fix token set to null

flynncs:
### Changes

This change should fix both E2E test runs targeting deployments, alongside a bug that occurred when refreshing deployed Tamanu instances - this would cause you to log out instead of remaining authenticated.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/index.js
+++ b/packages/web/app/index.js
@@ -17,6 +17,7 @@ import './fonts.css';
 function initPersistor(api, store) {
   const persistor = persistStore(store, null, () => {
     const { auth } = store.getState();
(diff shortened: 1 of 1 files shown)


packages/web/app/index.js
```diff
@@ -17,7 +17,9 @@ import './fonts.css';
 function initPersistor(api, store) {
   const persistor = persistStore(store, null, () => {
     const { auth } = store.getState();
-    api.setToken(auth.token);
+    if (auth.token) {
+      api.setToken(auth.token);
+    }
```
flynncs, line 22: not sure if we even need to set the auth token here, given that `restoreSession` should be called after this.
left it just in case though

dannash100 (approved):
good catch!

PR #7691, opened 2025-05-21
fix(sync): don't lock sync_sessions to write debug info

passcod:
### Changes

Fixes for sync database contention encountered on support (BESTA-0016).

When we snapshot, we take a REPEATABLE READ transaction to ensure consistency, but that comes with an implicit contract to never write to shared resources, only read, otherwise locks can be taken and block the rest of the server. In Samoa, we found that a write had slipped through by way of recording the syncLookup toggle to `sync_sessions.debug_info` just before we run the snapshot... which happens within the transaction. In high contention scenarios, this can result in that snapshot transaction blocking one, and via update cascading, ultimately blocking all other updates to the sync_sessions table, which leads to Postgres connection exhaustion.

This PR:
- adds `SKIP LOCKED` to the `addDebugInfo` function, which means that we no longer block on locked session rows just to write some debug data. We log (at debug level) to the console to avoid altogether losing debug data.
- adds a new `setParameters` function which does the same as `addDebugInfo` but explicitly writes data, and is also optimised to a single `UPDATE` statement (which may avoid some locks as the Postgres engine can re-apply the modification via compare-and-swap itself rather than relying on JS logic for the data change).
- moves the specific syncLookup `addDebugInfo` from within the transaction to happen outside of it. Because we're enabling syncLookup permanently anyway soon, it doesn't really matter, but let's still preserve that bit for now. (Also because this fix is going to be backported to 2.16.)
- change that `addDebugInfo` update to a `setParameters` call

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability of synchronization sessions by making debug information updates non-blocking, reducing the chance of sync interruptions.
- **Chores**
	- Enhanced debug logging for sync sessions to provide clearer insights during synchronization processes.

### DIFF
--- a/packages/database/src/models/SyncSession.ts
+++ b/packages/database/src/models/SyncSession.ts
@@ -2,6 +2,7 @@ import { DataTypes } from 'sequelize';
 import { SYNC_DIRECTIONS } from '@tamanu/constants';
 import { Model } from './Model';
 import { type InitOptions } from '../types/model';
+import { log } from '@tamanu/shared/services/logging';
 
 export class SyncSession extends Model {
   declare id: string;
@@ -41,7 +42,11 @@ export class SyncSession extends Model {
   }
 
   static async addDebugInfo(id: string, info: object) {
-    const session = await this.findOne({ where: { id } });
+    // the SKIP LOCKED means we don't lock the sync_sessions table's row
+    // if it's already contended. but that also means that we might lose
+    // some debug info in some cases. so let's debug log it as well.
+    log.debug('Sync session debug', { id, ...info });
+    const session = await this.findOne({ where: { id }, skipLocked: true });
     await session?.update({
       debugInfo: { ...session.debugInfo, ...info },
     });
--- a/packages/database/src/models/SyncSession.ts
+++ b/packages/database/src/models/SyncSession.ts
@@ -41,7 +41,29 @@ export class SyncSession extends Model {
     );
   }
 
-  static async addDebugInfo(id: string, info: object) {
+  /**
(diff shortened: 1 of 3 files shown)


packages/database/src/models/SyncSession.ts
```diff
@@ -41,7 +42,11 @@ export class SyncSession extends Model {
   }
 
   static async addDebugInfo(id: string, info: object) {
-    const session = await this.findOne({ where: { id } });
+    // the SKIP LOCKED means we don't lock the sync_sessions table's row
+    // if it's already contended. but that also means that we might lose
+    // some debug info in some cases. so let's debug log it as well.
+    log.debug('Sync session debug', { id, ...info });
```
edmofro, line 48: Unfortunately `debugInfo` is now a misnomer, it will soon be renamed `info` and is used for actual business rules 

We could split out `addDebugInfo` and `addInfo` utils, which can write to the same column but with debug level allowing skips?

edmofro (commented):
Great catch

packages/database/src/models/SyncSession.ts
passcod: Done, but I don't know which are critical and which aren't, so I've changed one but left the others as debug for now.

edmofro (approved):
Nice, thanks! @dannash100 can take care of moving any important business logic parameters to `setParameters`, and either breaking out a `parameters` column from `debugInfo` or renaming `debugInfo` to `info`. This is great though, cheers.

PR #7692, opened 2025-05-21
fix(suggester): Hotfix 2.30: Update translated suggester logic

tcodling:
### Changes

During my devoncall shift I did a hotfix that changed the way translations work with suggesters in order to fix a bug and improve performance. Unfortunately in the rush I missed both the consideration of visibilityStatus and I was completely overriding the `whereBuilder` for anything translated meaning all custom `whereBuilder` logic broke 🤦 

In this new fix I have taken a step back and updated the suggester logic with 2 main things.
- A subquery that adds a `translation` attribute to returned records with the relevant translated text is then assigned to the name property if it exists through the mapper to display on the front end
- A subquery within the `whereBuilder` that checks the `searchQuery` against the translated text if it exists and if not searches against the original `searchColumn` for the record

Note: I am a bit concerned about the performance of these subqueries so please flag if this looks alarming and I need to take another look!

It might seem like a lot of lines changed for a hotfix so let me know if I should trim it down or do the refactors separately. But I feel the changes I made (parameterising `whereBuilder` and its usages) are low risk and worth it

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/__tests__/apiv1/Suggestions.test.js
+++ b/packages/facility-server/__tests__/apiv1/Suggestions.test.js
@@ -714,6 +714,8 @@ describe('Suggestions', () => {
       expect(result3.body.length).toEqual(1);
       expect(result3.body[0].name).toEqual('banana');
     });
+
+    it.todo('translated ordering');
   });
 
   it('should respect visibility status', async () => {
--- a/packages/facility-server/__tests__/apiv1/Suggestions.test.js
+++ b/packages/facility-server/__tests__/apiv1/Suggestions.test.js
@@ -15,9 +15,10 @@ import {
   randomSensitiveLabRequest,
 } from '@tamanu/database/demoData';
 import { findOneOrCreate } from '@tamanu/shared/test-helpers';
-import { fake } from '@tamanu/fake-data/fake';
+import { fake, chance } from '@tamanu/fake-data/fake';
 import { createTestContext } from '../utilities';
 import { testDiagnoses } from '../seed';
+import { sortBy } from 'lodash';
 
 describe('Suggestions', () => {
   let userApp = null;
@@ -606,8 +607,9 @@ describe('Suggestions', () => {
 
   describe('Translations', () => {
     beforeEach(async () => {
-      await models.ReferenceData.truncate({ cascade: true, force: true });
-      await models.TranslatedString.truncate({ cascade: true, force: true });
+      const { TranslatedString, ReferenceData } = models;
+      await ReferenceData.truncate({ cascade: true, force: true });
+      await TranslatedString.truncate({ cascade: true, force: true });
     });
 
     it('should return translated labels for current language if present in the db', async () => {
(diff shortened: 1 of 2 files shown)


packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -42,26 +31,34 @@ const ENDPOINT_TO_DATA_TYPE = {
...
-        AND "string_id" = '${getTranslationPrefix(dataType)}' || "${modelName}"."id"
-        LIMIT 1
+          SELECT "text" 
+          FROM "translated_strings" 
+          WHERE "language" = :language
+          AND "string_id" = :translationPrefix || "${modelName}"."id"
+          LIMIT 1
       )`),
-        'translated_name',
+        'translation',
```
tcodling, line 46: Wont always be the name key

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -42,26 +31,35 @@ const ENDPOINT_TO_DATA_TYPE = {
...
+// Search against the translation if it exists, otherwise search against the searchColumn
+const getTranslationWhereLiteral = (modelName, searchColumn) =>
+  Sequelize.literal(`COALESCE(
+      (SELECT "text" 
+        FROM "translated_strings" 
+        WHERE "language" = :language
+        AND "string_id" = :translationPrefix || "${modelName}"."id"
+        LIMIT 1),
+      "${modelName}"."${searchColumn}"
+    ) ILIKE :searchQuery`);
```
tcodling, line 61: Don't know enough about SQL/DBs to tell if this is really bad performance wise. Definitley flag if I need to change approach here 🙏

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -42,26 +29,37 @@ const ENDPOINT_TO_DATA_TYPE = {
...
-        SELECT "text" 
-        FROM "translated_strings" 
-        WHERE "language" = :language
-        AND "string_id" = '${getTranslationPrefix(dataType)}' || "${modelName}"."id"
-        LIMIT 1
+          SELECT "text" 
+          FROM "translated_strings" 
+          WHERE "language" = :language
+          AND "string_id" = :translationPrefix || "${modelName}"."id"
+          LIMIT 1
```
tcodling, line 42: Same as below I am a little concerned about bad performance with this

packages/facility-server/__tests__/apiv1/Suggestions.test.js
tcodling, line 1: Made a translation describe block for the 3 translation tests
- Edited the suggestion endpoint test to also test for falling back to original label
- Made a test for the single record suggester endpoint making sure it works as expected with and without translations

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -291,10 +278,9 @@ const referenceDataBodyBuilder = ({ type, name }) => {
 createSuggester(
   'multiReferenceData',
   'ReferenceData',
-  (search, { types }) => ({
+  ({ search, query: { types } }) => ({
+    ...DEFAULT_WHERE_BUILDER({ search, modelName: 'ReferenceData' }),
```
tcodling, line 282: I tweaked the declaration of `whereBuilders` to make use of DEFAULT_WHERE_BUILDER where possible and also supply it with the new `modelName` value needed to created the translation where literal

packages/facility-server/app/routes/apiv1/suggestions.js
passcod: There's an index on those exact fields so it should be fine.

passcod: Are we able to use binds instead of replacements?

passcod (approved):
LGTM in general. I'm a bit concerned with the size of the fix, but given we're too late to release that as a hotfix this week, it would be good to get some tester attention on it if possible next week.

tcodling:
@edmofro If you wanted to have a pass at this. Going to make sure this gets a really detailed test from me and a tester today 🙏

packages/facility-server/app/routes/apiv1/suggestions.js
tcodling: I think it should be an easy swap out but will make the change even bigger 😬

edmofro (commented):
The code all looks really solid, have just made a couple of suggestions for readability

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -185,23 +163,21 @@ function createAllRecordsRoute(
...
       const results = await model.findAll({
         where,
-        attributes: getTranslationAttributes(endpoint, modelName),
         order: [[Sequelize.literal(searchColumn), 'ASC']],
-        replacements: {
+        attributes: getTranslationAttributes(modelName),
+        bind: {
           language,
+          translationPrefix: `${REFERENCE_DATA_TRANSLATION_PREFIX}.${getDataType(endpoint)}.`,
+          searchQuery: '%',
```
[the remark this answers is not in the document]
tcodling: Ah yep the search in there whereBuilder is no longer valid

edmofro (commented):
Sorry I know you have already tested the heck out of this but just have some more comments - we can talk about how necessary they are tomorrow if you're really confident in the current implementation and it would add risk

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -386,11 +389,16 @@ REFERENCE_TYPE_VALUES.forEach((typeName) => {
 });
 
 createSuggester('labTestType', 'LabTestType', () => VISIBILITY_CRITERIA, {
-  mapper: ({ name, code, id, labTestCategoryId }) => ({ name, code, id, labTestCategoryId }),
+  mapper: ({ dataValues: { translation } = {}, name, code, id, labTestCategoryId }) => ({
+    name: translation || name,
```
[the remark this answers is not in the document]
tcodling: That first one looks PR looks good actually Ill try implement that for sure I think it will just need to supply `searchColumn` to `name` instead. Not sure if you have tested the second one but the risk on that one worries me a lot more 🤔

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -204,8 +195,20 @@ function createSuggesterCreateRoute(
...
+        FROM "translated_strings" 
+        WHERE "language" = $language
+        AND "string_id" = '${getTranslationPrefix(endpoint)}' || "${modelName}"."id"
+        LIMIT 1),
+      "${modelName}"."${searchColumn}"
+    ) ILIKE $searchQuery`);
+};
+
+const DEFAULT_WHERE_BUILDER = ({ endpoint, modelName, searchColumn = 'name' }) => ({
+  [Op.or]: [getTranslationWhereLiteral(endpoint, modelName, searchColumn)],
```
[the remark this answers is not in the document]
tcodling: Oo thats kind of a leftover bit i never figured out a way around but forgot to add a todo 🤔 Basically the sequilize literal seemingly needed to be supplied to an array to work. Im sure there is a proper way to insert into the where object but I obviously never got around to figuring it out. Ill look into this

tcodling:
@edmofro Quick update: this is so close and working pretty much as you expected :) Just struggling with an ambiguous column error for endpoints that have a custom `includeBuilder` due to the "name" key. 

This "working" version uses a hack that doesn't apply the special order logic for these cases which is obviously not ok. Just thought id flag as almost done incase you wanted to look or had any ideas.

packages/facility-server/app/routes/apiv1/suggestions.js
```diff
@@ -42,25 +29,32 @@ const ENDPOINT_TO_DATA_TYPE = {
...
+  WHERE "language" = $language
+  AND "string_id" = '${getTranslationPrefix(endpoint)}' || "${modelName}"."id"
+  LIMIT 1
+)`;
+
+const getTranslationAttributes = (endpoint, modelName, searchColumn = 'name') => ({
+  include: [
+    [
+      Sequelize.literal(
+        `COALESCE(${getTranslationSubquery(endpoint, modelName)}, "${modelName}"."${searchColumn}")`,
```
tcodling, line 50: This gets repeated quite a few times so might pull it out

PR #7695, opened 2025-05-22
refactor(audit): Remove two columns from changelogs

dannash100:
### Changes
Requested in program registry demo with Data team
Removes record_update, record_sync_tick from changelogs
Removing updated_at deleted_at messes with ability to query through model, id rather just leave tbh

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


edmofro (commented):
I think the metadata might also be able to be dropped, using `timestamps: false` (see e.g. `DebugLog`). That would also turn off the automatic population of `created_at`, but you could set that up manually through the migration to default at the db level rather than relying on Sequelize.

Rest of the changes lgtm though

dannash100:
> population
Ohk! ill give it a go

edmofro (commented):
Does `created_at` have a default of now in the db?

dannash100:
> Does `created_at` have a default of now in the db?

I thought so ill double check tho

dannash100:
@edmofro yeah it does
Screenshot 2025-05-26 at 2 53 35 PM

edmofro (approved):

PR #7699, opened 2025-05-22
fix(programRegistry): SAV-974: Fix desktop program registration caching bug

tcaiger:
### Changes

This change fixes an error where a stale program registration id from the react-query cache is being used as the basis for a subsequent data fetch to get the registration conditions. The stale id no longer exists which results in a 404 error. When invalidating a query in react-query, it will return stale data for a render or 2 until the new data is fetched in the background. In most cases this is fine, but in the case of a deletion this is causing the above problem. I fixed the issue by resetting the relevant query rather than invalidating it which completely clears out the stale data from the cache.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/features/ProgramRegistry/DeleteProgramRegistryFormModal.jsx
+++ b/packages/web/app/features/ProgramRegistry/DeleteProgramRegistryFormModal.jsx
@@ -31,18 +31,18 @@ const Body = styled.div`
 const useDeletePatientProgramRegistration = (patientProgramRegistration = {}, onSuccess) => {
   const api = useApi();
   const queryClient = useQueryClient();
-  const { id, patientId } = patientProgramRegistration;
+  const { id: registrationId, patientId } = patientProgramRegistration;
 
   return useMutation({
     mutationFn: async () => {
-      return api.delete(`patient/programRegistration/${id}`);
+      return api.delete(`patient/programRegistration/${registrationId}`);
     },
     onSuccess: async () => {
-      onSuccess();
-      await queryClient.invalidateQueries(['patient', patientId]);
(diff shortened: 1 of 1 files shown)


packages/web/app/features/ProgramRegistry/DeleteProgramRegistryFormModal.jsx
```diff
@@ -31,18 +31,18 @@ const Body = styled.div`
 const useDeletePatientProgramRegistration = (patientProgramRegistration = {}, onSuccess) => {
   const api = useApi();
   const queryClient = useQueryClient();
-  const { id, patientId } = patientProgramRegistration;
+  const { id: registrationId, patientId } = patientProgramRegistration;
```
tcaiger, line 34: Just added a slight rename of the id to make it clearer too.

packages/web/app/features/ProgramRegistry/DeleteProgramRegistryFormModal.jsx
```diff
@@ -31,18 +31,18 @@ const Body = styled.div`
...
 
   return useMutation({
     mutationFn: async () => {
-      return api.delete(`patient/programRegistration/${id}`);
+      return api.delete(`patient/programRegistration/${registrationId}`);
     },
     onSuccess: async () => {
-      onSuccess();
-      await queryClient.invalidateQueries(['patient', patientId]);
+      await queryClient.resetQueries(['patient', patientId]);
```
tcaiger, line 41: This is the key line

dannash100 (approved):
Great find with this yuss

PR #7700, opened 2025-05-22
fix(desktop): Properly display custom grab cursor

NavarroEmilioLuis:
### Changes

Yeah very weird issue, the custom cursor grab SVG we were using stopped appearing. Given it only happens on production builds I get the feeling we probably changed the way we did auto deploys in some way, or a ghost dependency is resolving differently so when it packs the production build, something funky goes on with the CSS and it just shows an empty class, with no rule whatsoever declared.

We could also solve this by embedding the SVG directly in the CSS as a data URL, but this seemed unideal, instead I went with this, [see Vite docs](https://vite.dev/guide/assets#explicit-url-imports) for more info on that. Anyway this approach work, and every other similar SVG/CSS/URL combo we have doesn't really have a problem so it's something way too specific to keep looking IMO.


tcodling (approved):
This looks like it would have been so tricky to figure out! Lgtm! Great explanation + link

PR #7468, opened 2025-03-26
feat(audit): NASS-1609: Patient router auditing

tcodling:
### Changes

This pr adds auditing of the single record views endpoints in patient.js. This is requires some small code changes to some endpoints but nothing major. Most annoying thing is I had to send facility id from the front end for any endpoint being audited. I also standardised a bunch of places to use the same PAD query

### Deploys

- [x] **Deploy to Tamanu Internal** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

PR #7701, opened 2025-05-22
fix(certificates): Patient birth certificate not loading

NavarroEmilioLuis:
### Changes

We recently did a [change in this PR](https://github.com/beyondessential/tamanu/pull/7468/files#diff-7e2162ae0a18d7cb4043d0508018b8aac3b82b7e4ee80feed6f66019b8d58e62L22-R25) that didn't quite match what we were doing, as we previously relied on controlling `useQuery` `enabled` option, without it, the `isLoading` property is always `true` and leaves the certificate loading forever if the patient doesn't have mother and father specified.

I checked the PR and this seems to be the only place where we would change the behavior so yeah this is it. I decided to update the hook to be more generic so we're able to override with any `fetchOptions`. Also went with this name because it seems the most repeated one on the web package (while `queryOptions` and `useQueryOptions` coming second).

### DIFF
--- a/packages/web/app/api/queries/usePatientAdditionalDataQuery.js
+++ b/packages/web/app/api/queries/usePatientAdditionalDataQuery.js
@@ -2,14 +2,15 @@ import { useQuery } from '@tanstack/react-query';
 import { useApi } from '../useApi';
 import { useAuth } from '../../contexts/Auth';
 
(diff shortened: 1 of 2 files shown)


dannash100 (approved):
Great catch thank u

dannash100 (approved):
Great catch thank u

PR #7703, opened 2025-05-23
feat(programRegistry): SAV-970: Migrate patient program registration registration_date

tcaiger:
### Changes

The registration date for patient program registration records is stored in the historical records rather than the most_recent one. This change is to add a migration step to set the patient_program_registrations.date field from the oldest record in the history.

Note I also had to move the migration which sets deleted_at for records with status recordedInError to be before this new change.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
This LGTM, but I'm thinking if we should update the date before entering the logs? do we ever read that field when reading history?

New changes will always have the same `date` value so not sure how that will work?

tcaiger:
> This LGTM, but I'm thinking if we should update the date before entering the logs? do we ever read that field when reading history?
> 
> New changes will always have the same `date` value so not sure how that will work?

If we update the date before the logs does that mean that the date of the last history update will be wrong? Is that what you mean by "New changes will always have the same `date` value so not sure how that will work?"

NavarroEmilioLuis:
> If we update the date before the logs does that mean that the date of the last history update will be wrong? Is that what you mean by "New changes will always have the same `date` value so not sure how that will work?"

Not really, so whenever we start using program registries after this migration, new change logs will be generated, however the `date` column will remain the same through all of those, unlike the ones we are creating in the migration, which will have different `date` values for the same registration chain.

I can think of two possible questions:
- Do we change the `date` of all registrations to match the first one, not just the recent one?
- When we read the history on the frontend, the "date recorded" column will always be the same on new changes and different on historical ones as it is. Which I don't think it's expected at all cause then what's the point of that column? It's probably meant to display the date of the registration history 🤔

PR #7730, opened 2025-05-29
fix(mobile): Fix build issue

NavarroEmilioLuis:
### Changes

So a series of unfortunate events broke mobile with the most unhelpful error message.

So in the inactive state PR, [this file here](https://github.com/beyondessential/tamanu/pull/7666/files#diff-442a60ef69628b22cb04709045cb76c92df05c330bd47b6d8970c607fdeefe20L63-R134) had changes that imported a migration that didn't exist as it was renamed at some point. The linter didn't complain and no one noticed, and it seems mobile is happy to run a file that doesn't exist without any error whatsoever, great stuff.

### Deploys


NavarroEmilioLuis:
Beat by #7728, will include the leftover commit in #7731

PR #7739, opened 2025-05-29
fix(sync): Don't lock sync_sessions to write debug info 2.32 hotfix

NavarroEmilioLuis:
### Changes

Given we might use this version, this is an important port missing from v2.16 (https://github.com/beyondessential/tamanu/pull/7702)


dannash100 (approved):

chris-bes (approved):
Approved since this is already hotfixed

PR #7740, opened 2025-05-29
fix(programRegistry): SAV-865: Update value as it should be all lowercase

NavarroEmilioLuis:
### Changes

Well, I missed this when updating the migrations in #7630, my bad!


passcod (approved):

tcaiger (approved):
Good catch

PR #7772, opened 2025-06-04
fix(sync): SAV-983: Ensure records updated by a sync have a unique sync tick on the central-server

rohan-bes:
### Changes

Bit of a sneaky issue here. It was possible that edits made to the central database while a sync was persisting would share a `updated_at_sync_tick` with the records from that sync. The problem is that when we build the `sync_lookup` table, we rely on the `synx_device_ticks` table to tell us which records were pushed by which device. If records are incorrectly treated as part of a sync, they won't sync down to the facility.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sync behavior to ensure records modified during sync have distinct sync ticks, preventing issues with record visibility during pull operations.

- **Tests**
  - Added new test cases to verify correct handling of sync ticks during device synchronization and concurrent edits.
  - Enhanced test setup to ensure a clean state before each test run.

### DIFF
--- a/packages/central-server/app/sync/CentralSyncManager.js
+++ b/packages/central-server/app/sync/CentralSyncManager.js
@@ -649,6 +649,11 @@ export class CentralSyncManager {
           { direction: SYNC_SESSION_DIRECTION.INCOMING },
         );
 
+        // Tick tock once more to ensure that no records that are subsequently modified will share the same sync tick as the incoming changes
+        // notably so that if records are modified by adjustDataPostSyncPush(), they will be picked up for pulling in the same session
+        // (specifically won't be removed by removeEchoedChanges())
+        await this.tickTockGlobalClock();
+
         return tock;
       });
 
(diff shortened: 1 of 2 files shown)


packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -647,16 +647,18 @@ export class CentralSyncManager {
...
         return tock;
       });
 
       await models.SyncDeviceTick.create({
         deviceId,
         persistedAtSyncTick,
       });
-      // tick tock global clock so that if records are modified by adjustDataPostSyncPush(),
-      // they will be picked up for pulling in the same session (specifically won't be removed by removeEchoedChanges())
-      await this.tickTockGlobalClock();
```
rohan-bes, line 659: Previously it was this tickTock that might've been blocked (perhaps by another sync) which meant there was a reasonable time window in which other changes on the central-server might occur and share the sync tick with the synced records.

Decided to cut it as now we're tickTocking in the transaction anyway so there's no need to do it here.

edmofro (approved):
Thank you, this is a great catch and solve

@chris-bes will need to apply this on Tupaia if we are using echo-removal

PR #7812, opened 2025-06-12
feat(programRegistry): SAV-976: Support configurable condition categories in desktop

NavarroEmilioLuis:
### Changes

We have moved condition categories from being a hardcoded enum to being reference data in their own table `program_registry_condition_category`. This updates the desktop web frontend to use it and cleanup the older references, plus, a couple of additions to the facility API.


packages/web/app/features/ProgramRegistry/FormTable.jsx
```diff
@@ -86,8 +86,6 @@ const StyledTableRow = styled(TableRow)`
         padding-bottom: 40px !important;
       }
     `};
-}
```
dannash100, line 89: Is this intentional 🤔

packages/web/app/features/ProgramRegistry/ConditionHistoryModal.jsx
```diff
@@ -57,10 +55,11 @@ export const ConditionHistoryModal = ({ open, onClose, condition }) => {
       id: 'category',
       title: <TranslatedText stringId="programRegistry.category" fallback="Category" />,
       width: '25%',
-      accessor: ({ conditionCategory }) => (
-        <TranslatedEnum
-          value={conditionCategory}
-          enumValues={PROGRAM_REGISTRY_CONDITION_CATEGORY_LABELS}
+      accessor: ({ conditionCategoryCode, conditionCategoryName }) => (
+        <TranslatedReferenceData
+          value={conditionCategoryCode}
```
dannash100, line 60: I think this has to be id not code 🙏

dannash100 (changes requested):
Small tweak but looks great!

packages/web/app/features/ProgramRegistry/FormTable.jsx
NavarroEmilioLuis: Yeah that's the "remove extra bracket" commit. I found it very odd that it was there, probably from a merge conflict or something idk

dannash100 (approved):
Love it!

PR #7823, opened 2025-06-13
feat(translations): NASS-1711: Add support for translated surveys/programs

tcodling:
### Changes

In the first implementation of translations we didnt include programs/surveys. With our current systems It was pretty easy to add in with these following changes
- Support the option format used in surveys for nested option translations
- Support translating multiple text columns for one record

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **New Features**
  - Added translation support for survey question labels, helper texts, and option labels in the survey interface.
  - Extended translation handling to include programs, surveys, program data elements, survey screen components, and nested options during data import.
  - Enhanced program and survey selection dropdowns with translated labels for better internationalization.
  - Enabled translation of vitals table headings and select option values for improved localization.
  - Improved translation rendering in vital editing modals and history logs for clearer multilingual support.
  - Added translation support for survey response details and answer results with translated options.

- **Bug Fixes**
  - Ensured translation entries are correctly created for programs, surveys, program data elements, survey screen components, and nested options during data import.

- **Style**
  - Improved code consistency in survey question processing and survey question import logic (no impact on user experience).

- **Tests**
  - Expanded and reorganized test coverage for translation of imported data and nested options.


packages/central-server/app/admin/importer/translationHandler.js
```diff
@@ -1,38 +1,69 @@
...
+  } catch (e) {
+    parsedOptions = options;
+  }
+
+  if (isArray(parsedOptions)) return parsedOptions;
+  if (isObject(parsedOptions)) return Object.values(parsedOptions);
+  if (isString(parsedOptions)) return parsedOptions.split(/\s*,\s*/).filter((x) => x);
+
+  throw new Error('Invalid options format for translations');
+}
```
tcodling, line 37: we dont have a standard options format in tamanu. They can come in a range of comma seperated array formats or a JSON object. This just safely captures all the formats in use currently

packages/central-server/app/admin/importer/translationHandler.js
```diff
@@ -1,38 +1,69 @@
...
+
+    // Handle records with multiple translatable text fields by adding another layer of nesting
+    if (isObject(recordText)) {
+      Object.entries(recordText).forEach(([key, text]) => {
+        const stringId = `${REFERENCE_DATA_TRANSLATION_PREFIX}.${dataType}.${key}.${values.id}`;
+        if (text) {
+          translationData.push([stringId, text, DEFAULT_LANGUAGE_CODE]);
+        }
+      });
+    }
```
tcodling, line 61: handling for the special case `surveyScreenComponent` mentioned above (plus any future multi column translated records)

packages/central-server/app/admin/importer/translationHandler.js
```diff
@@ -48,14 +79,19 @@ export function collectTranslationData(model, sheetName, values) {
 export async function bulkUpsertTranslationDefaults(models, translationData) {
   if (translationData.length === 0) return;
 
+  // throw out duplicate stringIds from invalid configuration to prevent confusing errors
+  const filteredTranslationData = translationData.filter(
+    (item, index, self) => self.findIndex((t) => t[0] === item[0]) === index,
+  );
```
tcodling, line 85: Again feels a little hacky but not sure where this should go other than here. This is really just to prevent translation errors from occuring and distracting from the core configuration error

NavarroEmilioLuis (commented):
The extensive tests are nice. Overall looks good, I just had more than a couple of comment/questions that I feel we should address first. Although on a second look, probably just the last one issue where I think we will never hit that "OR"

packages/web/app/components/Surveys/SurveyQuestion.jsx
```diff
@@ -86,22 +87,49 @@ const getCustomComponentForQuestion = (component, required, FieldComponent) => {
...
-  const configObject = getConfigObject(id, componentConfig);
-  const text = componentText || defaultText;
+
+  const text = (
+    <TranslatedReferenceData
+      category="surveyScreenComponent"
+      value={componentId}
+      fallback={componentText}
+    />
+  ) || <TranslatedReferenceData category="programDataElement" value={id} fallback={defaultText} />;
```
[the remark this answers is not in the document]
tcodling: Ohhhhhhh yes great catch!

packages/central-server/app/admin/importer/translationHandler.js
tcodling: Hmm no great call. I want to have another look at this one 🙏

tcodling:
@NavarroEmilioLuis Ready for re review. Only thing Is I am a little unsure about how to properly show errors for this. The translated strings have never really quite fit into the importer stats structure. I am wondering if the ideal solution would be to have a wrapper around `bulkUpsertTranslationDefaults` that populates the stats somehow in a way that makes sense? I dont know I feel like it needs a bit more thinking. I think in this case I am ok with it silently failing as this case only happens currently when misconfigured which has its own custom error message.

Agree though I would like better error handling for translations on import. I might make a card even

tcodling:
> @NavarroEmilioLuis Ready for re review. Only thing Is I am a little unsure about how to properly show errors for this. The translated strings have never really quite fit into the importer stats structure. I am wondering if the ideal solution would be to have a try catch around `bulkUpsertTranslationDefaults` that populates the stats somehow in a way that makes sense? I dont know I feel like it needs a bit more thinking. I think in this case I am ok with it silently failing as this case only happens currently when misconfigured which has its own custom error message.
> 
> Agree though I would like better error handling for translations on import. I might make a card even

actually i do need to fix that failed test quickly!

NavarroEmilioLuis (approved):
Amazing all LGTM and no I'm actually happy with the silent fail

packages/central-server/app/admin/importer/importRows.js
```diff
@@ -237,15 +235,17 @@ export async function importRows(
...
     } catch (err) {
       updateStat(stats, statkey(model, sheetName), 'errored');
       errors.push(new UpsertionError(sheetName, sheetRow, err));
     }
   }
 
-  await bulkUpsertTranslationDefaults(models, translationData);
+  if (errors.length === 0) {
+    await bulkUpsertTranslationDefaults(models, translationData);
+  }
```
tcodling, line 248: @NavarroEmilioLuis Realised this is way cleaner/proper than chucking out duplicate ids within the `bulkUpsertTranslationsDefaults` also probably saves us from some potential bugs/bad stringids down the road

packages/constants/src/importable.ts
```diff
@@ -75,8 +75,6 @@ export const OTHER_REFERENCE_TYPES = {
   PATIENT_FIELD_DEFININION_CATEGORY: 'patientFieldDefinitionCategory',
   SCHEDULED_VACCINE: 'scheduledVaccine',
   PROGRAM_REGISTRY: 'programRegistry',
-  PROGRAM_REGISTRY_CLINICAL_STATUS: 'programRegistryClinicalStatus',
-  PROGRAM_REGISTRY_CONDITION: 'programRegistryCondition',
```
tcodling, line 86: This was moved to a constant below and not removed from here. Checked usages and confident in deleting this. The effect will be is that it wont show up in the importable dropdown which is actually the intended behaviour as these are imported elsewhere

packages/web/app/components/VitalsTable.jsx
```diff
@@ -8,31 +8,77 @@ import { useVitalsQuery } from '../api/queries/useVitalsQuery';
...
+      if (isTranslatableOption) {
+        const optionStringId = getReferenceDataOptionStringId(
+          dataElementId,
+          'programDataElement',
+          value,
+        );
+
+        processedRecord[date] = {
+          ...record[date],
+          value: getTranslation(optionStringId, 'programDataElement'),
```
dannash100, line 75: I think the fallback here is wrong 🤔 should be value I think!

tcodling: Oh whoops i swear i fixed that ty for the catch

packages/web/app/components/EditVitalCellModal.jsx
```diff
@@ -2,12 +2,19 @@ import React, { useCallback } from 'react';
...
+    category="programDataElement" 
+    value={dataPoint?.component.dataElement.id} 
+    fallback={dataPoint?.component.dataElement.name} 
+  />
   const date = formatShortest(dataPoint?.recordedDate);
   const time = formatTime(dataPoint?.recordedDate);
-  const title = `${vitalLabel} | ${date} | ${time}`;
+  const title = <TranslatedText 
+    stringId="vitals.editCell.title" 
+    fallback=":vitalLabel | :date | :time" replacements={{ vitalLabel, date, time }} />
```
dannash100, line 17: Think funky formating here?

tcodling: I guess this is rare case where we are using text as a UI element almost. I feel like this should be string interpolated with getTranslation maybe?

PR #7826, opened 2025-06-14
fix(procedures): TAMOC-295: Separate suggesters - HOTFIX 2.30

chris-bes:
### Changes
- Separate suggesters so that the fetched current values don't interfere each other


anilvkaran (approved):

edmofro (commented):
Post-hoc review approved, thank you @chris-bes

PR #7844, opened 2025-06-18
fix(sync): TAMOC-296: Fix child records get soft deleted notes when patient merge HOTFIX 2.32

chris-bes:
Cherry-picked from 2.30

### DIFF
--- a/packages/shared/src/utils/index.js
+++ b/packages/shared/src/utils/index.js
@@ -17,3 +17,4 @@ export * from './getLabTestsFromLabRequests';
 export * from './numeralTranslation';
 export * from './enumRegistry';
 export * from './invoice';
+export * from './refreshChildRecordsForSync';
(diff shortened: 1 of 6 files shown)


Harnesh787 (approved):

edmofro (approved):

packages/shared/src/utils/index.js
```diff
@@ -17,4 +17,3 @@ export * from './getLabTestsFromLabRequests';
 export * from './numeralTranslation';
 export * from './enumRegistry';
 export * from './invoice';
-export * from './refreshChildRecordsForSync';
```
chris-bes: The test was failing because of this export. Somehow this was overwriting some of the models exported from `@tupaia/database` when using them in tests. Still don't fully understand it but removing this works and using existing import `import { refreshChildRecordsForSync } from '@tamanu/shared/utils/refreshChildRecordsForSync';` should still work

PR #7848, opened 2025-06-19
feat(programRegistry): SAV-975: Include extra data migration and fixes

NavarroEmilioLuis:
### Changes

Okay so this is all left over issues I found when testing, plus, updating the migration so that we can truly upgrade.

Updates:
- Add handling for unexpected null pointers
- Update migration to now insert the set of hardcoded categories, plus every category used so far
- Fix display issue with add program registry button
- Bring back the translated enum for backwards compatibility with old formatted `logs.changes`


dannash100 (approved):
Looks great thank you! 
Great to catch the enum registry translations

PR #7854, opened 2025-06-19
fix(upcomingVaccinations): TAMOC-298: hotfix 2.16: toggle upcoming vaccinations patient level

dannash100:
### Changes

Samoa is facing major facility server load due to historical data entry on vaccinations patient page.
Each load and each submittion of creation will re query the upcoming_vaccinations view at great expense due to the number of vaccines on the server.
This pr adds a new setting that will hide the info and is only shown if a button is pressed.

Screenshot 2025-06-20 at 2 08 14 PM

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/views/patients/panes/VaccinesPane.jsx
+++ b/packages/web/app/views/patients/panes/VaccinesPane.jsx
@@ -17,6 +17,7 @@ import {
 } from '../../../components/PatientPrinting';
 import { ImmunisationsTable, ImmunisationScheduleTable } from '../../../features';
 import { useAdministeredVaccines } from '../../../api/queries';
+import { useSettings } from '../../../contexts/Settings';
 
 const CovidCertificateButton = styled(Button)`
   margin-left: 0;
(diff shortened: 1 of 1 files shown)


packages/web/app/views/patients/panes/VaccinesPane.jsx
```diff
@@ -36,6 +37,10 @@ const TableWrapper = styled.div`
 `;
 
 export const VaccinesPane = React.memo(({ patient, readonly }) => {
+  const { getSetting } = useSettings();
+  const [showUpcomingVaccines, setShowUpcomingVaccines] = useState(
+    !getSetting('features.hideUpcomingVaccines'),
+  );
```
NavarroEmilioLuis, line 43: Don't love having both show and hide, I think it would be better to keep one. Not a strong opinion though

NavarroEmilioLuis (approved):
Hmmm don't love it can't be toggled but okay because it's a hotfix. I believe we should be making this new key a constant and using the constant instead, but also bc this is a hotfix I think it's weird to do that, so just to be mindful on port. Also I would push to not make this a setting at all and have this be standard behavior.

I'll leave it up to you

packages/web/app/views/patients/panes/VaccinesPane.jsx
dannash100: Oh yeah true that

PR #7856, opened 2025-06-20
fix(programRegistry): SAV-986: Bump configurable condition category records

NavarroEmilioLuis:
### Changes

So this ~migration~ step will bump these records to be able to sync them on new devices, otherwise there will be foreign key issues and format issues. PS: DBT issue wasn't caused here, update is on another PR with more work!


dannash100 (approved):
Very cool to use this system!

PR #7867, opened 2025-06-23
feat(db): add a trigger to set updated_at column for all changes, not just sequelize

passcod:
### Changes

Less a definite change and more of a proposal, this comes from discussions last week which highlighted that `updated_at` is set via Sequelize, and thus is missing from a bunch of updates made directly to the db or bypassing the sequelize model in some other way.

The way that timestamp updates is done in Diesel, the database framework used by the Meta server, is that there's a trigger that takes care of it. This means that any change, regardless of source, is covered. I've copied the trigger from Diesel here, and then adapted it for our circumstances because tests were failing due to some weird way we use the database.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

- **New Features**
  - The system now automatically updates the "updated_at" timestamp on records whenever they are modified, ensuring more accurate tracking of changes.

- **Chores**
  - Improved database migration handling to better support automatic timestamp updates.


edmofro (commented):
Very straightforward, and I like that it still lets the sequelize one win if provided
Are there any potential inconsistencies in the clock used by sequelize vs. db trigger?

passcod:
> Are there any potential inconsistencies in the clock used by sequelize vs. db trigger?

Technically yes in that sequelize (for some reason) resolves the timestamp locally (in JS) instead of passing through the `now()` sql function to the database on the insert, but as long as both the server and the database are on the same machine (which is 99% of our deployments), practically that will resolve to the same timestamp (although the JS timestamp is millisecond-precision, while the Postgres timestamp is microsecond or better).

passcod:
Only timesync would matter, both of these go from UTC time.

edmofro:
Right right, yeah cool - well then I think will always be perfectly appropriate, we don't rely on this column to be perfectly ordered

edmofro:
@passcod would be good to pick this up during cooldown, or close 🙏

tcodling:
bugbot run

dannash100 (approved):
Love it! Great addition

PR #7873, opened 2025-06-23
fix(audit): 2.34 fix: Backfill initial sync_lookup_tick

dannash100:
### Changes
Add an initial sync_lookup_tick where appropriate. 
Three cases:
- There exists an initial sync_lookup_tick starting at -1 or the sync_lookup table is empty: Skip migration
- There exists an initial sync_lookup_tick with source_start_tick > -1: Say its lookup_end_tick is 2132 - Grab nearest updated_at_sync_tick from lookup less than 2132. Creates initial lookup_sync_tick with (-1, 1225)
- There is no lookup_sync_tick: Create initial lookup_sync_tick with lookup_end_tick = max updated_at_sync_tick in lookup: say (-1, 52529)

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/database/src/migrations/1750719607520-backfillInitialSyncLookupTick.ts
```diff
@@ -0,0 +1,33 @@
...
+  // and a source_end_tick equal to the highest updated_at_sync_tick of any sync_lookup record with a
+  // updated_at_sync_tick less than the minimum source_start_tick.
+  // If there is already an initial sync_lookup_tick, we can skip the insert.
+  await query.sequelize.query(`
+    INSERT INTO sync_lookup_ticks (source_start_tick, lookup_end_tick)
+    SELECT -1,
+      (SELECT MAX(updated_at_sync_tick)
+        FROM sync_lookup
+        WHERE updated_at_sync_tick < (
+          SELECT MIN(source_start_tick)
```
edmofro, line 21:
```suggestion
          SELECT MIN(lookup_end_tick) 
```
I think?

edmofro (changes requested):
Will the `sync_lookup_ticks` table have entries at the time this is run?

packages/database/src/migrations/1750719607520-backfillInitialSyncLookupTick.ts
dannash100: If there are no sync_lookup ticks it will fail due to a null lookup_end_tick on insert.
So I maybe idea is to coalesce the lookup_end_tick.

dannash100: So currently three paths
- There exists an initial sync_lookup_tick starting at -1: DO NOTHING
- There exists an initial sync_lookup_tick > -1: Say its 2000 Grab nearest updated_at_sync_tick from lookup less than 2000. Creates with `(-1, 1225)`
- There is no sync_lookup_tick: Not sure here? Grab the max(updated_at_sync_tick) on sync_lookup?

dannash100: Ohk also will need to handle no lookup table entries too - see CI failing

edmofro (approved):
lgtm!

PR #7877, opened 2025-06-24
fix(programRegistry): SAV-975: Fix exporting condition categories

NavarroEmilioLuis:
### Changes

Okay so this had the issue that the name is too long. The last time we resolved with removing the program registry clinical status from the normal list so I'm taking that same route. And similar to that one, we are exporting them from the programs importer.

Note that program registry conditions was also updated to behave this way

### Deploys


packages/constants/src/importable.ts
```diff
@@ -80,6 +80,12 @@ export const OTHER_REFERENCE_TYPES = {
   PROGRAM_REGISTRY_CONDITION: 'programRegistryCondition',
 };
 
+const EXCLUDE_FROM_GENERAL_IMPORTABLE = [
```
dannash100, line 83: It seems that the only thing lost by removing them from `OTHER_REFERENCE_TYPES` Would be that they would be exlcuded from TRANSLATED_REFERENCE_TYPES.
So you could possibly remove them from OTHER and make own category

```
export const TRANSLATABLE_REFERENCE_TYPES = [
  ...REFERENCE_TYPE_VALUES,
  ...OTHER_REFERENCE_TYPE_VALUES,
  ...PROGRAM_REGISTRY_REFERENCE_TYPE_VALUES
];
```

dannash100 (approved):
Looks good I have an optional idea for mabye cleaner approach to consider :)

PR #7879, opened 2025-06-24
fix(programRegistry): SAV-977: Mobile update

NavarroEmilioLuis:
### Changes

Couple of changes from QA:
- Only show current condition categories
- Properly filter out in suggester

### Deploys


dannash100 (approved):
LGTM

PR #7885, opened 2025-06-25
fix(web): hotfix 2.34: Fix note modal height.

flynncs:
### Changes

Fixed note modal height on smaller screens using MUI media query.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/NoteModal/NoteModal.jsx
+++ b/packages/web/app/components/NoteModal/NoteModal.jsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState, useMemo } from 'react';
 import { useHistory, matchPath } from 'react-router-dom';
 import styled from 'styled-components';
 import MuiDialog from '@material-ui/core/Dialog';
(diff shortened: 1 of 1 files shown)


packages/web/app/components/NoteModal/NoteModal.jsx
```diff
@@ -73,14 +81,16 @@ const MemoizedNoteModalContents = React.memo(
...
     return (
       <FloatingMuiDialog
         open={open}
         onClose={onClose}
         baseWidth={500}
-        baseHeight={775}
+        baseHeight={isHeightBreakpoint ? 775 : 500}
         minConstraints={getMinConstraints(noteFormMode, note)}
-        maxConstraints={[500, 775]}
+        maxConstraints={getMaxConstraints(isHeightBreakpoint)}
```
tcodling, line 93: This feels a bit hacky imo. I can better understand why we do the min based on the content but I think the max should really by dynamic if possible?

Asked cursor and they suggested something like this? Worth a try maybe?

```
// Calculate maxConstraints based on screen size
    const [maxConstraints, setMaxConstraints] = React.useState([500, 775]);

    useEffect(() => {
      const updateConstraints = () => {
        const { innerWidth, innerHeight } = window;

        // Calculate modal size based on screen dimensions
        const maxWidth = Math.min(innerWidth * 0.9, 1200); // 90% of screen width, max 1200px
        const maxHeight = Math.min(innerHeight * 0.8, 800); // 80% of screen height, max 800px

        setMaxConstraints([maxWidth, maxHeight]);
      };

      updateConstraints();
      window.addEventListener('resize', updateConstraints);

      return () => window.removeEventListener('resize', updateConstraints);
    }, []);
```

flynncs: My concern with a fully dynamic width and height is that resizing the window will cause a lot of recalculations / re-renders of these values. 
I think that the current approach aligns with how CSS media queries / breakpoints work - but I agree it's not ideal, especially given the magic numbers. I'll do a quick refactor to put these as constants and memoise the min and max constraint functions.

tcodling (approved):
Awesome i like this way more. Thanks!

PR #7909, opened 2025-06-29
feat(portal): PHX-239: Patient portal project setup

flynncs:
### Changes

- New Patient Portal web package
- Deployment steps

### Deploys

- [x] **Deploy to Tamanu Internal**  %opsref=patient-portal-pulumi

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


.github/workflows/cd-package-patient-portal.yml
flynncs, line 1: would it be worth combining this with the `cd-package-frontend` yaml?

passcod: Given they're otherwise identical, yes.

packages/patient-portal/public/fonts/Moul-Regular.ttf
tcodling, line 1: I am guessing we couldn't move these fonts and icons to shared?

passcod (approved):
✅ for .github changes only

packages/patient-portal/public/fonts/Moul-Regular.ttf
flynncs: just had a further look into this - seems like the only way to share the files in the `public` dir between the two web apps is to have a script copy the files on build or on dev, as we can't import them via JS
happy to do that if you think it's worth it?

packages/patient-portal/src/fonts.css
```diff
@@ -0,0 +1,26 @@
...
+  unicode-range: U+1780-17FF; /* Khmer characters */
+}
+
+@font-face {
+  font-family: 'Moul';
+  font-style: normal;
+  font-display: swap;
+  src: url(shared/fonts/Moul-Regular.ttf) format('truetype');
+  unicode-range: U+1780-17FF; /* Khmer characters */
+}
```
tcodling, line 26: I think these have been removed from main so can probably be removed from here too

packages/patient-portal/src/views/LoginView.tsx
```diff
@@ -0,0 +1,19 @@
...
+          <Typography variant="h4" component="h1" gutterBottom>
+            Patient Login
+          </Typography>
+          <Typography variant="body1" gutterBottom>
+            Login Page Content
+          </Typography>
+        </Paper>
+      </Box>
+    </Container>
+  );
```
tcodling, line 18: Would be cool to maybe follow the styling patterns we use in the web package but totally fine to leave that to a later time as this is setup

tcodling (changes requested):
Looks amazing! Only thing is that you are referencing the old hardcoded khmer fonts that have been removed now. Its probably better to fix that up sooner rather than later. If you would rather fix inside the epic thats probably fine though just let me know

tcodling (approved):
Thanks!

PR #7911, opened 2025-06-29
feat(fhir): TAN-2524: Display village in address line of FHIR Patient if defined (hotfix 2.32)

rohan-bes:
### Changes

No other integrations care about the Patient address as far as I'm aware, so I think this change is totally safe.

This is just a basic fix to allow CanScreen to get the info they're after. A more comprehensive improvement will be made here: https://linear.app/bes/issue/TAN-2525/improve-fhir-address-for-fhir-patients

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/hl7fhir/materialised/Patient.test.js
+++ b/packages/central-server/__tests__/hl7fhir/materialised/Patient.test.js
@@ -23,16 +23,23 @@ describe(`Materialised FHIR - Patient`, () => {
 
   describe('full resource checks', () => {
     beforeEach(async () => {
-      const { FhirPatient, Patient, PatientAdditionalData } = ctx.store.models;
+      const { FhirPatient, Patient, PatientAdditionalData, ReferenceData } = ctx.store.models;
       await FhirPatient.truncate();
       await PatientAdditionalData.truncate();
       await Patient.truncate();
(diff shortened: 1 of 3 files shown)


packages/central-server/__tests__/hl7fhir/materialised/Patient.test.js
```diff
@@ -58,7 +65,7 @@ describe(`Materialised FHIR - Patient`, () => {
         address: [
           {
             city: additionalData.cityTown,
-            line: [additionalData.streetVillage],
+            line: [village.name],
```
edmofro, line 68: Do we want a test that it maintains the fallback? Maybe that's implicit in some other test in this file, I haven't checked

edmofro (commented):
Just one question

packages/central-server/__tests__/hl7fhir/materialised/Patient.test.js
rohan-bes: Yeah that functionality is tested in most of the other tests in this suite. Good point though, I was a bit lazy not creating an explicit test for this behaviour. But not too fussed as plan to change it rather soon.

edmofro (approved):

PR #7928, opened 2025-07-01
perf(mobileSync): KOR-35: Swap to optimised unsafe pragma settings during initial sync

dannash100:
### Changes
Alot of things in consideration here so this is likely to change inside epic

Toggle to unsafe but faster pragma settings for initial sync only
- As we just spun up the empty migrated database we don't care that initial sync isn't safe and just care about speed.
- Also set the defaults explicitly on load of app just incase
- I should double check that default pragma settings are identical on debug build and prod build ✅ 
- The actual usage in save changes is going to be changed within epic very shortly so don't worry about that really.

We might tweak these settings - but it seems we can be fully cowboy mode due to circumstances. I will investigate failure consequences. 
We also could tweak those defaults

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/mobile/App/infra/db/index.ts
```diff
@@ -117,6 +117,25 @@ class DatabaseHelper {
         console.error(error);
       }
     }
+    await this.setDefaultPragma();
```
dannash100, line 120: Reset default pragma on connection just incase

packages/mobile/App/infra/db/index.ts
```diff
@@ -117,6 +117,32 @@ class DatabaseHelper {
...
+    // Disables rollback journal - no transaction rollback or crash recovery
+    await this.client.query(`PRAGMA journal_mode = OFF;`); 
+    // Disables fsync() - SQLite doesn't wait for OS to confirm disk writes
+    await this.client.query(`PRAGMA synchronous = 0;`); 
+    // Sets page cache to 1M pages (~1GB with default 1KB pages)
+    await this.client.query(`PRAGMA cache_size = 1000000;`); 
+    // Locks database exclusively - prevents other processes from accessing
+    await this.client.query(`PRAGMA locking_mode = EXCLUSIVE;`); 
+    // Stores temporary tables, indices, and views in RAM instead of disk
+    await this.client.query(`PRAGMA temp_store = MEMORY;`);
```
NavarroEmilioLuis, line 144: This is the only bit I'd be a bit more concerned, cause when we do initial sync, some production devices do have out of memory issue (sometimes it's not only the device but the fact it has a very big dataset). So putting even more data there... idk. Maybe we could check system memory and dynamically activate it?

packages/mobile/App/services/sync/MobileSyncManager.ts
```diff
@@ -321,24 +322,50 @@ export class MobileSyncManager {
...
 
-      // update the last successful sync in the same save transaction,
-      // if updating the cursor fails, we want to roll back the rest of the saves
-      // so that we don't end up detecting them as needing a sync up
-      // to the central server when we attempt to resync from the same old cursor
-      await setSyncTick(this.models, LAST_SUCCESSFUL_PULL, pullUntil);
-    });
+    try {
+      // Save all incoming changes in 1 transaction so that the whole sync session save
+      // either fail 100% or succeed 100%, no partial save.
```
NavarroEmilioLuis, line 339: Whack that we weren't using that

NavarroEmilioLuis (approved):
I mean, overall looks good to me. I was a tad worried with the asynchronous one as well, but after reading a bit about it, it's perfectly fine for this initial sync work.

I did leave one comment in the only spot that I'm not too happy, but maybe the overhead is so slim that it wouldn't matter at all. (Like, huge initial syncs will fail no matter what, and the amount of extra memory used is negligible that it would cause us more issues with that out of memory problem)

packages/mobile/App/infra/db/index.ts
dannash100: Yes I am doing benchmarking on each change with diffrerent initial sync sizes in terms of timing per step and memory usage.
I will run a larger set of benchmarks after I add the batch snapshot table, streaming and the single transaction initial sync change to get a fairer test of the whole pull/save part of optimisation.

PR #7929, opened 2025-07-01
perf(mobileSync): KOR-32: Remove 999 restriction on SQLITE_MAX_VARIABLE_NUMBER

dannash100:
### Changes

This stuff is subject to change as we are moving a bunch of things around to align various optimisations and changes.

[SQLITE_LIMIT_VARIABLE_NUMBER](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber) was upped from 999 to 32766 in our min targeted android api's sqlite3 version
- This means we can remove most extra layers of batching that previously had to fit under this limit.
- In the case of patients we can insert many more than the previous 99 limit
- There is one place where we are hit with another limit and that is in the bulk insert of records in sync
  - Inserting batches over 997 rows will hit the compile time defined [SQLITE_LIMIT_EXPR_DEPTH](https://www.sqlite.org/c3ref/c_limit_attached.html#sqlitelimitexprdepth) limited at 1000.
  - This is fine tho as during benchmarks smaller batches of 500 or so records seems to be optimal for fast consistent bulk inserts to sqlite3 when compared against these batches `[20, 50, 80, 100, 200, 300. 400. 500. 600. 700. 800. 997]`

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


NavarroEmilioLuis (approved):
I think this makes sense but seeing some more recent messages about sqlite builds in the koror channel - IDK if this still holds true? Also, what API are you targetting? Are we completely sure production follows it?

Here is a link to an issue we had a while back, that basically we can install the app in a lower version than our recommended and that they do use older sqlite: https://beyondessential.slack.com/archives/GE9NHB95F/p1708641495125849?thread_ts=1708641404.890299&cid=GE9NHB95F

So depending on those two outcomes, I'd suggest we have a way of flipping a switch to provide one or the other dynamically or via settings. Though I see no point in blocking this work

dannash100:
@NavarroEmilioLuis luckily min supported apk is now 31 which is 3.32.0 sqlite and is first version to have that param limit increase to  SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0.

PR #7932, opened 2025-07-01
test(e2e): Fix broken patient sidebar tests

avaek:
### Changes

- Updates getCurrentBrowserDateISOFormat() to use browser timezone rather than UTC
- Updates locators for getOngoingConditionEditSubmitButton() which have changed recently

### Deploys

### Tests

- [x] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Date values now correctly reflect the browser's local timezone instead of UTC in relevant patient details.
  * Improved reliability in identifying the "Save" button during ongoing condition edits.


packages/e2e-tests/pages/patients/PatientDetailsPage/PatientDetailsPage.ts
```diff
@@ -336,10 +336,17 @@ export class PatientDetailsPage extends BasePatientPage {
...
+  * Gets current browser date in the YYYY-MM-DD format in the browser timezone
+  */
+async getCurrentBrowserDateISOFormat() {
+  const currentDate = new Date();
+  const year = currentDate.getFullYear();
+  const month = String(currentDate.getMonth() + 1).padStart(2, '0');
+  const day = String(currentDate.getDate()).padStart(2, '0');
+
+  return `${year}-${month}-${day}`;
+}
```
NavarroEmilioLuis, line 349: This looks like an adequate solution - however there is a util from tamanu shared called `getCurrentDateString` that will achieve just this FYI. Though given that this is e2e we might not want to actually use the same dependency as the one we use to populate normally, I'm really not familiar to how you're managing that 🙏

NavarroEmilioLuis (approved):
Left one small comment just in case it might be useful. Overall LGTM

packages/e2e-tests/pages/patients/PatientDetailsPage/PatientDetailsPage.ts
avaek: Thanks for the suggestion! 🙏 

Unfortunately we have an issue in the E2E package at the moment where utils like this are unable to be imported - they throw an error like `SyntaxError: Named export 'getCurrentDateString' not found. The requested module '@tamanu/utils/dateTime' is a CommonJS module, which may not support all module.exports as named exports.`
- I've spent a bit of time trying to fix this error but I wasn't able to get it working - in the interest of fixing this test as fast as possible I just went with the above solution

PR #7936, opened 2025-07-03
feat(mobileSync): KOR-6: Use snapshot table with batched records in mobile sync

dannash100:
### Changes
- There will be big changes to flow in follow up pr for KOR-34 following Data track single transaction initial sync 
- All syncs excluding initial sync will use a snapshot like table to save records in bulk batches during pull phase
  - This replaces the file system read/write RNFS usage as it is very slow writing files in base64
- Add tests for managing snapshot utilities

**Todo:**
- Progress display is messed up for save steps ✅  done
I made it a bit more responsive so you can tell if it frozen easier or anything. I personally love love love to see the progress be explicit with record batches as they persist

**To note:**
- When paired with streaming we can determine batch size based on the 2mb `CursorWindowSize` limit and really optimise pull performance.
- When combined with previous 2 PRs this already increases the overall initial sync time by 3.6 times. 
- Latest initial sync test was 127283 records in 2min

**Not handled in this pr:**
- Single transaction initial sync
- I will test larger flow with the addition of single transaction initial sync which is next up.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/mobile/App/services/sync/utils/manageSnapshotTable.ts
```diff
@@ -0,0 +1,85 @@
...
+  return JSON.parse(rows[0].data);
+};
+
+export const createSnapshotTable = async (queryRunner: QueryRunner, sessionId: string) => {
+  const tableName = getSnapshotTableName(sessionId);
+
+  // Just save the batches straight from the stream
+  // No outgoing changes snapshotting on push
+  try {
+  await queryRunner.query(`
```
dannash100, line 68: Possibly it shud just be one table? and truncated at start of sync? Not sure

packages/mobile/App/services/sync/utils/manageSnapshotTable.ts
```diff
@@ -0,0 +1,82 @@
...
+  return JSON.parse(rows[0].data);
+};
+
+export const createSnapshotTable = async (sessionId: string) => {
+  const tableName = getSnapshotTableName(sessionId);
+
+  // Just save the batches straight from the stream
+  // No outgoing changes snapshotting on push
+  try {
+    await Database.client.query(`
```
dannash100, line 56: Mabye not needed to have a table per session
We could aslo

Have one table created in a migration that is truncated after and before sync to be safe
- Just have a fixed table without session id like like sync_snapshot that is created and dropped

packages/mobile/App/services/sync/utils/saveIncomingChanges.ts
```diff
@@ -90,9 +88,31 @@ export const saveChangesForModel = async (
   }
 };
 
+const groupRecordsByType = async (
```
dannash100, line 91: Note to self that I am pretty certain we should batch this grouping from snapshot step and down to avoid loading every record into memory. 
looking into this now

dannash100: Done with new setting.

packages/mobile/App/services/sync/utils/manageSnapshotTable.ts
dannash100: I went with one table - that we remove - it kinda make sense to me as its just like a temp table to hold records.

dannash100: But im pretty sure truncate table method is faster - but I think the difference a few ms. Anywho

packages/mobile/App/services/sync/MobileSyncManager.ts
```diff
@@ -322,20 +334,19 @@ export class MobileSyncManager {
...
-    if (isInitialSync) {
+    const shouldUseUnsafeSchema = isInitialSync && mobileSyncSettings.useUnsafeSchemaForInitialSync;
+
+    if (shouldUseUnsafeSchema) {
       await Database.setUnsafePragma();
     }
-    
+
     try {
       await Database.client.transaction(async () => {
```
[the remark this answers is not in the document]
dannash100: Oh yeah you have to use the callbacks transactionalEntityManager! ill fix now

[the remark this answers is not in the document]
dannash100: In the upcoming PR KOR-34 I will try to work using the entityManager into the save a bit closer, might actually make things a bit cleaner.
It will be a bit of double work if I do anything more here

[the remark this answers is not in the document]
dannash100: I have improved this flow a bit in [feat(mobileSync): KOR-34: Initial sync single transaction](https://github.com/beyondessential/tamanu/pull/7990#top)

Still could be improved more I think so I will look into that

packages/mobile/App/services/sync/utils/saveIncomingChanges.ts
```diff
@@ -77,59 +75,77 @@ export const saveChangesForModel = async (
...
+        const chunkProgressCallback = (processedCount: number) => {
+          savedRecordsCount += processedCount;
+          progressCallback(
+            incomingChangesCount,
+            savedRecordsCount,
+            `Processed ${savedRecordsCount}/${incomingChangesCount} records...`,
+          );
+        };
+
+        await saveChangesForModel(model, sanitizedRecords, maxRecordsPerInsertBatch, chunkProgressCallback);
```
chris-bes, line 147: Do you reckon we should return the processed number to this caller and call `chunkProgressCallback()` here?

packages/mobile/App/services/sync/utils/getModelsForDirection.ts
```diff
@@ -4,8 +4,26 @@ import { MODELS_MAP } from '~/models/modelsMap';
...
+
+  // Create a model map with getTransactionalRepository method bound to the transactional entity manager
+  const transactionalModels = {};
+  for (const [modelName, modelClass] of Object.entries(modelsForDirection)) {
+    transactionalModels[modelName] = modelClass;
+    transactionalModels[modelName].getTransactionalRepository = () => entityManager.getRepository(modelClass);
+  }
+
+  return transactionalModels;
+};
```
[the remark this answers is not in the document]
dannash100: Yeah thats right! It needs to do everything through the entityManager that comes from transaction 🤯

packages/mobile/App/services/sync/utils/saveIncomingChanges.ts
dannash100: Yeah Im gonna rework that in next pr i rekon!

packages/mobile/App/services/sync/utils/manageSnapshotTable.ts
```diff
@@ -0,0 +1,50 @@
...
+import { chunk } from 'lodash';
+
+export const insertSnapshotRecords = async (
+  records: Record<string, any>[],
+  maxRecordsPerBatch: number,
+) => {
+  for (const batch of chunk(records, maxRecordsPerBatch)) {
+    await Database.client.query(`INSERT INTO sync_snapshot (data) VALUES (?)`, [
+      JSON.stringify(batch),
+    ]);
```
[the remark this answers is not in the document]
dannash100: Oh this is because the snapshot table inserts batches of 1000 records at at a time by default we talked a bit about weither to do single records or batches - and batches won out when chatting with Félix.

packages/mobile/App/services/sync/utils/saveIncomingChanges.ts
dannash100: As that intial single transaction sync changes the progress a bit like stage 2 becomes 33% - 100%
So will need some tweaks. I like that idea tho its much cleaner

dannash100: Actully looking at this agian, due to the various chunking that happens in the process - it makes for a smoother more accurate progress display having it passed down.

chris-bes (approved):
I think the changes are addressed in subsequent PR so LGTM!

PR #7937, opened 2025-07-03
refactor(api): bring api-client package up to parity with Facility's

passcod:
### Changes

The purpose is to replace the facility-server and mobile implementations with this package, while keeping it working for web (and in fact improve the web functionality).

Recommend reviewing one commit at a time, they're almost all split to one atomic change each.

This was not a copy-paste of functionality, but rather a feature-by-feature rethink and integration into the package, for consistency and quality, but also because api-client has to work both in the browser (via vite) and in Node.js. There's also some drive-by improvements that weren't in the Facility's implementation.

- f0a938a09d: when the api-client package was made, the info it returned from the login endpoint was everything the endpoint returned. however, things have evolved since, but the api-client didn't follow. this update instead returns everything the endpoint provides, extracting some specific things
- 0561dd7b3c, 21b2b86c78, 709cf0667f, 8c78681dc7, bd7e4cd088: misc improvements and bugfixes to the login and refreshToken routines
- beff3262cc: the `waitForAuth` functionality, on by default, delays api calls until an auth token is available. that means you can call an endpoint and the login in parallel and have the correct ordering.
- e78938a154: include the response object in the error instance when available
- 4b7dd8f148: return a specific error class for invalid credentials responses
- bb2e615220: export all the new (and some existing!) error types in the main package index
- 19e16e1865: if an endpoint returns no content and we're expecting json, return an empty object instead of an error
- 0dc0c25205: use the native fetch api `Headers` object to merge and create headers, instead of a plain object
- bf10ebafcc: allow a custom logger, default it to console, and use the logger in various places
- 90b0484ca4: automatically encode to JSON when the content-type matches and a body is provided that is not a buffer/uint8array
- 1beb4f13a0: in the constructor, allow specifying custom fetch options as defaults for the instance
- c7a134e52c: the `waitUntilOk` (in the facility code, called `waitUntilTrue`) method
- 5e832ce270: add default `accept` and `content-type` headers
- 7b8c93b6b5: the `timeout` option (as a shorthand to doing abortcontroller stuff)
- 73b8e5ef8d: the retry-with-backoff feature (off by default)

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

## Summary by CodeRabbit

* **New Features**
  * Enhanced authentication handling with improved token management and support for concurrent login serialization.
  * Added robust request retry logic with exponential backoff for improved reliability.
  * Introduced a polling utility to repeatedly fetch a resource until available or timeout.
  * Expanded error handling with more granular error types and richer error context.
  * Added request timeout support for fetch operations.

* **Bug Fixes**
  * Improved consistency and clarity in error extraction and handling for various HTTP response codes.

* **Chores**
  * Refined and extended error class hierarchy for clearer error categorization and handling.

### DIFF
--- a/packages/api-client/src/TamanuApi.js
+++ b/packages/api-client/src/TamanuApi.js
@@ -1,11 +1,13 @@
 import qs from 'qs';
 
 import { SERVER_TYPES } from '@tamanu/constants';
-import { NotFoundError, ForbiddenError } from '@tamanu/shared/errors';
 import { buildAbilityForUser } from '@tamanu/shared/permissions/buildAbility';
 
 import {
   AuthExpiredError,
+  AuthInvalidError,
+  ForbiddenError,
+  NotFoundError,
   ResourceConflictError,
   ServerResponseError,
   VersionIncompatibleError,
@@ -70,7 +72,10 @@ export class TamanuApi {
       );
       const serverType = response.headers.get('X-Tamanu-Server');
       if (![SERVER_TYPES.FACILITY, SERVER_TYPES.CENTRAL].includes(serverType)) {
-        throw new Error(`Tamanu server type '${serverType}' is not supported.`);
+        throw new ServerResponseError(
+          `Tamanu server type '${serverType}' is not supported.`,
+          response,
+        );
       }
 
       const {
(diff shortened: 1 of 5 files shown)


passcod:
I'm not sure how it's possible for the central-server tests to fail given central-server doesn't depend on api-client in any way

dannash100 (approved):
Ahh great improvements, I had a good look and nothing jumping out. Good stuff!

PR #7947, opened 2025-07-06
feat(desktopCharting): SAV-692: Support photo question type on desktop

NavarroEmilioLuis:
### Changes

We now support the ability to upload photo question types on desktop. Also, there is a difference when displaying them on the forms tab at patient level vs on the charts tab at encounter level. Lastly, a scheduled task to remove attachments from facility server after they have been synced.

### Deploys


packages/web/app/utils/survey.jsx
```diff
@@ -296,16 +298,24 @@ export function getFormInitialValues(
...
 }
 
-export const getAnswersFromData = (data, survey) =>
-  Object.entries(data).reduce((acc, [key, val]) => {
-    if (
-      survey.components.find(({ dataElement }) => dataElement.id === key)?.dataElement?.type !==
-      'PatientIssue'
-    ) {
+export const getAnswersFromData = async (data, survey) =>
+  Object.entries(data).reduce(async (accPromise, [key, val]) => {
```
edmofro, line 302: This is a little unusual to me, but it could be a personal preference. I would do something like `.map`, and then an `const entries = await Promise.all(`
```suggestion
  const answerEntryPromises = Object.entries(data).map(async ([key, val]) => {
  ...
  };
  const answerEntries = await Promise.all(entryPromises);
  const answers = Object.fromEntries(answerEntries);
```
Up to you though!

packages/database/src/models/SurveyResponse.ts
```diff
@@ -388,12 +388,25 @@ export class SurveyResponse extends Model {
...
         throw new Error(`no data element for question: ${dataElementId}`);
       }
-      const body = getStringValue(dataElement.type, value);
+      let body = getStringValue(dataElement.type, value);
       // Don't create null answers
       if (body === null) {
         continue;
       }
 
+      if (dataElement.type === PROGRAM_DATA_ELEMENT_TYPES.PHOTO) {
```
edmofro, line 397: I think this block suits better inside `getStringValue`. That way:
- We aren't running `getStringValue` with other code unnecessarily when it is a photo
- All data element types are handled in the same util function
- Any other checks like the `body === null` above get run on photo answers like all others, no risk of missing it from something common

packages/web/app/components/Field/FileChooserField.jsx
```diff
@@ -25,10 +37,68 @@ const ChangeSelectionButton = styled.a`
...
+  const ellipsisOffset = 3;
+  const lastHalfIndex = value.length - middlePoint + ellipsisOffset;
+  return value.slice(0, middlePoint) + '...' + value.slice(lastHalfIndex, value.length);
+};
+
+const ValueSection = ({ value, smallDisplay, showFileDialog, onClear }) => {
+  if (smallDisplay) {
+    const maxLength = 50;
+    const needEllipsis = value.name.length > maxLength;
+    const smallName = needEllipsis ? getSmallFileName(value.name, maxLength) : value.name;
```
edmofro, line 52: TIL "ellipsis" refers to the whole cluster of three dots, I always assumed that was the plural ellipses!

packages/web/app/components/Field/FileChooserField.jsx
```diff
@@ -42,9 +112,26 @@ export const FileChooserInput = ({ value = '', label, name, filters, onChange, .
     const file = event.target.files[0];
     if (!file) return;
 
+    const fileSize = file.size;
+    const maxFileSize = 10 * 1000 * 1000; // 10MB
```
edmofro, line 116: Should this be configurable per deployment, or tuneable once in production, or is there a reason it will be permanently the right number?

packages/facility-server/app/tasks/AttachmentCleanupTask.js
```diff
@@ -0,0 +1,91 @@
+import config from 'config';
+import { Op } from 'sequelize';
+
+import { ScheduledTask } from '@tamanu/shared/tasks';
+import { log } from '@tamanu/shared/services/logging';
+import { FACT_LAST_SUCCESSFUL_SYNC_PUSH } from '@tamanu/constants/facts';
+import { sleepAsync } from '@tamanu/utils/sleepAsync';
+
+export class AttachmentCleanupTask extends ScheduledTask {
```
edmofro, line 9: I wonder if we need this...we have the `PUSH_TO_CENTRAL_THEN_DELETE` sync direction, which works well for patient communications

packages/web/app/components/ViewPhotoLink.jsx
```diff
@@ -3,61 +3,134 @@ import styled from 'styled-components';
...
       setImageData(null);
       setErrorMessage(
         'You do not currently have an internet connection. Images require live internet for viewing.',
       );
-      setShowModal(true);
       return;
     }
 
     try {
+      await sleepAsync(2000);
```
edmofro, line 104: What's this for?

edmofro (changes requested):
Mostly very minor comments only, but one bigger question about the need for the cleanup task

packages/web/app/components/Field/FileChooserField.jsx
NavarroEmilioLuis: Was going for the second one, maybe we could make it configurable later via settings but that would also involve changing the text copy accordingly

packages/facility-server/app/tasks/AttachmentCleanupTask.js
NavarroEmilioLuis: Oh right yeah no I guess we don't need that then, I wasn't aware of that sync direction. I'll remove these and update the sync direction instead

packages/web/app/components/ViewPhotoLink.jsx
NavarroEmilioLuis: Testing purposes left behind

packages/web/app/utils/survey.jsx
NavarroEmilioLuis: Yeah no I just went for the path of least resistance on the rewrite, this pattern is only found on 1 other place in the codebase so I considered it okay. I'll change it up

packages/database/src/models/SurveyResponse.ts
edmofro: You've thought about it a lot harder than I have, thanks for all your reasoning and happy to go with whatever you choose!

packages/web/app/components/Field/FileChooserField.jsx
edmofro: I think text copy could be determined via whatever was in settings hey? Did the 10MB come from somewhere specific? I'm just conscious that in the past we've ended up with conflicts between the file size limits in different places (caddy/nginx vs. express vs. front end) so it's nice to be able to access it as a setting

packages/web/app/utils/survey.jsx
edmofro: Yeah great call!

packages/web/app/components/Field/FileChooserField.jsx
NavarroEmilioLuis: I think just historical reasons that was just the copy lying around and we haven't determined anything. I'll make it into a setting, which is always better anyway 🙏

edmofro (approved):
Looks great!

PR #7950, opened 2025-07-07
feat(patientMerge): SAV-988: Update patient merge logic

NavarroEmilioLuis:
### Changes

Changing to soft delete, because we don't really want to be hard deleting these - all changes end up propagating due to a sync issue and then appear soft deleted anyway, with the caveat that now all the sync ticks are messed up. With this change, we expect to see them soft deleted and propagate normally.

This is what used to happen:
- Merge and soft delete patient on central
- Hard delete e.g. PAD on central
- Patient soft deletion syncs to facility, hard delete of PAD doesn't
- Patient soft deletion side effect soft deletes PAD
- Soft deleted PAD syncs up to central and restores itself

### Deploys


edmofro (approved):
Awesome, tysm

edmofro:
@NavarroEmilioLuis note that it's not because of the sync lookup table that the records end up soft deleted. Could you please change the PR description to reflect this flow:
- Merge and soft delete patient on central
- Hard delete e.g. PAD on central
- Patient soft deletion syncs to facility, hard delete of PAD doesn't
- Patient soft deletion side effect soft deletes PAD
- Soft deleted PAD syncs up to central and restores itself

packages/central-server/app/admin/patientMerge/mergePatient.js
```diff
@@ -322,7 +313,9 @@ export async function reconcilePatientFacilities(models, keepPatientId, unwanted
   const existingPatientFacilityRecords = await models.PatientFacility.findAll({
     where,
   });
-  await models.PatientFacility.destroy({ where, force: true }); // hard delete
+  // This is the only place where hard deletion is allowed, as we will be creating
+  // new records to replace the all the old ones and then some.
+  await models.PatientFacility.destroy({ where, force: true });
```
edmofro, line 318: Just had a thought though, will this not end up re-created in soft-deleted form due to the same "bug" as the others?

NavarroEmilioLuis: Not really, cause we are hard deleting and then are creating both pairs again. We create a patient facility record for each patient for each facility any of them was in. This one is funky, but ultimately easier than checking which row exists and update it or create the missing one

NavarroEmilioLuis: Hmmm interesting. Somehow, it doesn't happen on main though (there are no soft deleted patient facilities with/without sync lookup - they remain hard deleted somehow), but I'm getting mixed results with the facility status as the patient facilities do appear there.

Regardless, I guess ideally we'd want to see soft deleted patient facilities for patient `b` originating from central, right?

edmofro: Yep that example you linked makes

But as discussed today, you've probably put in enough effort here - we can just mark it up with some comments about the hard bits and merge what we've got imo 🙏

PR #7953, opened 2025-07-07
refactor(sync): database cursor experiment

passcod:
### Changes

This is slower than polling the db because when we have auditing on (i think?) we need to add changelog records in and with the current architecture we end up needing to do that in JS, once per record. So it kills performance. If we could attach changelogs in sql then it would work smoooooth but I haven't figured that out.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


passcod:
Problem area: https://github.com/beyondessential/tamanu/pull/7953/files#diff-c0ba9c0ad38b4990c52ebeb5c11b035074e66710f37421c7d0e7358222d8209bR26-R29

PR #7962, opened 2025-07-08
tweak(pdf): NASS-1705: Global PDF font setting

tcodling:
### Changes

In testing for our MSF Syria deployment we noticed that Arabic characters were displaying as garbled english characters. We discovered this was because we were using the default Helvetica font for all pdfs which didn't support Arabic characters. To get around this we initially tried to change the default font for all PDFs but we had spacing issues meaning this change introduced big regressions. 

Due to time pressure and to prevent risks to other deployments we have decided to hide the "Global" font behind a feature flag so that we can turn it on for deployments that need it and tweak/improve specific PDFs rather than having to fix every single one in Tamanu to get this merged.

@dannash100 Did the initial work of registering a new font and I went through and hid it behind a setting + some cleanup.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/shared/src/utils/pdf/languageContext.jsx
```diff
@@ -1,87 +1,32 @@
...
-          if (boldFont.includes(currentFontFamily) && !forceToBodyFontWithGreaterFontWeight) {
-            newStyles.fontFamily = FONT.MOUL;
-          } else {
-            newStyles.fontFamily = FONT.BATTAMBANG;
-            if (forceToBodyFontWithGreaterFontWeight) {
-              newStyles.fontWeight = 500;
-            }
-          }
-        }
-        return newStyles;
```
tcodling, line 84: Old hardcoded fonts for Khmer being deleted in place of our new logic

packages/shared/src/assets/fonts/Vazirmatn-VariableFont_wght.ttf
dannash100, line 1: I think unused?

dannash100: Otherwise looks great :)

tcodling: oo thanks yea i just deleted one

rohan-bes (approved):
I think this looks good to me! Tbh a little out of my depth but nothing seems untoward :) Nice to see the custom khmer stuff cleaned up

PR #7970, opened 2025-07-09
fix(migrations): no-issue: Fix logic that detects if a trigger is installed during migrations (hotfix 2.35)

rohan-bes:
### Changes

The previous SQL for checking if a trigger was installed on a table was returning multiple rows due a faulty left join, which mean we tried to remove the trigger twice. This caused an error when we tried the second time: https://beyondessential.slack.com/archives/C0212PJ11DW/p1752033578080699

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


dannash100 (approved):
Great great catch

PR #7971, opened 2025-07-09
ci: try to fix central tests timeout by sharding more

passcod:
### Changes

I had an inkling that maybe central-server was failing and timing out in CI (but not locally!) because of the same kind of contention that led us to shard the tests in the first place. This change... doesn't solve the problem entirely, but does seem to make it less likely. So maybe it's some of this and some of something else, but this still helps a bit.

Additionally:
- also increased the sharding on facility-server tests to lower the overall CI time (now we're OSS and don't pay for minutes, it doesn't matter)
- add postgres 17 to the database tests
- use the node_modules caching better (saves ~1 minute per job)
- skip fetching and saving the node_modules cache if the package-lock.json hasn't changed (saves ~1-2 minutes per run)
- don't save and fetch a separate cache for mobile now that we have that package part of the main workspace

Timings:
- before
  - cold cache: [13m51](https://github.com/beyondessential/tamanu/actions/runs/16159282510?pr=7962)
  - hot cache: [9m20](https://github.com/beyondessential/tamanu/actions/runs/16158464326?pr=7938)
- after
  - cold cache: [11m46](https://github.com/beyondessential/tamanu/actions/runs/16165846389/attempts/1?pr=7971)
  - hot cache: [7m16](https://github.com/beyondessential/tamanu/actions/runs/16165846389/attempts/2?pr=7971)

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,7 @@ jobs:
           - { package: "@tamanu/database",                    postgres: 12 }
           - { package: "@tamanu/database",                    postgres: 14 }
           - { package: "@tamanu/database",                    postgres: 16 }
+          - { package: "@tamanu/database",                    postgres: 17 }
           - { package: "@tamanu/shared",                                   }
           - { package: "@tamanu/utils",                                    }
           - { package: "@tamanu/web-frontend" }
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (approved):
I'll take any form of time gains with this bit really, if we don't have any costs overhead then absolutely. Thanks for the update!

PR #7972, opened 2025-07-09
fix(desktopCharting): SAV-698: Remove flickering

NavarroEmilioLuis:
### Changes

Weird issue that ends up happening because of the way our contexts trigger re-renders. Wrapping them here makes their values stable without losing any data but avoiding the rendering issues


dannash100 (approved):
nice 1! hate those flashes

PR #7973, opened 2025-07-09
fix: test

NavarroEmilioLuis:
### Changes

Let's see what's up


NavarroEmilioLuis:
Okay so this test is not actually failing ([see this link](https://github.com/beyondessential/tamanu/actions/runs/16182092057/job/45680740259?pr=7973) and look for central server 2, logs should be there and you can see the test file itself passed) - the error log is expected and the test suite passes. This has to do with the general test failures.

PR #7903, opened 2025-06-26
fix(discharges): use information from facility and fallback to embedded info in discharge

passcod:
### Changes

_Add a brief description of the changes in this PR to help give the reviewer context._

### Deploys

- [x] **Deploy to Tamanu Internal** 

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

## Summary by CodeRabbit

* **New Features**
  * Discharge summary printouts now display facility details by retrieving the most up-to-date information from facility records when available.

PR #7974, opened 2025-07-09
fix(discharges): use information from facility and fallback to embedded info in discharge

passcod:
originally from #7903


dannash100 (approved):
# 100

PR #7975, opened 2025-07-09
Easier to visualize this way

NavarroEmilioLuis:


NavarroEmilioLuis:
Alright, closing because there is a strange issue happening with a unit test and we've already spent some time on this work. Plus, it seems to be okay without this change anyway

PR #7980, opened 2025-07-10
tweak(sync): NASS-1730: Update sync behaviour based on is_sensitive column

tcodling:
### Changes

In the epic we added a new `is_sensitive` column to facilities to support some new sync behaviour. This pr implements that behaviour. 

Included is the following:
- Update of `buildEncounterLinkedLookup` filter to add a `facilityId` to the lookup record if its from a `is_sensitive` facility
  - This should mean that these records only sync down to that facility and no others
- Swap out all models for encounter linked data to ensure they all use this sync filter
- Move any custom arguments/overrides to `buildEncounterLinkedLookup`
- Very large test suite checking that each data type does not sync away from a sensitive facility

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


tcodling:
bugbot run

tcodling:
bugbot run

packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
```diff
@@ -712,17 +712,11 @@ describe('Sync Lookup data', () => {
...
-      const expectedFacility = ['appointments', 'appointment_schedules'].includes(model.tableName)
-        ? facility.id
-        : null;
-
       expect(syncLookupRecord.dataValues).toEqual(
         expect.objectContaining({
           recordId: expect.anything(),
           recordType: model.tableName,
           patientId: expect.anything(),
-          facilityId: expectedFacility,
```
tcodling, line 725: Now facilityId is optional for all encounter linked models so this check is no longer valid

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,923 @@ describe('CentralSyncManager', () => {
       });
     });
   });
+
+  describe('facility sensitivity sync filtering', () => {
```
tcodling, line 2842: I worry a little about the performance of this describe block. It adds about a minute or more to the test suite on my local due to the amount of syncing/updating lookup table it does. 

I did look into trying to combine the test setup/syncing but i found negligible increases in performance and it became much harder to read. Am open to ideas here otherwise thinking I will create a card to work on improving performance if I have time.

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
```diff
@@ -1,12 +1,55 @@
...
+export function addSensitiveFacilityIdIfApplicable() {
+  return `
+    CASE
+      WHEN facilities.is_sensitive = TRUE THEN facilities.id
+      ELSE NULL
+    END
+  `;
+}
+
+export function buildEncounterLinkedLookupFilter(model: typeof Model, options?: Options) {
```
tcodling, line 32: Super open to suggestions on this one. I essentially had to move all the weird custom logic, arguments from the individual sync filters into here so its a bit messier than I would like ideally.

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
```diff
@@ -1,12 +1,55 @@
...
-    select: buildSyncLookupSelect(model, {
-      patientId: 'encounters.patient_id',
-    }),
-    joins: buildEncounterLinkedSyncFilterJoins([model.tableName, 'encounters']),
-  };
+export type JoinConfig = {
+  tableName: string;
+  columnName: string;
+  joinType?: 'LEFT' | 'INNER';
+};
```
tcodling, line 10: Follow sequilze syntax if possible (use type if possible)

tcodling:
bugbot run

tcodling:
Determinism failing but I think this should resolve once merged to the epic. Then I will start a proper dev test 🙏

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
```diff
@@ -1,12 +1,55 @@
...
+  extraSelects?: Record<string, string>; // extra selects to attach to the lookup select
+  extraJoins?: JoinConfig[]; // extra joins needed to traverse between this model and the encounters table
+};
+
+/**
+ * Helper function to determine if a facility_id should be populated in sync lookup
+ * Only populates facility_id when the encounter is from a sensitive facility
+ * This ensures sensitive encounters are only synced to their originating facility
+ */
+export function addSensitiveFacilityIdIfApplicable() {
```
edmofro, line 18: Could be a const rather than a function - all it does is return the same text every time. Funnily enough I think that will make the usage site less readable, but it is the normal thing to do.

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
```diff
@@ -1,12 +1,55 @@
...
+import {
+  buildEncounterLinkedSyncFilterJoins,
+  type JoinConfig,
+} from './buildEncounterLinkedSyncFilter';
 import { buildSyncLookupSelect } from './buildSyncLookupSelect';
 import type { Model } from '../models/Model';
 
+export type Options = {
+  extraSelects?: Record<string, string>; // extra selects to attach to the lookup select
+  extraJoins?: JoinConfig[]; // extra joins needed to traverse between this model and the encounters table
```
edmofro, line 10:
```suggestion
  joinsToEncounters?: JoinConfig[]; // extra joins needed to traverse between this model and the encounters table
```
Matching naming in params below

edmofro: Oh yo is this used??

packages/database/src/sync/buildEncounterLinkedSyncFilter.ts
```diff
@@ -1,18 +1,36 @@
...
-    `,
-    )
+    .map((table, i) => {
+      const currentTable = isString(tablesToTraverse[i])
+        ? tablesToTraverse[i]
+        : tablesToTraverse[i]?.model.tableName;
+
+      const joinTable = isString(table) ? table : table.model.tableName;
+      const joinColumn = isString(table) ? `${Utils.singularize(table)}_id` : table.joinColumn;
+      const joinType = isObject(table) && table.required ? 'INNER' : 'LEFT';
```
edmofro, line 23: Is `required` the same key sequelize uses?

packages/database/src/sync/buildEncounterLinkedSyncFilter.ts
```diff
@@ -1,18 +1,36 @@
...
+      return `
+        ${joinType} JOIN ${joinTable} ON ${currentTable}.${joinColumn} = ${joinTable}.id
+      `;
+    })
     .join('\n');
 }
 
 export function buildEncounterLinkedSyncFilter(
-  tablesToTraverse: string[], // e.g. [ 'survey_response_answers', 'survey_responses', 'encounters'] to traverse up from survey_response_answers
+  tablesToTraverse: JoinConfig[], // e.g. [ 'survey_response_answers', 'survey_responses', 'encounters'] to traverse up from survey_response_answers
```
edmofro, line 33: Could add an object to the e.g. in the comment

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,965 @@ describe('CentralSyncManager', () => {
...
+        maxRecordsPerSnapshotChunk: DEFAULT_MAX_RECORDS_PER_SNAPSHOT_CHUNKS,
+      },
+    };
+
+    beforeEach(async () => {
+      patient = await models.Patient.create(fake(models.Patient));
+      practitioner = await models.User.create(fake(models.User));
+      // Reset modules to ensure fresh imports
+      jest.resetModules();
+      // The config override actually doesnt apply to snapshotOutgoingChanges which uses the default.json
```
edmofro, line 2864:
```suggestion
      // The config override actually doesn't apply to snapshotOutgoingChanges which uses the default.json, so we need to mock it directly here
```

edmofro: What do you mean it uses `default.json` though?

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,965 @@ describe('CentralSyncManager', () => {
...
+            `No records found for record type ${model.tableName} in lookup table, Check the test setup!`,
+          );
+        }
+
+        expect(recordIds).not.toContain(sensitiveId);
+        expect(recordIds).toContain(nonSensitiveId);
+      };
+
+      // Basic encounter
+      it('wont sync sensitive encounters', async () => {
```
edmofro, line 2971:
```suggestion
      it("won't sync sensitive encounters", async () => {
```
etc. below

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,965 @@ describe('CentralSyncManager', () => {
...
+    const lookupEnabledConfig = {
+      sync: {
+        lookupTable: {
+          enabled: true,
+        },
+        maxRecordsPerSnapshotChunk: DEFAULT_MAX_RECORDS_PER_SNAPSHOT_CHUNKS,
+      },
+    };
+
+    beforeEach(async () => {
```
edmofro, line 2859: I think this could be a `beforeAll`, and save a lot of time?

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,965 @@ describe('CentralSyncManager', () => {
...
+        expect(encounterIds).not.toContain(sensitiveEncounterB.id);
+      });
+
+      it.todo(
+        'will leave historical data unsynced when the facility changes from sensitive to non-sensitive until it is edited',
+      );
+
+      it.todo(
+        'will leave historical data unsynced when the facility changes from non sensitive to sensitive',
+      );
```
edmofro, line 3799: 👀

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2838,4 +2838,965 @@ describe('CentralSyncManager', () => {
...
+        );
+
+        const outgoingChanges = await centralSyncManager.getOutgoingChanges(sessionId, {});
+        const encounterIds = outgoingChanges
+          .filter(c => c.recordType === 'encounters')
+          .map(c => c.recordId);
+
+        expect(encounterIds).toContain(sensitiveEncounterA.id);
+        expect(encounterIds).not.toContain(sensitiveEncounterB.id);
+      });
```
edmofro, line 3791: Very quick skim through these tests, didn't really read them after the first few as they were all so similar

edmofro (changes requested):
Mostly looking great! A few suggestions

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
tcodling: oh it isnt!

tcodling:
@edmofro Ready for review. If possible I was thinking of working on those edge case tests in a seperate PR so I can get Klaus started testing asap? Or do you think its a bit risky in terms of potentially having to double back on testing?

packages/database/src/sync/buildEncounterLinkedSyncFilter.ts
edmofro: Only thing is that the comment makes it more obvious that `encounters` needs to be included, though that could be done through a differently structured comment

edmofro: I am totally on board with the idea of letting the type do the talking in general though!

packages/central-server/__tests__/sync/CentralSyncManager.test.js
```diff
@@ -2856,12 +2856,12 @@ describe('CentralSyncManager', () => {
       },
     };
 
-    beforeEach(async () => {
+    beforeAll(async () => {
```
edmofro, line 2859: Did this speed things up?

packages/database/src/sync/buildEncounterLinkedLookupFilter.ts
```diff
@@ -5,32 +5,25 @@ import {
...
-
 /**
- * Helper function to determine if a facility_id should be populated in sync lookup
+ * Helper to determine if a facility_id should be populated in sync lookup
  * Only populates facility_id when the encounter is from a sensitive facility
  * This ensures sensitive encounters are only synced to their originating facility
  */
-export function addSensitiveFacilityIdIfApplicable() {
-  return `
+const addSensitiveFacilityIdIfApplicable = `
```
edmofro, line 13: Caps snake case is our usual global const convention. That's why I was meaning it would be less readable, it's definitely not as pleasant to look at, but we'd better stick with the pattern!

packages/central-server/__tests__/sync/CentralSyncManager.test.js
tcodling: That actually broke everything! I have reverted for now but I will address in my cleanup card maybe?

edmofro: Sweet

edmofro (approved):
lgtm!

PR #7986, opened 2025-07-11
fix(upload): Allow browser to set content-type & add boundary to uploads

dannash100:
### Changes
Make sure no content-type set on form data upload

See https://stackoverflow.com/questions/39280438/fetch-missing-boundary-in-multipart-form-data-post
And here for more information on boundary https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2
### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/api-client/src/TamanuApi.js
+++ b/packages/api-client/src/TamanuApi.js
@@ -202,18 +202,18 @@ export class TamanuApi {
       ...otherConfig,
     };
 
+    // For fetch we have to explicitly remove the content-type header
+    // to allow the browser to add the boundary value
+    if (config.body instanceof FormData) {
+      config.headers.delete('content-type');
+    }
+
     if (
       config.body &&
       config.headers.get('content-type')?.startsWith('application/json') &&
       !(config.body instanceof Uint8Array) // also covers Buffer
     ) {
       config.body = JSON.stringify(config.body);
-    }
-
-    // For fetch we have to explicitly remove the content-type header
-    // to allow the browser to add the boundary value
-    if (config.body instanceof FormData) {
-      config.headers.delete('content-type');
     }
 
     const requestInterceptorChain = [];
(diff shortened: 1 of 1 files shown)


packages/api-client/src/TamanuApi.js
```diff
@@ -335,9 +341,7 @@ export class TamanuApi {
     return this.fetch(endpoint, undefined, {
       method: 'POST',
       body: formData,
-      headers: {
-        'Content-Type': 'multipart/form-data',
-      },
+      headers: {},
```
dannash100, line 344: We could leave content type here for clarity? I dunno - it would still get choppeed off

NavarroEmilioLuis (commented):
Sorry this is a very small PR and thanks for grabbing it but this issue is very sneaky as it wasn't picked up by unit tests - were you able to properly import stuff?

This should undergo test, could be on main - I don't love the current resolution and I left some comments about it but ultimately okay if everything works + time constraint is too tight to keep looking

packages/api-client/src/TamanuApi.js
```diff
@@ -210,6 +210,12 @@ export class TamanuApi {
       config.body = JSON.stringify(config.body);
     }
 
+    // For fetch we have to explicitly remove the content-type header
+    // to allow the browser to add the boundary value
+    if (config.body instanceof FormData) {
+      config.headers.delete('content-type');
+    }
+
```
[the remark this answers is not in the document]
dannash100: Yeah we chop it off after all defaulting essentially. 
The steps for content type are
- default to json
- override from params ( in this case none )
- ultimatly remove defaulted content-type


Ill have a think re cleaner

NavarroEmilioLuis (approved):
Sweet, I'm happy with that

packages/api-client/src/TamanuApi.js
passcod: remove the entire headers field imo

passcod (approved):

PR #8004, opened 2025-07-15
fix(programRegistry): SAV-992: Get rid of duplicate patient program registrations

NavarroEmilioLuis:
### Changes

The reason why we need to do this is because otherwise the `logs.changes` table will have "missing records" - while we will be getting rid of these, it's possible that they are updated and create orphan rows at `logs.changes`. If you follow the linear card you'll find more info there and also a slack thread detailing.

Note that the SQL itself was reviewed already by Juliana (although it needed a slight tweak, see second commit) and the migration is inserted to run before `1742866911419-addPatientProgramRegistrationId.ts`.

This needs to be hotfixed to 2.33 and up


dannash100 (approved):
Oh man this migration was so difficult ay. 💯 Thank you Emilio

PR #8013, opened 2025-07-15
feat: medication tweaks and fixes

khoi-bes:
### Changes

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/database/src/models/MedicationAdministrationRecord.ts
```diff
@@ -242,6 +262,129 @@ export class MedicationAdministrationRecord extends Model {
...
+          model: Encounter,
+          as: 'encounter',
+          attributes: ['id', 'encounterType', 'endDate'],
+        },
+      ],
+    });
+    const encounter = encounterPrescription?.encounter;
+
+    // Skip if this is not an inpatient encounter or is discharged
+    if (!encounter || encounter.encounterType !== ENCOUNTER_TYPES.ADMISSION || encounter.endDate)
```
edmofro, line 287: This and the prescription equivalent [above](https://github.com/beyondessential/tamanu/pull/8013/files#diff-1c26a0c89b35775f89438f7f9624cb3178a31d9331cb8e3bb8ff0c08e921e52eR272) seem unnecessarily defensive to me. If we're inside this function, we expect both a prescription and an encounter to exist, right? So I suggest using assertions that throw errors rather than conditional checks that silently swallow.

packages/facility-server/app/routes/apiv1/user.js
```diff
@@ -247,9 +247,98 @@ user.get(
...
+                          NOT EXISTS (
+                            SELECT 1
+                            FROM encounter_pause_prescriptions epp
+                            WHERE epp.encounter_prescription_id = ep.id
+                              AND epp.deleted_at IS NULL
+                              -- Check if pause overlaps with the MAR's time slot
+                              -- A pause overlaps if: pause_start < slot_end AND pause_end >= slot_end
+                              -- This matches the frontend logic in MarStatus.jsx line 188
+                              AND epp.pause_start_date::timestamp < mar_time_slot.end_time
+                              AND epp.pause_end_date::timestamp >= mar_time_slot.end_time
```
edmofro, line 333: What's the performance of this like at high data volumes?

packages/web/app/components/Medication/MedicationImportModal.jsx
```diff
@@ -131,8 +132,30 @@ export const MedicationImportModal = ({ encounter, open, onClose, onSaved }) =>
...
+      .filter(
+        p =>
+          !encounterPrescriptions.some(
+            ep =>
+              ep.medicationId === p.medicationId &&
+              ep.doseAmount === p.doseAmount &&
+              ep.units === p.units &&
+              ep.route === p.route &&
+              ep.frequency === p.frequency,
+          ),
```
edmofro, line 156: This inner loop could be avoided by creating a `Set` with a hash of the important details

packages/web/app/forms/DischargeForm.jsx
```diff
@@ -673,9 +684,22 @@ export const DischargeForm = ({
   const { data: encounterMedications } = useEncounterMedicationQuery(encounter.id);
   const { data: ongoingPrescriptions } = usePatientOngoingPrescriptionsQuery(encounter.patientId);
 
-  const activeMedications =
-    encounterMedications?.data?.filter(medication => !medication.discontinued) || [];
-  const onGoingMedications = ongoingPrescriptions?.data?.filter(p => !p.discontinued) || [];
+  const activeMedications = (encounterMedications?.data || []).filter(
+    medication => !medication.discontinued,
+  );
+  const onGoingMedications = (ongoingPrescriptions?.data || [])
```
edmofro, line 690: "ongoing" is one word, casing is correct for `ongoingPrescriptions` 
```suggestion
  const ongoingMedications = (ongoingPrescriptions?.data || [])
```

packages/web/app/forms/DischargeForm.jsx
```diff
@@ -673,9 +684,22 @@ export const DischargeForm = ({
...
+    .filter(p => !p.discontinued)
+    .filter(
+      p =>
+        !activeMedications.some(
+          am =>
```
edmofro, line 695: Same comment about Set of hashes

edmofro (approved):
Sync part looks good, in fact we have this check on the servers so it is very good we now have it on mobile too! What was the specific circular dependency you needed to solve for? 

I have some other comments for you to review, but approving this from the sync codeowner side

packages/database/src/models/MedicationAdministrationRecord.ts
khoi-bes: it's absolutely right

packages/facility-server/app/routes/apiv1/user.js
namhoangbes: @edmofro This is due to the complexity required to determine a pausing MAR. We could probably improve performance a bit by adding indexes. However I would recommend to simplify the pausing logic instead if possible.

edmofro: Do you know how it does perform with a lot of data?

I'm not sure about how the pausing logic works, do you think it is worth taking that up with @MeganLane29?

namhoangsts: I don't think it will degrade much with a large dataset. However it still worth revising the logic to reduce complexity.

namhoangbes: I don't think it will degrade much with a large dataset. However it still worth revising the logic to reduce complexity.

MeganLane29: Happy to disucss further in our next refinement if you want to create a card @namhoangbes ?

khoi-bes: @MeganLane29, this is the complicated logic: when the next MAR is recorded and the user pauses the medication on the current MAR, the current MAR will not be marked as paused., so we must ensure the corresponding task has been showed correctly on the dashboard in this case

PR #8014, opened 2025-07-16
feat(audit): SAV-708: Include new column that stores reason for change

NavarroEmilioLuis:
### Changes

Desktop charting needs the ability to be edited (survey response answers) while keeping track of the edit history. The decision was to reuse the audit work for this. We need to keep track of previous vales, record the user that made the change, the time it was made and the reason for change, all are already part of the `logs.changes` table except the reason for change.

Here, the proposal is to add a new field that tracks the changelog `reason`, which we can set explicitly whenever appropriate, otherwise it will default to `unknown`, which is somewhat adequate (we always know how it was logged, not why).

With the current state of things, the only records that will have a non-default `reason` will be survey response answers and vital logs. The latter which should eventually be removed once we migrate vitals to keep track via audit.

### DIFF
--- a/database/model/logs/changes.md
+++ b/database/model/logs/changes.md
@@ -84,3 +84,7 @@ Note that as this is `JSONB`, some type information may be lost. However, row
 data in a Tamanu system is transported using JSON via the sync system anyway, so
 it is expected that all data trivially round-trips via JSON.
 {% enddocs %}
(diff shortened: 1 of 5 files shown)


dannash100 (approved):
Just small tweak! Good thinking

PR #8016, opened 2025-07-16
feat(portal): PHX-237: Patient portal dashboard endpoints

flynncs:
### Changes

Adds new routes for data fetching from patient portal dashboard frontend.

2500/3000 LOC are tests but let me know if it's too much

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/shared/src/schemas/patientPortal/responses/administeredVaccine.schema.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UserSchema } from './user.schema';
+import { LocationSchema } from './location.schema';
+import { VACCINE_STATUS, INJECTION_SITE_VALUES } from '@tamanu/constants';
+
+// Schema for administered vaccines returned to patient portal
+export const AdministeredVaccineSchema = z.object({
+  id: z.string(),
+  // Vaccine administration details
+  batch: z.string().nullish(),
+  status: z.enum(VACCINE_STATUS),
```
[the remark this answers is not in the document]
flynncs: `z.nativeEnum` is now deprecated and `z.enum` takes its functionality

edmofro:
bugbot run

packages/central-server/app/patientPortal/patientData/appointments.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const appointments = await models.Appointment.findAll({
+    where: {
+      patientId: patient.id,
+      locationGroupId: {
+        [Op.not]: null,
+      },
+      [Op.and]: [
+        where(literal('"Appointment"."start_time"::TIMESTAMP'), Op.gte, literal("DATE 'today'")),
+      ],
```
NavarroEmilioLuis, line 56: I don't think we need this `AND`? Also noting here that today won't account for time?

packages/central-server/app/patientPortal/patientData/ongoingPrescriptions.js
```diff
@@ -0,0 +1,40 @@
...
+    attributes: getAttributesFromSchema(OngoingPrescriptionSchema),
+    where: {
+      discontinued: false,
+    },
+    include: [
+      {
+        model: models.PatientOngoingPrescription,
+        as: 'patientOngoingPrescription',
+        where: { patientId: patient.id },
+        attributes: [],
```
NavarroEmilioLuis, line 20: Hmmm I'm not sure what would happen if we don't specify attributes here? If the intent is to ensure the join I'd use inside the where `patientOngoingPrescription: { [Op.ne]: null }`?

packages/central-server/app/patientPortal/patientData/vaccinations.js
```diff
@@ -0,0 +1,90 @@
...
+
+  res.send({
+    data: upcomingVaccinations.map(vaccination => UpcomingVaccinationSchema.parse(vaccination)),
+  });
+});
+
+export const getAdministeredVaccines = asyncHandler(async (req, res) => {
+  const { patient } = req;
+
+  const administeredVaccines = (await patient.getAdministeredVaccines())?.data;
```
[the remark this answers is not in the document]
NavarroEmilioLuis: AFAIK this can't be possible so not needed, but also means that we shouldn't be using the conditional operator. We know the function always returns an object with `data` property - it will be an empty array if nothing is found.

packages/central-server/app/patientPortal/patientData/vaccinations.js
```diff
@@ -0,0 +1,88 @@
...
+
+  res.send({
+    data: administeredVaccines.map(vaccine => AdministeredVaccineSchema.parse(vaccine)),
```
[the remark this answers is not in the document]
NavarroEmilioLuis: Not true once we solve the above

packages/central-server/app/patientPortal/patientData/ongoingConditions.js
```diff
@@ -0,0 +1,29 @@
...
+
+export const getOngoingConditions = asyncHandler(async (req, res) => {
+  const { patient } = req;
+  const { models } = req.store;
+
+  const ongoingConditions = await models.PatientCondition.findAll({
+    where: {
+      patientId: patient.id,
+      resolved: false, // Excluding resolved conditions for now
+    },
```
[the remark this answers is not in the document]
NavarroEmilioLuis: This is true, while unlikely here, cause `PatientCondition` has a default value of false, that's not the case with prescriptions and discontinued! So probably use `{ [Op.ne]: true }` instead, which will match nulls too.

...Unless we explicitly just truly want to get the ones assigned with `false`. However real data state might not be as clean as unit tests!

packages/central-server/app/utils/schemaUtils.js
```diff
@@ -0,0 +1,24 @@
+import { z } from 'zod';
```
NavarroEmilioLuis, line 1: Just flagging here that I don't really know this library or its usage. Overall looks good!

A couple of questions would be why not `yup`, or, given the shared packages use typescript what's the added value in these - initially I thought that we were going to predefine specific attributes for the queries but it seems we are almost retrieving everything everywhere.

This is more for my information really, but also will leave a trail of intent and justification.

NavarroEmilioLuis (changes requested):
First of all. Amazing work, it's a huge chunk of work. I really like the consistency of everything, very clean and organized.

There are a couple of changes (the one from the bots really) that probably need to happen. I also left a couple of notes here and there.

packages/central-server/app/utils/schemaUtils.js
flynncs: yeah fair questions, from my understanding the benefits that zod gives us are:

- runtime validation rather than just compile time w/ typescript
- will be really useful in many parts of the repo (fake data generation for E2E tests, synthetic tests, storybook, openAPI doc generation, type safety on frontend as we move to tsx)
- in terms of `yup`, `zod` (from what I've read) has better TS integration, is faster, smaller, and more concise in how you declare schemas

 in the case of the patient portal endpoints, I was intending to use zod predominantly to ensure that we know exactly what data the patient portal client is getting, just make sure we're not sending anything confidential / that we don't want a public user to know from raw DB queries.
 I noticed that we could also infer exactly what fields / attributes we wanted to query from models, but in practice I'm not sure if this is any more performant due to overhead

NavarroEmilioLuis (approved):
I'm happy with the changes, but I'm still not sure about that empty `attributes` array. I'm pre-approving this, but would be great to get confirmation it's working as intended

packages/central-server/app/patientPortal/patientData/ongoingPrescriptions.js
flynncs: yup - I've double checked the result of the query and will leave a comment to explain to any future readers :)

PR #8077, opened 2025-07-24
feat(desktopCharting): SAV-695: Add ability to edit complex charts core data

NavarroEmilioLuis:
### Changes

Pretty much what the title says. This workflow is... different. So I created a specific endpoint to avoid being able to edit other survey responses, which are usually very involved to create. Because the complex chart core is a very simple survey with text fields only (and 1 datetimestring), plus, it shouldn't have any complex actions like calculations or write to patient data, etc., we can be confident on updating the answers this way.

### DIFF
--- a/packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
+++ b/packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
@@ -351,10 +351,24 @@ describe('SurveyResponse', () => {
       const newValue1 = 'New answer 1';
       const newValue2 = 'New answer 2';
 
+      // Also create a missing answer
+      const newAnswerDataElement = await ProgramDataElement.create({
+        ...fake(ProgramDataElement),
+        type: PROGRAM_DATA_ELEMENT_TYPES.TEXT,
+      });
+      await SurveyScreenComponent.create({
+        ...fake(SurveyScreenComponent),
+        dataElementId: newAnswerDataElement.id,
+        surveyId: response.surveyId,
+        calculation: '',
+      });
+      const newAnswerValue = 'New answer value';
+
       const result = await app.put(`/api/surveyResponse/complexChartInstance/${response.id}`).send({
         answers: {
           [answer1.dataElementId]: newValue1,
           [answer2.dataElementId]: newValue2,
+          [newAnswerDataElement.id]: newAnswerValue,
         },
       });
       await answer1.reload();
--- a/packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
+++ b/packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
@@ -1,5 +1,6 @@
 import { disableHardcodedPermissionsForSuite } from '@tamanu/shared/test-helpers';
 import { fake } from '@tamanu/fake-data/fake';
+import { PROGRAM_DATA_ELEMENT_TYPES, SURVEY_TYPES } from '@tamanu/constants';
 
 import { createTestContext } from '../utilities';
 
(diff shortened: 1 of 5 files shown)


packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -81,3 +83,53 @@ surveyResponse.post(
...
+          await existingAnswer.update({ body: value });
+        } else {
+          await models.SurveyResponseAnswer.create({
+            dataElementId,
+            body: value,
+            responseId: params.id,
+          });
+        }
+      }
+    });
```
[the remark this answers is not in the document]
NavarroEmilioLuis: While this is interesting, the complex chart core survey can only ever have 4 question max so it's not that big of a deal. On the other hand this introduces a new pattern that I don't think it's worth the trouble

edmofro:
bugbot run

packages/web/app/components/Charting/ChartInstanceInfoSection.jsx
```diff
@@ -15,73 +26,144 @@ const ChartInstanceInfoLabel = styled(TranslatedText)`
...
+  const handleEdit = async ({ survey, ...data }) => {
+    const responseData = {
+      answers: await getAnswersFromData(data, survey),
+    };
+
+    await api.put(`surveyResponse/complexChartInstance/${chartInstanceId}`, responseData);
+    //queryClient.invalidateQueries(['encounterCharts', encounter.id, chartSurveyId]);
+    queryClient.invalidateQueries(['encounterComplexChartInstances', encounter.id, chartSurveyId]);
+    setIsEditModalOpen(false);
+  };
```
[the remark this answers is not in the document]
NavarroEmilioLuis: `chartInstanceId` is in fact a survey response ID - the one from the chart instance. I'll give you that the commented out should be removed, however, it won't display stale data after an edit cause that's not part of encounter charts, we're editing the instance only (which we are invalidating).

packages/web/app/components/ComplexChartModal.jsx
```diff
@@ -22,24 +21,16 @@ export const ComplexChartModal = ({
...
-          date={chartDate}
-          type={chartType}
-          subtype={chartSubtype}
-          isTypeVisible={isTypeVisible}
-          isSubtypeVisible={isSubtypeVisible}
-          data-testid="styledchartinstanceinfosection-y5ji"
+          complexChartInstance={complexChartInstance}
+          fieldVisibility={fieldVisibility}
+          patient={patient}
+          selectedChartSurveyName={selectedChartSurveyName}
```
edmofro, line 33: We've lost the testid, do we need one?

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -81,3 +84,54 @@ surveyResponse.post(
...
+      params,
+      db,
+    } = req;
+
+    const responseRecord = await models.SurveyResponse.findByPk(params.id);
+    if (!responseRecord) {
+      throw new NotFoundError('Response record not found');
+    }
+
+    req.checkPermission('write', subject('Charting', { id: responseRecord.surveyId }));
```
edmofro, line 103: IMO this should go before the lines above, because:
1. If they don't have write perms, it saves a db query which is a tiny performance nicety
2. They get information, albeit benign, by letting them know if they've hit an id that exists or not before checking any permissions

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -81,3 +84,54 @@ surveyResponse.post(
...
+    if (survey.surveyType !== SURVEY_TYPES.COMPLEX_CHART_CORE) {
+      throw new InvalidOperationError('Cannot edit survey responses');
+    }
+
+    const components = await models.SurveyScreenComponent.getComponentsForSurvey(survey.id);
+    const responseAnswers = await models.SurveyResponseAnswer.findAll({
+      where: { responseId: params.id },
+    });
+
+    await db.transaction(async () => {
```
edmofro, line 115: Nice one wrapping this in a transaction

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -81,3 +84,54 @@ surveyResponse.post(
...
+      for (const [dataElementId, value] of Object.entries(body.answers)) {
+        // Ignore null values or components that are not in the survey
```
edmofro, line 117: Should we error on components that aren't in the survey? That seems like something has gone quite wrong? I guess maybe they filled it out and it has since been updated to remove a question?

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -81,3 +84,54 @@ surveyResponse.post(
     res.send(responseRecord);
   }),
 );
+
+surveyResponse.put(
+  '/complexChartInstance/:id',
```
edmofro, line 89: Does `complexChartInstance` similarly have its own `POST` endpoint?

packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
```diff
@@ -82,6 +83,71 @@ describe('SurveyResponse', () => {
...
+    });
+    const answer = await SurveyResponseAnswer.create({
+      ...fake(SurveyResponseAnswer),
+      dataElementId: dataElement.id,
+      responseId: response.id,
+      body: 'Initial answer',
+    });
+
+    return { answer, response, dataElement, survey };
+  };
```
edmofro, line 149: Skimmed only

packages/facility-server/__tests__/apiv1/SurveyResponse.test.js
```diff
@@ -196,6 +262,152 @@ describe('SurveyResponse', () => {
...
+          [answer.dataElementId]: null,
+        },
+      });
+      await answer.reload();
+
+      expect(result).toHaveSucceeded();
+      expect(answer.body).toEqual(originalValue);
+    });
+
+    it('should update multiple answers in a single request', async () => {
```
edmofro, line 330: Might be nice to have a new answer or two created in the same hit as the multiple updates?

edmofro (changes requested):
Really lovely thorough and clean tests
Great call making it use a new endpoint
Just a few comments but looking good!

packages/facility-server/app/routes/apiv1/surveyResponse.js
NavarroEmilioLuis: I don't think we would save any db query? We need the `responseRecord` to check the permission, otherwise we don't know the `surveyId`. The second point is true, it's a bit leaky, so I'll just put it in between

packages/facility-server/app/routes/apiv1/surveyResponse.js
NavarroEmilioLuis: No, it's just the same for regular survey responses. The thing is that we don't edit survey responses (and should be kept that way for now)

packages/facility-server/app/routes/apiv1/surveyResponse.js
edmofro: True...ok I think some benign leakiness is better than that wrong permission error which is plain confusing. Back to the way it was! And totally right about that db query still being needed, missed that the id was used in the permission check 👌

edmofro (approved):
lgtm!

PR #8088, opened 2025-07-27
feat(sensitiveFacilities): NASS-1728: Incorporate facility sensitivity to login logic

tcodling:
### Changes

Updating the `user_facility` linked login logic to consider facility sensitivity.

For the new sensitivity logic, we want to check against `user_facilities` for any login to a sensitive facility regardless of the settings or permissions.

Key changes
- Now only super user gets the special "ALL" key that lets them bypass logic and access all facilities
- Even with the `restrictUsersToFacilities` setting off, we still check for `user_facility` links if the facility is sensitive
- Login permission does NOT override the `user_facility` link if the facility is sensitive

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/database/src/models/User.ts
```diff
@@ -295,22 +295,26 @@ export class User extends Model {
...
-    if (allowed === CAN_ACCESS_ALL_FACILITIES) return true;
+    const facility = await this.sequelize.models.Facility.findByPk(id);
 
-    return allowed?.includes(id) ?? false;
+    if (facility?.isSensitive && !(await this.checkCanAccessAllFacilities())) {
+      const allowed = await this.allowedFacilityIds();
+      if (allowed === CAN_ACCESS_ALL_FACILITIES) return true;
+
+      return allowed?.includes(id) ?? false;
+    }
```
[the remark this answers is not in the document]
tcodling: there is a return true under though

packages/database/src/models/User.ts
```diff
@@ -265,21 +265,8 @@ export class User extends Model {
...
-    // Allow for roles that have access to all facilities configured via permissions
-    // (e.g. a custom "AdminICT" role)
-    if (await this.hasPermission('login', 'Facility')) return true;
-    return false;
-  }
-
   async allowedFacilities() {
-    const canAccessAllFacilities = await this.checkCanAccessAllFacilities();
-    if (canAccessAllFacilities) {
+    if (this.isSuperUser()) {
```
tcodling, line 269: Super user is the only one that gets to use the special `CAN_ACCESS_ALL_FACILITIES` key now so no need for `checkCanAccessAllFacilities()`

packages/constants/src/auth.ts
```diff
@@ -6,3 +6,4 @@ export const JWT_TOKEN_TYPES = {
 export const SYSTEM_USER_UUID = '00000000-0000-0000-0000-000000000000';
 
 export const CAN_ACCESS_ALL_FACILITIES = 'ALL';
+export const CAN_ACCESS_ALL_NON_SENSITIVE_FACILITIES = 'ALL_NON_SENSITIVE';
```
tcodling, line 9: Would kind of prefer if we didn't need these magic constants and could infer. I did have a quick go at this but it starting feeling complicated quickly and was starting to blow out the scope so I tried to just clean up and follow the existing pattern

packages/database/src/models/User.ts
```diff
@@ -265,52 +266,56 @@ export class User extends Model {
...
-      return CAN_ACCESS_ALL_FACILITIES;
-    }
+    const { Facility, Setting } = this.sequelize.models;
+
+    if (this.isSuperUser()) return CAN_ACCESS_ALL_FACILITIES;
+
+    const restrictUsersToFacilities = await Setting.get('auth.restrictUsersToFacilities');
+    const hasLoginPermission = await this.hasPermission('login', 'Facility');
 
+    // Get user's linked facilities
```
edmofro, line 277: Feels like there's a performance optimisation in here where you check if there are any sensitive facilities at all, and return `CAN_ACCESS_ALL_FACILITIES` if not, before reloading their facilities. That would improve performance here, but also any time we check their list of facilities (because it can just match that const rather than looking through a potentially long list)

packages/mobile/App/models/User.ts
```diff
@@ -63,26 +67,44 @@ export class User extends BaseModel implements IUser {
     return this.role === 'admin' || this.id === SYSTEM_USER_UUID;
   }
 
-  async allowedFacilityIds() {
-    const canAccessAllFacilities = this.isSuperUser();
-    if (canAccessAllFacilities) {
```
edmofro, line 68: Good call here

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
+      await linkedUser.reload({ include: 'facilities' });
     });
 
     describe('allowedFacilities', () => {
       it('should get special "ALL" key when superuser', async () => {
         const superUserFacilities = await superUser.allowedFacilities();
         expect(superUserFacilities).toBe(CAN_ACCESS_ALL_FACILITIES);
       });
+
+      it('user with login permission should return all non-sensitive facilities plus linked facilities', async () => {
```
edmofro, line 418: Should read as a sentence starting with "it"
```suggestion
      it('should return all non-sensitive facilities plus linked facilities for a user with login permission', async () => {
```

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
+        mockLoginFacilityPermission(linkedUser, true);
+        const userFacilities = await linkedUser.allowedFacilities();
+        expect(userFacilities).toEqual(expect.arrayContaining(expectedCombinedFacilities));
+      });
+
+      it('setting disabled should return all non-sensitive facilities plus linked facilities', async () => {
```
edmofro, line 424:
```suggestion
      it('should return all non-sensitive facilities plus linked facilities when setting is disabled', async () => {
```

edmofro: Same below

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
       it('should return linked facilities from user_facilities table', async () => {
-        const userFacilities = await user.allowedFacilities();
-        expect(userFacilities).toStrictEqual(validUserFacilities);
+        const userFacilities = await linkedUser.allowedFacilities();
+        expect(userFacilities).toStrictEqual(linkedUserFacilities);
       });
-      it('should return empty array if no linked facilities', async () => {
-        const userFacilities = await noFacilityUser.allowedFacilities();
+
+      it('should return empty array if no linked facilities with setting on', async () => {
```
edmofro, line 435:
```suggestion
      it('should return an empty array if there are no linked facilities when restrictUsersToFacilities is enabled', async () => {
```
Again these it descriptions are conventionally very sentencey

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
     describe('allowedFacilityIds', () => {
       it('should get special "ALL" key when superuser', async () => {
-        jest
-          .spyOn(superUser, 'allowedFacilities')
-          .mockImplementation(() => CAN_ACCESS_ALL_FACILITIES);
         const superUserFacilityIds = await superUser.allowedFacilityIds();
         expect(superUserFacilityIds).toBe(CAN_ACCESS_ALL_FACILITIES);
       });
+
+      it('login permission should return all non-sensitive facilities plus linked facilities', async () => {
```
edmofro, line 453: the "login permission" idea is pretty murky and weird I reckon, would be worth describing it a little more specifically

edmofro: (this might not be the place for that)

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
   });
 
   describe('User facility methods', () => {
-    let user = null;
     let superUser = null;
-    let noFacilityUser = null;
+    let linkedUser = null;
+    let unlinkedUser = null;
```
edmofro, line 360: Personally don't love the term "linked" to describe this concept, it's about access/permission/allowedness and there are words for that, whereas linking could mean a lot of different things.
```suggestion
    let userWithAccess = null;
    let userWithNoAccess = null;
```

edmofro: Honestly for later use this is easier
```suggestion
    let user1 = null;
    let userWithNoAccess = null;
```

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
 
-    const validUserFacilities = [facility1, facility2];
-    const validUserFacilityIds = validUserFacilities.map((f) => f.id);
+    const nonSensitiveFacilities = [facility1, facility2, facility3];
+    const nonSensitiveFacilityIds = nonSensitiveFacilities.map(f => f.id);
 
-    beforeAll(async () => {
+    const linkedUserFacilities = [facility1, sensitiveFacility1];
+    const linkedUserFacilityIds = linkedUserFacilities.map(f => f.id);
```
edmofro, line 366: Again not a fan of `linked` (I know you've just taken it and run with it, that descriptor already existed). Additionally, when these refer to facilities, calling then a name that includes `userFacilities` is a bit confusing because there's another database table that is very involved in this area which is called `user_facilities`.

```suggestion
    const user1Facilities = [facility1, sensitiveFacility1];
    const user1FacilityIds = linkedUserFacilities.map(f => f.id);
```

packages/facility-server/__tests__/apiv1/User.test.js
```diff
@@ -340,108 +355,144 @@ describe('User', () => {
...
+
+    const expectedCombinedFacilities = [...nonSensitiveFacilities, sensitiveFacility1];
+    const expectedCombinedFacilityIds = expectedCombinedFacilities.map(f => f.id);
```
edmofro, line 369: There's a pattern of "arrange / act / assert" that tests generally follow, and it feels a bit weird to me to have the data that will be asserted in the arrange section. I'm not sure if that's just me though, have a chat to AI about it.

edmofro (changes requested):
Sorry I went a little ham when I got to the tests, they're a great test suite and it's more about conventions!

packages/facility-server/__tests__/apiv1/User.test.js
tcodling: Yep I can update. Personally not a fan of the userWithNoAccess though because to me that applies that user should never be able to access anything when there are many cases it can

edmofro: Yeah no worries, that's a great point. Go with whatever name you like, I just prefer to avoid the `link` idea

edmofro: Agree and agree! (I didn't see that comment before I posted mine)

packages/facility-server/__tests__/apiv1/User.test.js
tcodling: Yea I never really know how to go about writing out a permission like this even in linear cards. All i can really think of it to lay it out more explicitly something like this?

`it('[verb: "login", noun: "Facility"] permission should return all ...`

packages/facility-server/__tests__/apiv1/User.test.js
tcodling: Ah yep I see what you mean. I think maybe i went overboard with trying to make the code as dry as possible but in this case i totally agree with splitting it out

packages/facility-server/__tests__/apiv1/User.test.js
edmofro: How about `it('the "login" permission to "Facility" should....`

On Linear I'd go "the `login` permission to `Facility`" personally but yeah agree it can be hard to articulate

edmofro: The reason I don't like the bracketed version btw is because again it takes away the smooth-sentence feel that test descriptions should have

tcodling:
bugbot run

tcodling:
bugbot run

packages/facility-server/__tests__/apiv1/User.test.js
tcodling: wait `user_facilities` **IS** the table that we are getting this list from here? So isn't it suitable to have the name include `userFacilities`?

tcodling: Ok now I am confused 😆 where else would it come from? I thought the entire point of the `user_facilities` table was to link users to facilities for login restrictions?

image

What am i missing?

edmofro: Yeah, the name of the array should just say what's inside it

packages/facility-server/__tests__/apiv1/User.test.js
tcodling: Moved as much as I could into the individual cases to follow this pattern. There are a few things cursor was keen to keep higher up even if used in both an "arrange" and "assert" mostly for dry reasons

tcodling:
bugbot run

packages/database/src/models/User.ts
```diff
@@ -273,27 +273,32 @@ export class User extends Model {
...
+    const sensitiveFacilities = await Facility.count({ where: { isSensitive: true } });
+    if (hasAllNonSensitiveFacilityAccess && sensitiveFacilities === 0)
+      return CAN_ACCESS_ALL_FACILITIES;
 
     // Get user's linked facilities
     if (!this.facilities) {
       await this.reload({ include: 'facilities' });
     }
-    const userLinkedFacilities = this.facilities?.map(({ id, name }) => ({ id, name })) ?? [];
+    const userFacilities = this.facilities?.map(({ id, name }) => ({ id, name })) ?? [];
```
edmofro, line 286: Eeek sorry again with the naming nits, but I want to avoid this being read as containing user_facilities records too 
```suggestion
    const explicitlyAllowedFacilities = this.facilities?.map(({ id, name }) => ({ id, name })) ?? [];
```
Honestly don't love the name, but have tried a few and that's my best shot

packages/mobile/App/models/User.ts
```diff
@@ -85,8 +90,8 @@ export class User extends BaseModel implements IUser {
     });
     const userLinkedFacilityIds = userFacilities.map(f => f.facilityId);
```
edmofro, line 91: You've chosen not to update the naming here, and honestly I'm fine with it after seeing where we ended up on the server 😂

packages/mobile/App/services/auth/AuthService.ts
```diff
@@ -66,6 +66,13 @@ export class AuthService {
     { email, password }: SyncConnectionParameters,
     generateAbilityForUser: (user: User) => PureAbility,
   ): Promise<IUser> {
+    const configFacilityId = await readConfig('facilityId', null);
+    if (!configFacilityId) {
```
edmofro, line 70:
```suggestion
    const deviceFacilityId = await readConfig('facilityId', null);
    if (!deviceFacilityId) {
```
Just a personal preference, doesn't matter either way

packages/mobile/App/services/auth/AuthService.ts
```diff
@@ -66,6 +66,13 @@ export class AuthService {
...
+      throw new Error(
+        'You need to first link this device to a facility before you can login offline.',
```
edmofro, line 72: Good call on the assertion!

edmofro (approved):
One tiny nit but tbf deserves to be approved, lgtm!

packages/mobile/App/models/User.ts
tcodling: Not chosen! Just forgot to port the change over to mobile 🙏

PR #8105, opened 2025-07-29
feat(logs): add script to generate changelog import inserts

passcod:
### Changes

This firstly commits the script to the repo but also makes a couple changes to the original:
- by eliminating all joins we see a significant throughput improvement and avoid any sort of locking on tables that are not `logs.changes` or the source table
- we no longer skip rows that have a null `created_at` but instead coalesce from the `updated_at` (and vice versa) to provide a decent approximation without omitting data; this also makes it so that there's no `where` predicate on the select, which means postgres will not need to do any processing, just a full table scan
- we pull the `updated_at_sync_tick` (saw that in the thread but it wasn't in the import script we ran in fiji prime...)

Also, we get the list of tables from the dbt manifest.json, which is obtainable via

```
bestool tamanu download dbt-manifest 2.35 --info manifest.json
```

or locally by:

```
npm run dbt-generate-model
npm run dbt-generate-docs
cp database/docs/target/manifest.json .
```

Before:

```sql
-- Insert statement for administered_vaccines
insert into logs.changes (table_oid, table_schema, table_name, logged_at, updated_by_user_id,
        record_id, record_created_at, record_updated_at, record_deleted_at, record_data)
select
        c.oid as table_oid,
        c.relnamespace::regnamespace as table_schema,
        c.relname as table_name,
        t.updated_at as logged_at,
        uuid_nil() as updated_by_user_id,
        t.id::text as record_id,
        t.created_at as record_created_at,
        t.updated_at as record_updated_at,
        t.deleted_at as record_deleted_at,
        to_jsonb(t) as record_data
from administered_vaccines t
join pg_class c on c.relname = 'administered_vaccines'
        and c.relkind = 'r'
    and c.relnamespace::regnamespace::text = 'public'
where t.created_at is not null;
```

After:

```sql
insert into logs.changes (
  table_oid, table_schema, table_name,
  logged_at, record_created_at, record_updated_at, record_deleted_at,
  updated_at_sync_tick, updated_by_user_id, record_id,
  record_data
)
select
  'public.administered_vaccines'::regclass::oid as table_oid,
  'public' as table_schema,
  'administered_vaccines' as table_name,

  t.updated_at as logged_at,
  coalesce(t.created_at, t.updated_at) as record_created_at,
  coalesce(t.updated_at, t.created_at) as record_updated_at,
  t.deleted_at as record_deleted_at,

  t.updated_at_sync_tick as updated_at_sync_tick,
  uuid_nil() as updated_by_user_id,
  t.id::text as record_id,

  to_jsonb(t) as record_data
from public.administered_vaccines t;
```

In Palau Clone this was able to process an average of a million rows per minute.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/scripts/src/generateChangelogImports.mjs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const START = 'select now() as started;\n\n';
+const END = '\n\nselect now() as ended; select count(*) from logs.changes;';
+
+async function generate() {
+  return START + (await getTables())
+    .map(table =>
+      `
+      \echo ${table}
```
NavarroEmilioLuis, line 54: Hmmmmm I guess try to move \echo to the same line with the ``` symbol? This looks like a linter is tripping

.github/workflows/cd-changelog-import-script.yml
```diff
@@ -0,0 +1,101 @@
+name: CD Changelog Import Script
+
+on:
+  workflow_dispatch:
+  push:
+    tags: ['v*.*.*']
+    branches:
+      - main
+      - release/*
```
NavarroEmilioLuis, line 9: Hmmmmmmm sorry I don't really follow - I thought this was to insert initial `logs.changes` when we deploy? But it seems weird that we will run on every push? Then, we upload that file and I guess we grab that and run directly on the deployment.

I think that it might make sense on tags, because we'll only push on release. But main and releases we kinda just care about the first one? Cause it will never ever be "initial" after being deployed?

I guess we might not ever run those files but seems like a waste? Maybe I got something wrong though

packages/scripts/src/generateChangelogImports.mjs
```diff
@@ -0,0 +1,85 @@
...
+  'socket_io_attachments',
+
+  // deprecated
+  'note_items',
+  'note_pages',
+  'notes_legacy',
+  'vitals',
+
+  // write-once tables
+  'attachments',
```
NavarroEmilioLuis, line 17: This one won't be write once soon (though I wonder if right to be forgotten makes it so that we should also get rid of the data on the log... yikes). Makes me wonder why exclude these though - if we ever edit them, they'll still get a `log.change` right? But it'll look weird.

I guess we can leave out the really noisy ones if that's what we want to avoid?

NavarroEmilioLuis (approved):
I don't see anything wrong, but I'm not completely sure I understand what we are doing. Code looks clean though, no notes there. I don't think this needs a re-review so I'll approve now

packages/scripts/src/generateChangelogImports.mjs
passcod: Attachments in particular is just very very large, it would add tens of gigabytes for most deployments.

.github/workflows/cd-changelog-import-script.yml
passcod: It's more as a way to verify that the output is right before we publish a release. It costs a few minutes of free compute, so, nbd.

passcod:
Just going to mention that besides the issues (thanks bugbot) this is pending further benchmarking and possible optimisations discussed in slack.

passcod:
Dropping this for now as we've got a process that works without.

PR #8106, opened 2025-07-30
fix(desktopCharting): SAV-692: Photo component QA

NavarroEmilioLuis:
### Changes

- Selecting, deselecting and reelecting the same file didn't work - we were clearing out react state but not DOM, so the on change event wasn't really firing.
- Update button text
- Update legend text
- Update package.json and lock to be able to import from CASL ability


flynncs (approved):

tcodling:
bugbot run

PR #8109, opened 2025-07-30
fix(web): session restore logic and spurious network request failed error

passcod:
### Changes

Turns out this is the same thing: we store the session token but we really wanted to store the refresh token, and also we weren't disabling the error output when trying to restore the session (and failing due to outdated session token).

### Testing

Go to the deploy, and login once.

Refresh, and you should:
- be logged in without having to enter credentials again
- not see a network error toast

Then logout, and refresh. You should:
- not be logged in, and have to enter creds
- still not see a network error toast

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/api/TamanuApi.jsx
+++ b/packages/web/app/api/TamanuApi.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { TamanuApi as ApiClient, AuthExpiredError } from '@tamanu/api-client';
 import { ENGLISH_LANGUAGE_CODE, SERVER_TYPES } from '@tamanu/constants';
-import { buildAbilityForUser } from '@tamanu/shared/permissions/buildAbility';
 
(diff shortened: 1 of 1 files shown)


packages/web/app/api/TamanuApi.jsx
```diff
@@ -135,6 +131,11 @@ export class TamanuApi extends ApiClient {
     });
   }
 
+  setToken(token, refreshToken) {
+    localStorage.setItem(TOKEN, refreshToken);
+    return super.setToken(token, refreshToken);
+  }
```
passcod, line 137: by overriding this method we ensure we always save the token regardless of when this is called, saving us from separately handling all the different situations

dannash100 (approved):
Looks good 💯

edmofro:
bugbot run

PR #8117, opened 2025-07-31
db: better changelog indexes

passcod:
### Changes

Follows an investigation into indexing that table when it has real volumes of data. Full details [in slack](https://beyondessential.slack.com/archives/GE9NHB95F/p1753911199521059), summary:

> - hash indexes can be very fast to create/maintain!... if the input data varies a lot. That is, you want to minimise hash/bucket collision. So, record_id, super fast! takes 70 seconds to build. device_id, which is entirely identical in this data set? 6 hours. not a typo.
> - btree indexes are kinda the medium experience. There's no collisions, but they need balancing. OTOH btree is the only index type we have parallel build support for across all our versions (landed in pg 11, parallel BRIN indexes landed in 17, and parallel GIN are landing in 18 next year). The btree on created_at took 1h30.
> - `maintenance_work_mem` is critical! this is a database setting, by default it's set to 64MB. Lifting it to 1GB brings... btrees in 15 seconds, spread out hashes in 40 seconds. It doesn't help for hashes on collided data.
> - BRIN is a fun one. It's a lightweight partitioning index suited especially to time series data, like logged_at... which takes 16 seconds to build (4 seconds on 1GB work_mem), and adds up to 168KB of index data. yes, KB.
> - Now the chonker: the GIN index on record_data. Or is it? That index takes 7 minutes to build. There is a marginal (10%) improvement with a 1GB work_mem. Turns out, not the problem.
> 
> Thus, proposed solution:
> - switch logged_at, record_created_at, and record_updated_at to BRIN
> - keep record_id on hash and record_data on gin
> - move every other hash index to btree
> - set maintenance_work_mem to 1GB (at least for this, but I say we leave it there permanently unless we get a specific memory issue, to help with future migrations)

This delivers a 99.2% reduction in indexing time (19 hours to 10 minutes on the tested dataset).

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/migrations/1753999599109-betterIndexesForLogsChanges.ts
+++ b/packages/database/src/migrations/1753999599109-betterIndexesForLogsChanges.ts
@@ -3,45 +3,73 @@ import { QueryInterface } from 'sequelize';
 const TABLE = { schema: 'logs', tableName: 'changes' };
 
 export async function up(query: QueryInterface): Promise<void> {
-  await query.removeIndex(TABLE, 'changes_device_id');
-  await query.removeIndex(TABLE, 'changes_logged_at');
-  await query.removeIndex(TABLE, 'changes_record_created_at');
-  await query.removeIndex(TABLE, 'changes_record_updated_at');
-  await query.removeIndex(TABLE, 'changes_table_name');
-  await query.removeIndex(TABLE, 'changes_table_oid');
-  await query.removeIndex(TABLE, 'changes_updated_by_user_id');
+  // Drop existing indexes if they exist
+  await query.sequelize.query(`DROP INDEX IF EXISTS logs.changes_device_id;`);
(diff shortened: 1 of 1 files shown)


tcodling:
bugbot run

rohan-bes (approved):
Wow! Great stuff, I hadn't realised just how reliant on high cardinality in the dataset a `hash` index was! Also BRIN index is really cool :) hadn't seen that before.

Looks good to me, maybe the only thought that comes to mind is whether it'll be an issue that we still have the old indexes in place while we run the migrations leading up to this one? There's a few that might create a bunch of change logs, around the program registries stuff in particular :thinking: 

But it should probably be okay, I don't think it should be too much data.

passcod:
Yes and also it'd be quite a kerfuffle to insert a migration back before the patient regs import as we'd have to also rewrite the intervening migrations that add/remove columns from the logs.changes table.

PR #8132, opened 2025-08-03
feat(mobileSync): KOR-45: Prepared inserts and quick-react-native-sqlite

dannash100:
### Changes

**Cuts sync time in half agian! Around 9 or 10 times faster from main all up**

**Prepared insert**
- Generate a raw sql query from sync records 
  - way way way faster, typeorm slows stuff down alot
- Also did alot of type fixes in pull flow that hadn't shown up because of incorrect typescript config
  
**Custom sqlite engine**
- Uses a version of `react-native-quick-sqlite` that didn't require and upgrade of react-native!
  - Specifies version of cmake and a few tweaks to jest to mock it out as it dont like it
  - Upgrades seamlessly if database exists during my tests
  - Full run of migrations are so noticeably faster too - dev experience sped up noticably for me, and much faster
  
  
### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


tcodling:
bugbot run

packages/mobile/App/services/sync/utils/pullRecordsInBatches.ts
```diff
@@ -1,9 +1,12 @@
 import { calculatePageLimit } from './calculatePageLimit';
 import { SYNC_SESSION_DIRECTION } from '../constants';
 import { PullParams } from '../MobileSyncManager';
+import { type CentralServerConnection } from '../CentralServerConnection';
 
 export const pullRecordsInBatches = async (
-  { centralServer, sessionId, recordTotal, progressCallback = () => {} }: PullParams,
+  { centralServer, sessionId, recordTotal, progressCallback = () => {} }: PullParams & {
+    centralServer: CentralServerConnection;
```
edmofro, line 8: Why isn't this part of standard `PullParams`?

packages/mobile/App/services/sync/utils/saveIncomingChanges.ts
```diff
@@ -95,64 +103,62 @@ export const saveChangesForModel = async (
...
+    if (!recordsForModel.length) {
+      continue;
     }
+    modelChanges.push({
+      model,
+      records:
+        'sanitizePulledRecordData' in model
+          ? (model as any).sanitizePulledRecordData(recordsForModel)
+          : recordsForModel,
+    });
```
edmofro, line 121:
```suggestion
    const sanitizedData = 'sanitizePulledRecordData' in model
          ? (model as any).sanitizePulledRecordData(recordsForModel)
          : recordsForModel;
    modelChanges.push({
      model,
      records: sanitizedData,
    });
```

packages/mobile/App/services/sync/utils/preparedInsert.ts
```diff
@@ -0,0 +1,19 @@
+import { Repository } from "typeorm";
+import { DataToPersist } from "../types";
+
+/** 
+ * Much faster than typeorm bulk insert or save
+ * Prepare a raw query and execute it with the values
+ */
+export const preparedInsert = async (repository: Repository<any>, rows: DataToPersist[]) => {
```
edmofro, line 8:
```suggestion
export const executePreparedInsert = async (repository: Repository<any>, rows: DataToPersist[]) => {
```
or something...initial thought from the title was that this just prepared the prepared insert, but it runs it too (alternative is to make it just prepare the insert and then execute it from the call site)

packages/mobile/App/services/sync/utils/preparedInsert.ts
```diff
@@ -0,0 +1,19 @@
...
+    const tableName = repository.metadata.tableName;
+    const columns = Object.keys(rows[0]);
```
edmofro, line 10: Are we certain all rows will contain the same columns?

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "GPL-3.0-and-later AND BUSL-1.1",
   "type": "module",
   "scripts": {
-    "prepare": "patch-package",
+    "prepare": "patch-package && npm run postinstall --workspace @tamanu/mobile",
```
edmofro, line 11: Can you tell me about this one?

packages/mobile/App/services/sync/MobileSyncManager.ts
```diff
@@ -371,49 +366,43 @@ export class MobileSyncManager {
...
-  async pullInitialSync({
-    sessionId,
-    recordTotal,
-    pullUntil,
-    syncSettings,
-  }: PullParams): Promise<void> {
-    let totalPulled = 0;
+  async pullInitialSync({ sessionId, recordTotal, pullUntil }: PullParams): Promise<void> {
+    let totalSaved = 0;
     const progressCallback = (incrementalPulled: number) => {
```
edmofro, line 371:
```suggestion
    const progressCallback = (incrementalSaved: number) => {
```
To go with other renames

packages/mobile/App/services/sync/MobileSyncManager.ts
```diff
@@ -371,49 +366,43 @@ export class MobileSyncManager {
...
-      await this.postPull(transactionEntityManager, pullUntil);
+      try {
+        const incomingModels = getTransactingModelsForDirection(
+          this.models,
+          SYNC_DIRECTIONS.PULL_FROM_CENTRAL,
+          transactionEntityManager,
+        );
+        const sortedModels = await sortInDependencyOrder(incomingModels) as TransactingModel[];
+        const processStreamedDataFunction = async (records: any) => {
+          await saveChangesFromMemory(records, sortedModels, this.syncSettings, progressCallback);
```
edmofro, line 385: Great call moving settings into this.

packages/mobile/App/services/sync/MobileSyncManager.ts
```diff
@@ -436,23 +430,30 @@ export class MobileSyncManager {
...
   }
 
-  /**
-   * Post-pull actions should be run inside the save transaction
-   * @param entityManager - the entity manager for the save transaction
-   * @param pullUntil - the new value for the last successful pull
-   */
-  async postPull(entityManager: any, pullUntil: number) {
+  async postPull(entityManager: EntityManager, pullUntil: number) {
+    if (!entityManager.queryRunner?.isTransactionActive) {
```
edmofro, line 453: Great assertion.

packages/mobile/App/services/sync/utils/buildFromSyncRecord.ts
```diff
@@ -42,3 +42,22 @@ export const buildFromSyncRecord = (
...
   return dbRecord;
 };
+
+export const buildForRawInsertFromSyncRecord = (
+  model: typeof BaseModel,
+  { data, isDeleted }: SyncRecord,
+): DataToPersist => {
+  const includedColumns = extractIncludedColumns(model);
+  const record = pick(data, includedColumns);
+  record.isDeleted = isDeleted;
```
edmofro, line 52: I feel a bit funny about mixing a column that isn't going into the actual record inside the record object.

edmofro: Oh yeah, I checked the call site and see why you did it. But the weirdness of adding that isDeleted to the record, although it's fine to maintain it, should be restricted to that place, I reckon, rather than in this more generic feeling util. It does get used there, but in a way where it's very specific to how those bulk inserts are handled.

edmofro (changes requested):
This is an absolutely incredible speed up! Nice work. Amazing.

packages/mobile/App/services/sync/utils/preparedInsert.ts
dannash100: Yeah i thought of this actually! That it was a weird name in that sense

package.json
dannash100: Yeah its because typeorm is a rare package that lives inside mobile package itself - So the patch has to be inside that package. Other mobile patch is at top level

edmofro: Okd gotcha. I think it would be more clear if the mobile package had it under `npm run patch-package --workspace @tamanu/mobile` and then postinstall called `patch-package` within that same `package.json`. Does that make sense?

packages/central-server/app/facility.js
```diff
@@ -14,8 +14,11 @@ facilityRoutes.get(
     const userEntity = await User.findByPk(user.id);
 
     const allowed = await userEntity.allowedFacilityIds();
+    console.log('allowed', allowed);
     const facilities = await Facility.findAll({
-      where: allowed === CAN_ACCESS_ALL_FACILITIES ? {} : { id: { [Op.in]: allowed } },
+      // where: allowed === CAN_ACCESS_ALL_FACILITIES ? {} : { id: { [Op.in]: allowed } },
```
dannash100, line 19: Note to remove after klaus testing

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "GPL-3.0-and-later AND BUSL-1.1",
   "type": "module",
   "scripts": {
-    "prepare": "patch-package && npm run postinstall --workspace @tamanu/mobile",
+    "prepare": "patch-package && npm run patch-package --workspaces --if-present",
```
edmofro, line 11: Super cool solve!!

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -410,7 +410,7 @@ export class CentralSyncManager {
       // work out if any patients were newly marked for sync since this device last connected, and
       // include changes from all time for those patients
       const newPatientFacilitiesCount = await models.PatientFacility.count({
-        where: { facilityId: { [Op.in]: facilityIds }, updatedAtSyncTick: { [Op.gt]: since } },
+        where: { facilityId: { [Op.in]: facilityIds }, createdAtSyncTick: { [Op.gt]: since } },
```
edmofro, line 413: This is in an unexpected PR..?

packages/mobile/App/services/sync/utils/buildFromSyncRecord.ts
```diff
@@ -29,28 +29,25 @@ const mapFields = (mapping: [string, string][], obj: { [key: string]: unknown })
...
+): DataToPersist[] => {
   const includedColumns = extractIncludedColumns(model);
-  const record = pick(data, includedColumns);
-  record.isDeleted = isDeleted;
-  return record;
+  // populate `fieldMapping` with `RelationId` to `Relation` mappings
+  // (not necessary for `IdRelation`)
+  const fieldMapping = getRelationIdsFieldMapping(model);
+  return records.map(record =>
+    mapFields(fieldMapping, pickBySelectedColumns(record, includedColumns)),
```
edmofro, line 49: I know I pushed in other places for pulling stuff out where maps get too dense, but in this case I think it's _more_ reading effort to understand `pickBySelectedColumns` than just reading the `pick` directly
```suggestion
    mapFields(fieldMapping, pick(record.data, includedColumns)),
```

packages/mobile/App/services/sync/utils/buildFromSyncRecord.ts
```diff
@@ -59,5 +56,9 @@ export const buildForRawInsertFromSyncRecords = (
 ): DataToPersist[] => {
   const includedColumns = extractIncludedColumns(model);
   // Skip field mapping for raw insert - keep original field names
-  return records.map(({ data }) => pick(data, includedColumns));
+  return records.map(record => {
+    const data = pickBySelectedColumns(record, includedColumns);
```
edmofro, line 60: As above

edmofro (changes requested):
Only one tiny nit!

packages/mobile/App/services/sync/utils/preparedInsert.ts
edmofro: Was just wanting to make sure that every row has all column keys present, even if the data is undefined/null - I would assume that's how the lookup table works but I'm not 100% sure

edmofro (approved):
lgtm!

PR #8139, opened 2025-08-04
fix(location): TAMOC-320: Fix location bookings glitching (HOTFIX 2.36)

chris-bes:
### Changes
- Fix location bookings glitching

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/Appointments/LocationBookingsTable.jsx
+++ b/packages/web/app/components/Appointments/LocationBookingsTable.jsx
@@ -299,12 +299,15 @@ export const LocationBookingsTable = ({ patient }) => {
   });
 
   // Query to check if there are past location bookings
-  const pastBookingsQuery = useLocationBookingsQuery({
-    patientId: patient?.id,
-    before: getCurrentDateTimeString(),
-    after: '1970-01-01 00:00',
-    rowsPerPage: 1,
-  });
+  const pastBookingsQuery = useLocationBookingsQuery(
+    {
+      patientId: patient?.id,
+      before: getCurrentDateTimeString(),
+      after: '1970-01-01 00:00',
+      rowsPerPage: 1,
+    },
+    { keepPreviousData: true },
+  );
 
   const hasPastBookings = (pastBookingsQuery.data?.data?.length || 0) > 0;
 
--- a/packages/web/app/components/Appointments/OutpatientAppointmentsTable.jsx
+++ b/packages/web/app/components/Appointments/OutpatientAppointmentsTable.jsx
@@ -317,12 +317,15 @@ export const OutpatientAppointmentsTable = ({ patient }) => {
   });
 
   // Query to check if there are past appointments
-  const pastAppointmentsQuery = useOutpatientAppointmentsQuery({
-    patientId: patient?.id,
-    before: getCurrentDateTimeString(),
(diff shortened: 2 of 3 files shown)


packages/web/app/components/Appointments/LocationBookingsTable.jsx
```diff
@@ -299,12 +299,15 @@ export const LocationBookingsTable = ({ patient }) => {
...
-    rowsPerPage: 1,
-  });
+  const pastBookingsQuery = useLocationBookingsQuery(
+    {
+      patientId: patient?.id,
+      before: getCurrentDateTimeString(),
+      after: '1970-01-01 00:00',
+      rowsPerPage: 1,
+    },
+    { keepPreviousData: true },
```
chris-bes, line 309: The query uses all the fetch options as query key. Given before is different every time, it always refetch and creates the glitch. I've added `keepPreviousData` to keep displaying the previous query's data while a new query is loading, in order to fix the glitch.

packages/web/app/components/Appointments/LocationBookingsTable.jsx
```diff
@@ -299,12 +299,15 @@ export const LocationBookingsTable = ({ patient }) => {
...
+  );
```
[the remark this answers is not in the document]
chris-bes: We do want to refetch every time so that the past bookings are accurate

packages/web/app/components/Appointments/OutpatientAppointmentsTable.jsx
```diff
@@ -317,12 +317,15 @@ export const OutpatientAppointmentsTable = ({ patient }) => {
...
-  });
+  const pastAppointmentsQuery = useOutpatientAppointmentsQuery(
+    {
+      patientId: patient?.id,
+      before: getCurrentDateTimeString(),
+      after: '1970-01-01 00:00',
+      rowsPerPage: 1,
+    },
+    { keepPreviousData: true },
+  );
```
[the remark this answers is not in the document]
chris-bes: We do want to refetch every time so that the past bookings are accurate

packages/web/app/components/Appointments/LocationBookingsTable.jsx
edmofro: In the other places it's used, are we calculating the `before` live, so the same issue could occur?

edmofro (approved):
Changes look good, one question but the core is approved

edmofro:
bugbot run

packages/web/app/components/Appointments/LocationBookingsTable.jsx
chris-bes: Yep I did have a look, there are live calculated `before` and `after`, but they do not get updated as frequently as here. An example is:
```
before: values.startTime ? toDateTimeString(endOfDay(parseISO(values.startTime))) : null
```

which gets the `values.startTime` from a date picker, so I thought it's acceptable to leave it as it is in order to avoid regression risk

edmofro: Yes that's great - I didn't word that very well but by live I was meaning `now()` type of thing

PR #8145, opened 2025-08-04
feat(desktopCharting): SAV-773: Add complex chart graph importer validation

NavarroEmilioLuis:
### Changes

Currently we should not support graphs for complex charts. With this importer validation we will prevent they are added.


tcodling:
bugbot run

edmofro (commented):
There is a new file `packages/central-server/__tests__/importers/programs-charting-simple-valid.xlsx`, but I don't see where it's used?

NavarroEmilioLuis:
> There is a new file `packages/central-server/__tests__/importers/programs-charting-simple-valid.xlsx`, but I don't see where it's used?

No that's not a new file, that's an updated file - unfortunately xlsx doesn't show the diff but I just added information on two cells: `validationCriteria` and `visualisationConfig`. I kinda leave breadcrumbs on commits for intent on that. You can grep `charting-simple-valid`, which is the way the file is called from the tests (as the prefix `programs-` is appended on a common function)

edmofro (approved):
👌

PR #8152, opened 2025-08-05
fix(fhir): TAN-2514: Fix Lab Requests not being immediately sent to SENAITE

rohan-bes:
### Changes

The issue previously is that we were resolving resources with no respect to the order that they depend on one another. However, a resource can only be resolved if all of its dependencies have already been resolved. This meant that the resolver may need to run multiple times to resolve a resource.

Now we sort the resources in dependency order before we begin resolving, much like what we do with sync persisting. 

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/tasks/FhirMissingResources.test.js
+++ b/packages/central-server/__tests__/tasks/FhirMissingResources.test.js
@@ -16,13 +16,16 @@ describe('FhirMissingResources task', () => {
 
   beforeAll(async () => {
     ctx = await createTestContext();
-    const { FhirEncounter, FhirPractitioner } = ctx.store.models;
+    const { FhirEncounter, FhirPractitioner, MediciReport, FhirOrganization } = ctx.store.models;
 
     fhirMissingResourcesWorker = new FhirMissingResources(ctx);
     resources = await fakeResourcesOfFhirServiceRequest(ctx.store.models);
 
     await FhirEncounter.materialiseFromUpstream(resources.encounter.id);
+    await MediciReport.materialiseFromUpstream(resources.encounter.id);
     await FhirPractitioner.materialiseFromUpstream(resources.practitioner.id);
+    await FhirPractitioner.materialiseFromUpstream(SYSTEM_USER_UUID);
+    await FhirOrganization.materialiseFromUpstream(resources.facility.id);
   });
 
   beforeEach(async () => {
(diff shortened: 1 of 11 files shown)


packages/database/src/models/fhir/FhirSpecimen.ts
```diff
@@ -28,6 +28,10 @@ export class FhirSpecimen extends FhirResource {
 
     this.UpstreamModels = [models.LabRequest];
     this.upstreams = [models.LabRequest];
+    this.referencedResources = [
+      models.FhirPractitioner,
+      // FhirServiceRequest is also referenced by Specimen, but not flagging here to avoid a circular dependency
```
rohan-bes, line 33: This feels really quite awkward (similarly in FhirPatient). But it's a bit hard to avoid. FHIR contains circular dependencies by design, so we need to have some way of resolving despite this.

I considered correctly recording all the referenced resources here and then adding a separate field/function to filter those resources to avoid circular dependencies. But given the only use case for this field at the moment is sorting by dependency order I decided it's fine to just omit them here and leave a comment to avoid any confusion. 

If something else pops up in the future that requires the full list of referenced resources then we may need to change this.

packages/central-server/app/tasks/fhir/resolver.js
```diff
@@ -1,6 +1,43 @@
...
+    if (isEqual(previousPass, stillToSort)) {
+      // If nothing changes in a pass, we're stuck in a loop
+      throw new Error(`Circular dependency detected: ${Object.keys(stillToSort).join(', ')}`);
+    }
+
+    previousPass = { ...stillToSort };
+  }
+
+  return sorted;
+};
```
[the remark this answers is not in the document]
rohan-bes: Oh yeah I like this

packages/central-server/app/tasks/fhir/resolver.js
```diff
@@ -2,6 +2,41 @@ import { FHIR_INTERACTIONS } from '@tamanu/constants';
...
+    acc[resource.name] = resource.referencedResources.map(r => r.name);
+    return acc;
+  }, {});
+
+export const sortResourcesInDependencyOrder = resources => {
+  const dependencyMap = buildDependencyMap(resources);
+
+  const sorted = [];
+  const stillToSort = new Map(resources.map(r => [r.name, r]));
+  let lastSize = -1;
```
NavarroEmilioLuis, line 16: Not that it matters, but I would set this to `Infinity` or `null`. It's strikes me as weird that we're not setting it to `stillToSort.size`, but then we wouldn't be able to capture the circular dependency.

NavarroEmilioLuis (approved):
I was trying to come up of a way we could derive the referenced resources but I guess that's it - mostly because I think it's easy to overlook, but yeah, no clue.

This work looks good to me, I left one question/suggestion for the test (which might not need change) and the smallest of nitpick

packages/central-server/__tests__/tasks/FhirMissingResources.test.js
```diff
@@ -16,13 +16,16 @@ describe('FhirMissingResources task', () => {
...
+    const { FhirEncounter, FhirPractitioner, MediciReport, FhirOrganization } = ctx.store.models;
 
     fhirMissingResourcesWorker = new FhirMissingResources(ctx);
     resources = await fakeResourcesOfFhirServiceRequest(ctx.store.models);
 
     await FhirEncounter.materialiseFromUpstream(resources.encounter.id);
+    await MediciReport.materialiseFromUpstream(resources.encounter.id);
     await FhirPractitioner.materialiseFromUpstream(resources.practitioner.id);
+    await FhirPractitioner.materialiseFromUpstream(SYSTEM_USER_UUID);
+    await FhirOrganization.materialiseFromUpstream(resources.facility.id);
```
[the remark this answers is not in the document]
rohan-bes: I chose to materialise these here since they kind of just create unexpected results in the tests. These are all part of the test data that gets generated in order to setup. Since this test is about counting the unmaterialised resources, these kept showing up in the numbers. Now I think the tests read more clearly (eg. `should not create one FHIR fromUpstream job if the missing upstream resource do not meet pre-filter criteria` now creates 0 jobs, rather than the 2 it created before due to unmaterialised setup data)

packages/central-server/app/tasks/fhir/resolver.js
rohan-bes: yeah loops like this where you need to init to a value that will never be true are always a little funny

PR #8158, opened 2025-08-06
db(model): record table triggers

passcod:
### Changes

This can be very useful to generate further outputs without needing to have a live database.

The `config.meta` key in the dbt schema can contain arbitrary key-value data without breaking dbt, so we might later gather more information statically in there (indexes? constant/enum information? sync direction?).


database/model/public/patient_program_registration_conditions.yml
```diff
@@ -9,6 +9,11 @@ sources:
         config:
           tags:
             - clinical
+          meta:
+            triggers:
+              - notify_patient_program_registration_conditions_changed
+              - record_patient_program_registration_conditions_changelog
+              - set_patient_program_registration_conditions_updated_at_sync_tic
```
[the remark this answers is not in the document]
rohan-bes: No I think this is correct, postgres has a 63 char max for trigger name :\

edmofro:
bugbot run

packages/scripts/src/dbt/generateModel.js
```diff
@@ -36,24 +36,26 @@ async function getSchemas(client) {
...
- * @returns A list of table names that the given schema name has
+ * @returns {Array<{ name: string; oid: number }>} A list of table names that the given schema name has
  */
 async function getTablesInSchema(client, schemaName) {
   return (
     await client.query(
-      `SELECT DISTINCT table_name
+      `SELECT DISTINCT
+        table_name as name,
+         ('"' || table_schema || '"."' || table_name || '"')::regclass::oid as oid
```
rohan-bes, line 52: Oh this is so cool I had no idea you could get from the human readable table name to the `oid` this way!

packages/scripts/src/dbt/generateModel.js
```diff
@@ -79,6 +81,24 @@ async function getColumnsInRelation(client, schemaName, tableName) {
...
+async function getTriggers(client, oid) {
+  return (
+    await client.query(
+      `SELECT oid, tgname as name
+      FROM pg_trigger
+      WHERE tgisinternal = false AND tgrelid = $1`,
+      [oid],
+    )
+  ).rows
+    .filter(({ name }) => !name.startsWith('RI_'))
```
rohan-bes, line 98: What's the reason for filtering out the `RI_*` triggers? A comment might be nice :+1:

packages/scripts/src/dbt/generateModel.js
```diff
@@ -79,6 +81,24 @@ async function getColumnsInRelation(client, schemaName, tableName) {
   }));
 }
 
+/**
+ * @param {pg.Client} client
+ * @param {number} oid
+ * @returns {Array<{ oid: number; name: string; }>} A list of triggers that exist on the given table.
```
rohan-bes, line 87: Very much a nitpick, but do we need to be returning the `oid` here? I found the code a little tricky to follow but it looks like we never use it, and always end up calling `triggers.map(t => t.name)` at some point. Wondering if we could just return `Array<string>` of trigger names here?

rohan-bes: Do understand if we just want to keep it consistent with the other functions though :+1:

packages/scripts/src/dbt/generateModel.js
```diff
@@ -279,7 +305,7 @@ function generateColumnDoc(column) {
 
 /**
  * @param {string} schemaName
- * @param {object} table
+ * @param {{ name: string; oid: string; columns: object[]; triggers: { oid: number; name: string; }[] }} table
```
rohan-bes, line 308:
```suggestion
 * @param {{ name: string; oid: number; columns: object[]; triggers: { oid: number; name: string; }[] }} table
```
?

rohan-bes (approved):
Nice stuff! A few comments around the generateModel logic but nothing major. Approved :+1:

packages/scripts/src/dbt/generateModel.js
passcod: I didn't either but I knew you could do `'foo.bar'::regclass::oid` and thought... hmm, i wonder if it works dynamically... and it does! i love this stuff

packages/scripts/src/dbt/generateModel.js
passcod: They're postgres internals. idk why they're called RI_

packages/scripts/src/dbt/generateModel.js
passcod: thanks

PR #8216, opened 2025-08-13
fix(translations): TAMOC-318: Added support for non-string default option translation in programs (hotfix v2.36)

rohan-bes:
### Changes

Previously we were assuming all options were string values, and were making an unsafe call to `.replace()`.
Changed to:
1. Throw an error if passed a null/undefined value (this should never happen)
2. Cast the value to a string

Strictly this could create clashes again (for instance if one value was `"1"` and the other was `1`) but that's just a horrible idea in general so might be nice for us to prevent it.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/central-server/__tests__/importers/programs-vitals-valid.xlsx
rohan-bes, line 1: Added a question with default options:
```
{
  "Zero": 0,
  "One": 1
}
```

tcodling: Hahaha by that `todo` comment above on that link I think this is just a weird bit of logic that confuses everyone that comes across it. I had an extremely hard time myself wrapping my head around the survey program logic while working on it. 

I think it worth having a chat with @MeganLane29 as I suspect the same. It would be cool if we could maybe check the surveys in use to see if any of these confusing columns are unused and can be removed? I feel there is potential for a big cleanup/simplification here 🤔

tcodling (changes requested):
I think its suggestion about handing `undefined` looks valid so I would make that tweak. Otherwise looks great!

Also totally agree with the confusion regarding surveys

tcodling (approved):
lgtm!

PR #8217, opened 2025-08-13
db(model): masking definitions

passcod:
### Changes

This adds a simple masking definition to relevant columns in most tables in the DBT schema.

The idea here is that we don't yet have a solution/tool for masking data, and we don't want to lock ourselves into one, but with this simplified definition we can now easily generate a masking config for any given tool and experiment with that without requiring changes in the Tamanu source.

This PR defines 17 types of masking, and applies them to 251 individual columns and 13 entire tables. It loosely documents the expected behaviour for each masking type, and otherwise does no other processing.

In the dbt model generator, it makes three additions:
- it checks existing schema files to ensure no unknown (or typo'd) masking types are used
- it adds `truncate` masking for new fhir tables by default
- it adds `text` masking for new `text` columns by default (to hopefully encourage devs to set the correct masking for new tables and columns)

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


tcodling:
bugbot run

rohan-bes (approved):
Looks good :+1: Approved!

PR #8247, opened 2025-08-14
fix(sensitiveFacilities): NASS-1751: Keep historical sensitive data unsynced

tcodling:
### Changes

We need to handle this edge case a bit differently for sensitive facilities

If an encounter was created when a facility is **sensitive**, that encounter and its related data should never sync down to other facilities even if the parent facility is changed to **non-sensitive**. Currently an edit to any data will trigger that record to sync down which both creates foreign key issues and also exposes sensitive data.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/sync/CentralSyncManager.sensitiveFacilities.test.js
+++ b/packages/central-server/__tests__/sync/CentralSyncManager.sensitiveFacilities.test.js
@@ -1018,7 +1018,7 @@ describe('CentralSyncManager Sensitive Facilities', () => {
       expect(encounterIds).not.toContain(sensitiveEncounterB.id);
     });
 
-    it('will keep historical sensitive data unsynced to other facilities when a facility changes from sensitive to non-sensitive, until the data is edited', async () => {
+    it('will keep historical sensitive data unsynced to other facilities when a facility changes from sensitive to non-sensitive, even after the data is edited', async () => {
       // Create a facility that starts as sensitive
       const facility = await models.Facility.create(fake(models.Facility, { isSensitive: true }));
       const department = await models.Department.create(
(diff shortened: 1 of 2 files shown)


packages/central-server/app/sync/updateLookupTable.js
```diff
@@ -78,7 +78,10 @@ const updateLookupTableForModel = async (model, config, since, sessionConfig, sy
             is_lab_request = EXCLUDED.is_lab_request,
             patient_id = EXCLUDED.patient_id,
             encounter_id = EXCLUDED.encounter_id,
-            facility_id = EXCLUDED.facility_id,
+            facility_id = CASE 
+              WHEN EXCLUDED.facility_id IS NOT NULL THEN EXCLUDED.facility_id
+              ELSE sync_lookup.facility_id
```
tcodling, line 83: The thought behind this is that if a record has a "facility_id" populated at any point it will never clear and and stay restricted to that facility. 

This was a cursor suggestion and still investigating this to make sure it is acting how I understand and wont have any side effects 🤔

tcodling: I seems to be working fine but finding it tricky to justify as I don't think I fully understand what situation triggers this ON CONFLICT

packages/central-server/__tests__/sync/CentralSyncManager.sensitiveFacilities.test.js
```diff
@@ -1058,13 +1058,13 @@ describe('CentralSyncManager Sensitive Facilities', () => {
...
 
-      // Check that the new encounter changes are synced to the non-sensitive facility
+      // Check that the new encounter changes are still not synced to the non-sensitive facility
       const updatedEncounterIds = await getOutgoingIdsForRecordType(
         centralSyncManager,
         nonSensitiveFacility.id,
         'encounters',
       );
-      expect(updatedEncounterIds).toContain(encounter.id);
+      expect(updatedEncounterIds).not.toContain(encounter.id);
```
tcodling, line 1067: We actually want the opposite behaviour for this last check

PR #8250, opened 2025-08-14
ci(audit): COOL-7: Test environment

NavarroEmilioLuis:
### Changes

So this will be the testing branch for this card as we kinda need to include a migration to test it

### Deploys


NavarroEmilioLuis:
Ok, environment is finally down

PR #8251, opened 2025-08-15
feat(mobileSync): KOR-53: X-Large syncs and standardise prepared queries for sync persist

dannash100:
### Changes
Last word on the mobile sync stuff, just discovered that updates and deletes and restores are still very slow, and 11 million pull had some issues.
A bit of alot of things for one pr, but I need to handover 🙏 

**Misc**
- Fix deletion of tables for full resync
- Added a check for maximum parameters to optimise insert batches. It will error on patient_additional_data at 2k otherwise.
- Added a max updates per batch setting too to match insert (I am finding the settings very useful to tweak and explore further optimisations, plus could come in handy if we have to alter something in the field)

**To address x-large pulls**
 - Index change + vacuum seemed to unstick queries to snapshot table - otherwise it just grabs 1000 at time and is very slow (not entirely 100% on this one but it worked to unstuck it)
- Added setting for unsafe schema as it can out of memory in late state of 11 million sync I tested (open to removing outright?)

**To address slow delete, restore, update crud during sync**
- Now there is just two prepared queries that handle everything that was covered in `executeCrud` update and insert prepared queries
- Incorporate restore and delete logic into prepared inserts because we can set deleted_at in raw query, this saves alot of iterations and data prep.
- I don't think anything is missed this way 🤔 

45-50 ish mins for 11 min initial sync!
testing again with other changes.
Confirmed changes work with 11 mil!

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


tcodling:
bugbot run

packages/mobile/App/services/sync/utils/executePreparedQuery.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export const executePreparedInsert = async (
+  repository: Repository<any>,
+  rows: DataToPersist[],
+  maxRecordsPerBatch: number,
+  progressCallback: (processedCount: number) => void,
+) => {
+  if (!rows.length) return;
+
+  // Can end up with duplicate create records, e.g. if syncAllLabRequests is turned on, an
```
dannash100, line 39: Is this still valid with the lookup table?

packages/mobile/App/services/sync/utils/executeCrud.ts
```diff
@@ -1,131 +0,0 @@
...
-  const idsAdded = new Set();
-  const softDeleted = rows.filter(row => row.isDeleted).map(strippedIsDeleted);
-
-  for (const row of rows) {
-    const { id } = row;
-    if (!idsAdded.has(id)) {
-      deduplicated.push({ ...strippedIsDeleted(row), id });
-      idsAdded.add(id);
-    }
-  }
```
rohan-bes, line 33: @edmofro As part of this change we're thinking of removing this de-duplication logic, as it shouldn't be needed with sync lookup anymore.

Just wanna confirm that we've fully deprecated the old sync system? And we're happy that all deployments going forward are using sync_lookup

rohan-bes (approved):

PR #8252, opened 2025-08-15
fix(sync): TAMOC-331: Fixup lookup table state for prescriptions and encounter_prescriptions (hotfix 2.34)

rohan-bes:
### Changes

Introduced a new local_system_fact called `lookupModelsToRebuild` as well as a utility function to add table names to it. That way in migrations that modify a table schema we can just add this function call at the end and let the lookup table rebuild the records.

Does handle the issue of removing deleted/renamed tables tho, so that still needs to be done manually.

Also added a migration that calls this function on `prescriptions` and `encounter_prescriptions` as they currently aren't populating in the lookup table by default.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/app/sync/updateLookupTable.js
+++ b/packages/central-server/app/sync/updateLookupTable.js
@@ -140,7 +140,7 @@ export const updateLookupTable = withConfig(
         const modelChangesCount = await updateLookupTableForModel(
           model,
           config,
-          shouldFullyRebuild ? 0 : since,
+          shouldFullyRebuild ? -1 : since,
           sessionConfig,
           syncLookupTick,
         );
--- a/packages/central-server/app/sync/updateLookupTable.js
+++ b/packages/central-server/app/sync/updateLookupTable.js
@@ -3,7 +3,14 @@ import { log } from '@tamanu/shared/services/logging/log';
 import { withConfig } from '@tamanu/shared/utils/withConfig';
 import { buildSyncLookupSelect, SYNC_TICK_FLAGS } from '@tamanu/database/sync';
 
-const updateLookupTableForModel = async (model, config, since, sessionConfig, syncLookupTick) => {
+const updateLookupTableForModel = async (
+  model,
+  config,
+  since,
+  sessionConfig,
+  syncLookupTick,
+  shouldFullyRebuild,
+) => {
   const CHUNK_SIZE = config.sync.maxRecordsPerSnapshotChunk;
   const { perModelUpdateTimeoutMs, avoidRepull } = config.sync.lookupTable;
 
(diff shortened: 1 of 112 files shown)


packages/central-server/app/sync/updateLookupTable.js
```diff
@@ -132,17 +132,23 @@ export const updateLookupTable = withConfig(
...
+    const modelsToRebuild = await models.LocalSystemFact.getLookupModelsToRebuild();
 
     for (const model of Object.values(outgoingModels)) {
       try {
+        const shouldFullyRebuild = modelsToRebuild.includes(model.tableName);
         const modelChangesCount = await updateLookupTableForModel(
           model,
           config,
-          since,
+          shouldFullyRebuild ? 0 : since,
```
rohan-bes, line 143: I'm pretty sure 0 is the right value here? Feels a little magic-numbery tho so might define a constant for it.

packages/database/src/models/LocalSystemFact.ts
```diff
@@ -88,4 +88,32 @@ export class LocalSystemFact extends Model {
...
+  static async markLookupModelRebuilt(modelName: string) {
+    await this.sequelize.query(
+      `
+      UPDATE local_system_facts
+      SET value =
+      CASE
+        WHEN value IS NULL THEN NULL
+      ELSE array_to_string(array_remove(string_to_array(value, ','), :modelName), ',')
+      END
+      WHERE key = :key
```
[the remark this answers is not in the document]
rohan-bes: Yeah good call :+1:

passcod (approved):
optional minor changes but otherwise looks good

packages/database/src/sync/buildSyncLookupSelect.ts
```diff
@@ -9,18 +9,32 @@ interface Columns {
...
+/**
+ * When fully rebuilding we want to use the underlying
+ */
+const updatedAtSyncTickClause = (table: string, isFullyRebuilding: boolean) => {
+  const historicRecordSyncTick = `${table}.updated_at_sync_tick`;
+  const newRecordSyncTick = `COALESCE(:updatedAtSyncTick, ${table}.updated_at_sync_tick)`;
+
+  return isFullyRebuilding
+    ? `CASE WHEN ${table}.updated_at_sync_tick <= :since THEN ${historicRecordSyncTick} ELSE ${newRecordSyncTick} END`
+    : newRecordSyncTick;
```
chris-bes, line 21: I think it might be worth putting a comment explaining this one

chris-bes (approved):
LGTM!

packages/database/src/sync/buildSyncLookupSelect.ts
rohan-bes: Yeah that's a really good call, will do :+1:

PR #8254, opened 2025-08-17
fix(surveys): TAMOC-332: Display correct decimal places for calculated question values when completing a survey

rohan-bes:
### Changes

Switched to using the same logic on the frontend when displaying calculated question values that we use in the backend when saving the value. Useful as it both displays `0` correctly now and also rounds to the correct number of decimal places.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/shared/src/utils/calculations.js
+++ b/packages/shared/src/utils/calculations.js
@@ -1,3 +1,4 @@
+import { isNil } from 'lodash';
 import { all as allMath, create } from 'mathjs';
 
 // set up math context
@@ -56,7 +57,7 @@ export function runCalculations(components, values) {
           calculatedValues[c.dataElement.id] = null;
           continue;
         }
-        
+
         let value = math.evaluate(c.calculation, inputValues);
 
         if (Number.isNaN(value)) {
@@ -76,3 +77,15 @@ export function runCalculations(components, values) {
 
   return calculatedValues;
 }
+
+export const getStringOfCalculatedValue = value => {
+  if (isNil(value)) {
+    return '';
+  }
+
+  if (!(typeof value === 'number')) {
+    throw new Error(`Unexpected calculated value is not a number: ${value}`);
+  }
+
+  return value.toFixed(1);
+};
--- a/packages/shared/src/utils/calculations.js
+++ b/packages/shared/src/utils/calculations.js
@@ -84,7 +84,9 @@ export const getStringOfCalculatedValue = value => {
   }
 
   if (!(typeof value === 'number')) {
-    throw new Error(`Unexpected calculated value is not a number: ${value}`);
+    throw new Error(
+      `The result of a calculation must be a number, but got: ${value} (${typeof value})`,
+    );
   }
 
   return value.toFixed(1);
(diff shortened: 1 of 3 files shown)


packages/shared/src/utils/calculations.js
```diff
@@ -76,3 +77,15 @@ export function runCalculations(components, values) {
...
+export const getStringOfCalculatedValue = value => {
+  if (isNil(value)) {
+    return '';
+  }
+
+  if (!(typeof value === 'number')) {
+    throw new Error(`Unexpected calculated value is not a number: ${value}`);
+  }
+
+  return value.toFixed(1);
```
rohan-bes, line 90: Got this from the original implementation of `getStringValue`

packages/web/app/utils/survey.jsx
```diff
@@ -46,15 +52,21 @@ const QUESTION_COMPONENTS = {
...
   [PROGRAM_DATA_ELEMENT_TYPES.NUMBER]: NumberField,
   [PROGRAM_DATA_ELEMENT_TYPES.BINARY]: NullableBooleanField,
   [PROGRAM_DATA_ELEMENT_TYPES.CHECKBOX]: NullableBooleanField,
-  [PROGRAM_DATA_ELEMENT_TYPES.CALCULATED]: ReadOnlyTextField,
+  [PROGRAM_DATA_ELEMENT_TYPES.CALCULATED]: ({ field, ...props }) => (
+    <ReadOnlyTextField
+      {...props}
+      field={{
+        ...field,
+        value: getStringOfCalculatedValue(field?.value),
```
rohan-bes, line 66: A change in behaviour with this is that now if the value isn't a number, we'll hit an uncaught error here. Since that's a mis-configuration and the user would be hitting an error on submission previously I think that's okay. But arguably a toast message would be cleaner...

packages/shared/src/utils/calculations.js
```diff
@@ -76,3 +77,15 @@ export function runCalculations(components, values) {
...
 }
+
+export const getStringOfCalculatedValue = value => {
+  if (isNil(value)) {
+    return '';
+  }
+
+  if (!(typeof value === 'number')) {
+    throw new Error(`Unexpected calculated value is not a number: ${value}`);
+  }
```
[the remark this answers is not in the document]
tcodling: Honestly I am finding this message quite confusing? Especially if it is being displayed to the end user 🤔 It could just be that I am a bit unfamiliar with how the calculated value works though so wont fail the review for this just a thought

tcodling (approved):
Logic all looks solid to me. Only thing is I personally find the variable names + error message a little confusing for some reason.

I think its just my personal preference/lack of knowledge around how calculations work though. Definitely not going to fail for this.

packages/shared/src/utils/calculations.js
```diff
@@ -76,3 +77,15 @@ export function runCalculations(components, values) {
 
   return calculatedValues;
 }
+
+export const getStringOfCalculatedValue = value => {
```
tcodling, line 81: I feel like this could have a better name but I cant think of one 😆 

`formatCalculatedValue` maybe? Honestly wouldn't stress unless you have a good idea

packages/shared/src/utils/calculations.js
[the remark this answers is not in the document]
rohan-bes: Yeah that's a good point, the wording definitely isn't the best. Really we're saying 'A calculated question result must be a number, but instead we got: <value>`, I'll change it to be clearer.

tcodling (approved):

PR #8256, opened 2025-08-17
fix(sensitiveFacilities): Release fix 2.38: Fix medication administration record sync

tcodling:
### Changes

We discovered some issues in testing with a new case related to models linked thourgh `prescription`. We have an issue where `presciption` can be linked to a facility thorough 2 possible join tables. To get around this we have to extend the fix done to the `prescription` model to these models that link through `prescription`. Now that there are 3 cases like this its possible we should make some kind of cleaner util to dry up the code but feels out of scope for a last min release fix.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
+++ b/packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
@@ -162,6 +162,8 @@ describe('Sync Lookup data', () => {
       Notification,
       EncounterPausePrescription,
       EncounterPausePrescriptionHistory,
+      MedicationAdministrationRecord,
+      MedicationAdministrationRecordDose,
     } = models;
 
     await Asset.create(fake(Asset), {
@@ -308,6 +310,21 @@ describe('Sync Lookup data', () => {
       fake(PatientOngoingPrescription, {
         patientId: patient.id,
         prescriptionId: prescription.id,
+      }),
+    );
+
+    const mar = await MedicationAdministrationRecord.create(
+      fake(models.MedicationAdministrationRecord, {
+        prescriptionId: prescription.id,
+        recordedByUserId: examiner.id,
+      }),
+    );
+
+    await MedicationAdministrationRecordDose.create(
+      fake(models.MedicationAdministrationRecordDose, {
+        marId: mar.id,
+        givenByUserId: examiner.id,
+        recordedByUserId: examiner.id,
       }),
     );
 
--- a/packages/database/src/models/MedicationAdministrationRecord.ts
+++ b/packages/database/src/models/MedicationAdministrationRecord.ts
@@ -544,7 +544,6 @@ export class MedicationAdministrationRecord extends Model {
       return null;
     }
     return `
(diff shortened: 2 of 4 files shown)


packages/database/src/models/MedicationAdministrationRecord.ts
```diff
@@ -538,11 +539,35 @@ export class MedicationAdministrationRecord extends Model {
...
+    return `
+      LEFT JOIN encounter_prescriptions ON medication_administration_records.prescription_id = encounter_prescriptions.prescription_id
+      LEFT JOIN encounters ON encounter_prescriptions.encounter_id = encounters.id
+      LEFT JOIN patient_ongoing_prescriptions ON medication_administration_records.prescription_id = patient_ongoing_prescriptions.prescription_id
+      WHERE (
+        (encounters.patient_id IS NOT NULL AND encounters.patient_id IN (SELECT patient_id FROM ${markedForSyncPatientsTable}))
+        OR 
+        (patient_ongoing_prescriptions.patient_id IS NOT NULL AND patient_ongoing_prescriptions.patient_id IN (SELECT patient_id FROM ${markedForSyncPatientsTable}))
+      )
+      AND medication_administration_records.updated_at_sync_tick > :since
```
tcodling, line 555: These sql snippets are copied/extended from what was defined on the Prescription model. As mentioned in description there is definitely room for some sort of util to clean up these models that use a join table

packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
```diff
@@ -311,6 +313,21 @@ describe('Sync Lookup data', () => {
...
+    );
+
+    await MedicationAdministrationRecordDose.create(
+      fake(models.MedicationAdministrationRecordDose, {
+        marId: mar.id,
+        givenByUserId: examiner.id,
+        recordedByUserId: examiner.id,
+      }),
+    );
+
```
NavarroEmilioLuis, line 330: We don't need to expect anything different after adding these?

packages/database/src/models/MedicationAdministrationRecord.ts
NavarroEmilioLuis: I think we do have a util to do the joins? But this is very specific and I kinda like more that it's explicit on the file.

NavarroEmilioLuis (approved):
LGTM, just one question around one test

packages/central-server/__tests__/sync/CentralSyncManager.syncLookup.test.js
tcodling: This actually fixes test that was failing after I corrected the logic. Otherwise this test throws an error as its missing these lookup records

tcodling: Previously it was syncing everywhere and wasn't considered a patientLinkedModel (which it is) so it was skipping this test

PR #8264, opened 2025-08-18
fix: TAMOC-330: Add deviceId to central server

tcodling:
### Changes

We now also want to track a device id in the central server for audit reasons. For this I pulled out the `initDeviceId` function into shared and made it generic. 

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


.github/scripts/test-facility-offline.sh
```diff
@@ -33,7 +33,8 @@ test_facility_offline_central_start() {
             "name": "central",
             "verbose": true,
             "username": "tamanu",
-            "password": "tamanu"
+            "password": "tamanu",
+            "migrateOnStartup": true
```
passcod, line 37: I don't think that's needed

passcod (approved):
I actually reviewed this fully, looks good

Also great coincidence because of SAV-456 that also deals with devices.

packages/facility-server/config/default.json5
```diff
@@ -47,6 +47,7 @@
     }
   },
   "port": 4000,
+  "deviceId": null,
```
tcodling, line 50: This was already being accessed but wasn't in the default. Just adding it here and in central for clarity

edmofro (approved):
Codeowner stamp

Harnesh787 (approved):

PR #8267, opened 2025-08-18
feat(alerts): TAMOC-334: Detect a gap in sync_sessions being created on central

tcodling:
### Changes

We discovered during an OTS that all facilities on Fiji prime had gone 3 days without syncing which was completely unnoticed. This seemed like it would have been easily caught by a simple alert so I am just adding that in here.

The issue was related to the database being overloaded and hanging meaning no sync sessions were ever created

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


rohan-bes (changes requested):

alerts/sync-no-sessions.yml
```diff
@@ -0,0 +1,37 @@
...
+      **Facilities with no recent sync sessions:**
+      {% for row in rows %}
+      - Facility ID: {{ row.facility_id }} - Last sync: {{ row.last_successful_sync if row.last_successful_sync else 'Unknown' }}
+      {% endfor %}
+
+      **Immediate action required:**
+      1. Check if sync service is running
+      2. Verify database connectivity
+      3. Check facility network status
+      4. Review system logs for errors
```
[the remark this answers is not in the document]
tcodling: Hmm I feel like this is ok? This would only be a problem for completely new deployments right?

alerts/sync-no-sessions.yml
```diff
@@ -0,0 +1,39 @@
+sql: |
+  SELECT facility_id, last_successful_sync FROM (
+    SELECT parameters->'facilityIds'->>0 as facility_id, max(completed_at) as last_successful_sync
+    FROM sync_sessions
+    WHERE errors IS NULL
+    GROUP BY parameters->'facilityIds'->>0
+  ) AS last_successful_facility_syncs
+  WHERE last_successful_sync < now() - interval '1 hour'
```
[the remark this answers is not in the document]
rohan-bes: Yeah I've only ever seen one value in `facilityIds`, so I got a bit lazy and just went with the `[0]`. 

The correct way to do this is:
```
SELECT facility_id, last_successful_sync FROM (
    SELECT facility_id, max(completed_at) as last_successful_sync
    FROM (
      SELECT jsonb_array_elements_text(parameters->'facilityIds') as facility_id, completed_at
      FROM sync_sessions
      WHERE errors IS NULL
    ) AS successful_syncs
    GROUP BY facility_id
  ) AS last_successful_facility_syncs
  WHERE last_successful_sync < now() - interval '1 second'
```
Should probs do that instead :+1:

tcodling:
@rohan-bes Maybe we should get someone else to pick this up since we both pretty much worked on it

tcodling:
bugbot run

alerts/sync-no-sessions.yml
```diff
@@ -0,0 +1,42 @@
...
+      This could indicate:
+      - Sync service is down or not running
+      - Network connectivity issues
+      - Database connection problems
+      - Specific facilities are offline
+      - Sync sessions are failing with errors
+
+      **Facilities with no recent successful syncs:**
+      {% for row in rows %}
+      - Facility ID: {{ row.facility_id }} - Last successful sync: {{ row.last_successful_sync if row.last_successful_sync else 'Unknown' }}
```
NavarroEmilioLuis, line 34: I don't think `row.last_successful_sync` could be missing though - but I'm happy to leave it. The question is: if it can be empty, could `row.facility_id` be empty too?

NavarroEmilioLuis (approved):
Everything checks out tbh so I'm happy to approve

alerts/sync-no-sessions.yml
tcodling: I can see last_successful_sync being empty if the facility never synced? I would  think facility_id would always exist unless the database was in a really weird broken state

tcodling:
@rohan-bes looks like it needs a stamp from you

alerts/sync-no-sessions.yml
NavarroEmilioLuis: Nah okay yeah the subqueries are fine. Nevermind that part. I still think we won't get any unknown but let's leave it, better to be safe than sorry 😂

rohan-bes (approved):

PR #8273, opened 2025-08-18
fix(appointments): TAMOC-335: Set patient_facility on Appointment creation

tcodling:
### Changes

Currently creating an appointment doesn't mark a patient for sync on the facility of the appointment

If an appointment is created for a patient that has is not marked for sync on a facility ( no PatientFacility record ) and then that patient is merged ->

Then the base patient becomes deleted on facility with a merged_into id - but the appointment does not get the patient_id updated as the patient old and new are not marked for sync on that facility

This bug slipped through as this is one of the only ways to "link" a patient to a facility that is completely seperate from the encounter view which usually enforces that the patient is marked before creating all these data types.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/facility-server/__tests__/apiv1/Appointments.test.js
```diff
@@ -46,6 +46,7 @@ describe('Appointments', () => {
       startTime: add(new Date(), { days: 1 }), // create a date in the future
       clinicianId: userApp.user.dataValues.id,
       appointmentTypeId: 'appointmentType-standard',
+      facilityId,
```
tcodling, line 49: This is in line with the pattern of explicitly posting the facilityId to the endpoint if it requires it

tcodling:
bugbot run

tcaiger (approved):
LGTM

PR #8274, opened 2025-08-19
feat(auth): SAV-456: Device registration

NavarroEmilioLuis:
### Changes

This includes the core work of enforcing only approved devices can log in to central server (and thus, do any action that needs authentication).

### Deploys


passcod (approved):
for Kamaka lol

passcod:
Lmao whoops
image

packages/mobile/package.json
```diff
@@ -38,7 +38,6 @@
     "@sayem314/react-native-keep-awake": "^1.0.4",
     "@storybook/addon-actions": "^6.5.16",
     "@storybook/react-native": "^6.5.7",
-    "@tamanu/shared": "*",
```
NavarroEmilioLuis, line 41: Yeah I was trying to import but got this error on the mobile test. However I could pass the test locally and realized it was because we were not building share in CI? https://github.com/beyondessential/tamanu/pull/8277/commits/b679214859610ee805fd6606c63e8db54f0ddc69 so yeah we can undo this. I was a bit wary but on that branch I was also able to build and use mobile just fine on auto deploy!

Or I wonder if I'm missing anything?

packages/central-server/app/sync/buildSyncRoutes.js
```diff
@@ -53,21 +74,22 @@ export const buildSyncRoutes = ctx => {
         log.info('StaleSyncSessionCleaner.closedReconnectedSession', {
           sessionId: session.id,
           durationMs,
-          facilityIds: session.parameters.facilityIds,
-          deviceId: session.parameters.deviceId,
+          facilityIds,
+          deviceId: device.id,
```
NavarroEmilioLuis, line 78: Hmmmm well this is just a log, but noting that `facilityIds` seems like it could be mismatched?

NavarroEmilioLuis (commented):
I think everything looks fine with the changes, I'll take a look to the failing test

packages/central-server/app/sync/buildSyncRoutes.js
```diff
@@ -53,21 +74,22 @@ export const buildSyncRoutes = ctx => {
...
         });
       }
 
       // now update our position in the queue and check if we're at the front of it
-      const queueRecord = await SyncQueuedDevice.checkSyncRequest({
+      const queueRecord = await SyncQueuedDevice.checkSyncRequest(device.id, {
         lastSyncedTick: 0,
```
NavarroEmilioLuis, line 84: Hmmm right, so the tests seemed fishy because we never actually checked the device ID matched the one we expected. After a closer look, my working theory is because we're always using `0` here, so we never order anything, I'll test this in a sec

NavarroEmilioLuis: Yeah okay that's the issue, just pushed the fix. I was thinking that if we just blindly accept whatever the client passes, we could theoretically hijack the order and do like a "sync prevent" attack by scheduling urgent ones with tick 0... OFC with this work it's only going to be possible from authorized devices so yeah I guess it's ok to leave as is.

TBH I'm not really sure that this is the change we are supposed to make but it makes things "work" - I'll now create a separate PR to prove that the tests were flawed before this work.

NavarroEmilioLuis: Oh okay nevermind. So previously the test was working, the issue is that we were passing `...req.body` which would overwrite the `lastSyncedTick` and `urgent` variables. They were just the default values, so I'm adding them back 👍

chris-bes (approved):
Reviewed the sync part and looks good

PR #8308, opened 2025-08-21
feat(audit): SAV-1007: Populate devices table

NavarroEmilioLuis:
### Changes

The requirement is to populate this table with all known devices so far. I'm not sure how we will gather the `name` column, or if it might be available at this point but left a placeholder to include, if it doesn't happen then it's just a explicit insert which is nice anyway.

### Deploys


packages/database/src/migrations/1755803615873-populateDevicesTable.ts
```diff
@@ -0,0 +1,15 @@
+import { QueryInterface } from 'sequelize';
+import { SYSTEM_USER_UUID } from '@tamanu/constants';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    INSERT INTO devices (id, last_seen_at, name, registered_by_id)
+    SELECT device_id, now(), NULL, '${SYSTEM_USER_UUID}' FROM sync_device_ticks GROUP BY device_id;
+  `);
```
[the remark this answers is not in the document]
NavarroEmilioLuis: This would be vulnerable if an attacker managed to update the constants package at repo level (which would be pretty weird) or at file level (at this point, they had to gain access to the deployment) which means we're already pwned anyway as they could simply change the migration file directly if for some reason they couldn't do it directly to the database.

packages/database/src/migrations/1755803615873-populateDevicesTable.ts
```diff
@@ -0,0 +1,15 @@
+import { QueryInterface } from 'sequelize';
+import { SYSTEM_USER_UUID } from '@tamanu/constants';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    INSERT INTO devices (id, last_seen_at, name, registered_by_id)
+    SELECT device_id, now(), NULL, '${SYSTEM_USER_UUID}' FROM sync_device_ticks GROUP BY device_id;
```
passcod, line 7:
```suggestion
    INSERT INTO devices (id, last_seen_at, scopes, registered_by_id)
    SELECT device_id, now(), '["sync-client"]'::jsonb, uuid_nil() FROM sync_device_ticks GROUP BY device_id;
```

packages/database/src/migrations/1756167650873-populateDevicesTable.ts
```diff
@@ -0,0 +1,15 @@
+import { QueryInterface } from 'sequelize';
+import { SYSTEM_USER_UUID, DEVICE_SCOPES } from '@tamanu/constants';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    INSERT INTO devices (id, last_seen_at, scopes, registered_by_id)
+    SELECT device_id, now(), '["${DEVICE_SCOPES.SYNC_CLIENT}"]'::jsonb, '${SYSTEM_USER_UUID}' FROM sync_device_ticks GROUP BY device_id;
```
passcod, line 7: pretty sure we can use `bind` with `query()` instead of doing string substitution, be a bit safer

passcod (approved):
LGTM

tcodling:
bugbot run

PR #8339, opened 2025-08-28
feat(DHIS2): TAN-2540: Push reports to DHIS2 on schedule

tcodling:
### Changes

This contains the basic implementation for pushing reports to DHIS2

Included here is:
- Auth and host config
- Basic auth
- Posting the csv data to the host
- Error handling and logging

Out of scope and coming soon is
- Retry logic
- Auditing

Tested against a local DHIS2 instance running on Docker

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


rohan-bes (changes requested):
Looks great! Nice work!

This would be a good chance to add some unit tests, to me it makes sense to wrap them around the whole scheduled task. 

Got a couple options here:
- Move the `postToDhis2` function out into a separate class or even function might be fine and then mock that. Sometimes feels a bit awkward doing this, since it's kinda nice keeping it inside the Dhis2IntegrationProcessor, but its the easiest way to make it testable.
- Mock the `fetch` call instead. It might make the tests a bit more fiddly but its a more accurate test.

Might be other ways to approach it too :+1: lemme know what ya think

rohan-bes (approved):
Great tests! Awesome work :)

Couple minor comments but nothing blocking :+1:

PR #8350, opened 2025-08-29
feat: EPI-715: PatientData question type pulling through id instead of name

khoi-bes:
### Changes

- Breaking changes:
 + New components for PatientData questions in view-only mode. This component will be used for PatientData (which is able to write to patient) in the form response.
 + Fix the Patient source autocomplete question (.i.e mother, father,..) issue
 + Update translations for PatientData questions result
 + Refactor logic for SurveyAnswer question (which has source = PatientData question)

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


jaskfla:
bugbot run

packages/constants/src/surveys.ts
```diff
@@ -121,18 +132,24 @@ export const VITAL_CHARTS = {
...
-const makeLookupFields = (model: string, fields: string[]) =>
-  Object.fromEntries(fields.map(f => [f, [model, f]]));
+const makeLookupFields = (model: string, fields: (string | [string, Record<string, string>])[]) =>
+  Object.fromEntries(
+    fields.map(f => [Array.isArray(f) ? f[0] : f, [model, ...(Array.isArray(f) ? f : [f])]]),
+  );
 
 type PatientDataFieldLocationsType = {
-  [key: string]: Array<string>;
+  [key: string]: [string, string, Record<string, string>] | [string, string];
```
passcod, line 141: preference would be to change this to some kind of descriptive interface, but probably out of scope; at least can you add a comment describing what the different string fields of the array do please

packages/shared/src/reports/utilities/transformAnswers.js
```diff
@@ -1,22 +1,33 @@
...
+  format,
+  formatShort,
+  isISOString,
+  parseDate,
+} from '@tamanu/utils/dateTime';
+import { PATIENT_DATA_FIELD_LOCATIONS, PROGRAM_DATA_ELEMENT_TYPES } from '@tamanu/constants';
 
-// also update getNameColumnForModel in /packages/mobile/App/ui/helpers/fields.ts when this changes
-function getNameColumnForModel(modelName) {
+// also update getDisplayNameForModel in /packages/mobile/App/ui/helpers/fields.ts when this changes
```
passcod, line 12: similarly as constants, we can used shared package in mobile directly

passcod (changes requested):
a pretty large PR; could maybe do with some breaking apart. regardless, needs package reuse from shared/constants instead of doubling up. otherwise okay

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -141,3 +145,45 @@ surveyResponse.put(
...
+      throw new InvalidParameterError('Column parameter is required');
+    }
+    if (!value) {
+      res.json({
+        data: null,
+      });
+      return;
+    }
+
+    const [modelName, fieldName] = PATIENT_DATA_FIELD_LOCATIONS[column] || [];
```
passcod, line 168: we should probably return an error if the column is not available to be queried (ie not in the constant) rather than returning null

packages/facility-server/app/routes/apiv1/surveyResponse.js
```diff
@@ -141,3 +145,45 @@ surveyResponse.put(
...
+
+    if (!modelName) {
+      res.json({
+        data: value,
```
passcod, line 172: I'm also unsure why this is a thing. Why do we return the passed value if we don't find a model?

packages/mobile/App/migrations/1756663661549-addMotherFatherIdToPatientAdditionalData.ts
```diff
@@ -0,0 +1,62 @@
...
+          name: 'fatherId',
+          type: 'varchar',
+          length: '255',
+          isNullable: true,
+        }),
+      );
+    }
+
+        // Note: Foreign key constraints are handled by TypeORM at the application level
+    // Not creating database-level foreign keys to avoid constraint violations during data sync
```
passcod, line 42:
```suggestion
    // Note: Foreign key constraints are handled by TypeORM at the application level
    // Not creating database-level foreign keys to avoid constraint violations during data sync
```

passcod (approved):
a few more things but happy to call those optional to passing

passcod (approved):

PR #8355, opened 2025-08-31
feat: embed node engine in our dependencies

passcod:
### Changes

Because npm run's PATH includes all the relevant node_modules/.bin dirs this change means that we always use the required Node.js version.

It also means that we no longer need to install Node.js on Windows when deploying servers, and that updating the nodejs version here will update it on Windows deployments, too. That in turn means we can opt into newer Node.js much faster, without requiring extremely long runways for the deployment team to coordinate runtime upgrades.

Further, this commit changes from the legacy .nvmrc file to the common standard .node-version as described, with a symlink to keep nvm working: <https://github.com/shadowspawn/node-version-usage>

Finally, we add a CI check to make sure that the three different places that the node version is specified are kept in sync.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


jaskfla:
bugbot run

dannash100 (approved):
Amazing! cheers 💯

PR #8362, opened 2025-09-02
ci: Michael's demo

NavarroEmilioLuis:
### Changes

Branch created for a demo. NOTE: **DO NOT BRING DOWN THIS ENVIRONMENT** as it will be reused for a couple of weeks. Should anything happen - contact me or Michael directly to check whether it's fine to bring it down.

### Deploys


NavarroEmilioLuis:
Alright, we finally don't need this anymore

PR #8365, opened 2025-09-02
fix(desktop): SAV-1010: Download data button fix

NavarroEmilioLuis:
### Changes

So the issue happens because the table `data` had translation components and they don't work with how we download stuff. This happens in vitals/charts because the data object in those tables is different so this shouldn't be a problem elsewhere.

### Deploys


jaskfla:
bugbot run

rohan-bes (approved):
Oh yep makes sense :+1:

PR #8374, opened 2025-09-03
repo: upgrade to Node.js version 24

passcod:
### Changes

Experimental to see if it's possible.

Ideally we want to wait until we've deployed a release with #8355 to Windows servers, and verified that this works and makes use of the embedded Node.js version.

### Deploys

### Tests

- [ ] **Run E2E Tests**


passcod:
!lint-fix

PR #8416, opened 2025-09-09
refactor: create @tamanu/errors package

passcod:
### Changes

This is a load-bearing component of #8419.

- moves the errors out from shared
- converts the source to typescript
- implements the HTTP API Problem Details RFC
- narrows (makes more specific) a bunch of the authentication errors
- separates the concept of API errors (this package) and non-API errors (e.g. InvalidConfigError is not an API error, it's only used for tasks)
- makes the errors package fairly lightweight (single dependency we could do without if we cared) so it can go everywhere
- makes matching on errors use constants instead of random strings
- adds a serialisation/deserialisation so we can trivially have "proper" errors across remote calls to central/facility
- uses that to simplify a bunch of fetch code esp in sync (almost managed to delete as much code as I added!)
- adds a standard way to attach arbitrary extra data to errors
- uses that to simplify how version incompatibility errors work
- have a central concept of sensitive extra data that is removed in production
- (through fixing all the tests) add a bunch of debuggability aids to the errors

For example, this error in the code:

```js
import { NotFoundError } from '@tamanu/errors';
throw new NotFoundError('appointment gone').withExtraData({
	appointmentId,
	patientId,
});
```

Becomes this at the API layer:

```json
{
	"type": "/problems/not-found",
	"title": "Not Found",
	"status": 404,
	"detail": "appointment gone",
	"appointment-id": "fac0afda-1d70-410f-999b-e63310200475",
	"patient-id": "6d0ecbe2-9a1b-4b63-a9a2-d98269d6d474"
}
```

If you then fetch that using `@tamanu/api-client`, you'll get a `Problem` instance thrown (the response is parsed into something that almost matches the original object, across an HTTP boundary):

```text
Problem {
	type: 'not-found',
	title: 'Not Found',
	status: 404,
	detail: 'appointment gone',
	extra: {
		'appointment-id': 'fac0afda-1d70-410f-999b-e63310200475',
		'patient-id': '6d0ecbe2-9a1b-4b63-a9a2-d98269d6d474',
		'request-url': 'http://localhost:4000/api/whatever/endpoint',
	},
	response: Response { ... },
}
```

you can match on the `type` really easily thanks to the `ERROR_TYPE` constant enum:

```js
if (problem.type === ERROR_TYPE.NOT_FOUND) {
	//
}
```

and if you re-throw that `Problem` into an API, you'll get the same structure again:

```json
{
	"type": "/problems/not-found",
	"title": "Not Found",
	"status": 404,
	"detail": "appointment gone",
	"appointment-id": "fac0afda-1d70-410f-999b-e63310200475",
	"patient-id": "6d0ecbe2-9a1b-4b63-a9a2-d98269d6d474"
}
```

In development, the `request-url` and also the stack of the original Error will be left on and available at the API layer, so you automatically get more data out of it (but we don't expose too much stuff in production).

The old-style `error: {}` structure is also left on API responses transitionally (ie so that old versions of Tamanu clients can still get the errors when their servers get upgraded), and the FHIR and integration errors explicitly use their own style (so we don't expect to break any non-Tamanu clients).

The PR has grown somewhat through merging from main repeatedly.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/central-server/app/tasks/index.js
```diff
@@ -31,6 +31,8 @@ import { MedicationDiscontinuer } from './MedicationDiscontinuer';
 
 export { startFhirWorkerTasks } from './fhir';
 
+export class InvalidConfigError extends Error {}
+
```
NavarroEmilioLuis, line 35: Why don't we create an error in the package for this one?

NavarroEmilioLuis (approved):
Happy to stamp, just one ramble and one overall question on why we declare an error on the central package

edmofro (approved):
Extremely stampy but here you go, codeowner approved

PR #8419, opened 2025-09-10
feat(web): SAV-1008: new login errors (with lockout support)

passcod:
### Changes

Builds on top of #8416 which adds a way to carry arbitrary context with errors in a structured way, and unifies API errors across server and client.

Includes design changes to match new views from figma.

This sketches out a scheme where:

- if a user is fully locked out, the endpoint should return a `RateLimitedError`, with the normal `retry-after` extra data (in seconds)
- if a user has entered wrong credentials, the endpoint should return `InvalidCredentialError` (as currently) and additionally include two new extra data fields:
  - `lockout-attempts`, an integer number of remaining login attempts
  - `lockout-duration`, the number of seconds that a lockout lasts for

When those errors / extra data fields are received by the frontend, it renders the errors as per figma, and otherwise falls back to the normal error messages.

The cool thing about this approach is we totally can translate those errors; the PR doesn't do that _yet_ just to avoid churn if we change things around a bit.

### Testing notes

For this PR in isolation: login on both central and facility, ensure that works, then logout, then:
- fail to fill in either or both of the email/password, and check the fields go red with "Required" labels
- fill both fields but with an invalid email address e.g. `invalid`, and check the email field goes red with an appropriate label
- fill both fields but with a wrong password, and verify the style of the alert match the figma designs

The lockout functionality is _not_ implemented in this PR, so there's no need to test for those scenarios yet.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/forms/LoginForm.jsx
+++ b/packages/web/app/forms/LoginForm.jsx
@@ -92,9 +92,6 @@ const StyledCheckboxField = styled(Field)`
   }
 `;
 
-const INCORRECT_CREDENTIALS_ERROR_MESSAGE =
-  'Server error response: Incorrect username or password, please try again';
-
 const LoginFormComponent = ({
   errorMessage,
   onNavToResetPassword,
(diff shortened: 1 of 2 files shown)


NavarroEmilioLuis (approved):
This looks fine, there is a lint issue on the LoginForm file but that's ok.

The only other thing I was thinking is that we might be better off passing error details instead of creating the whole message so it can be translated later, but that's really secondary to this, we can leave that out for later

NavarroEmilioLuis (approved):

PR #8445, opened 2025-09-16
feat(translations): KOR-60: Import default translations in auto-deploys

rohan-bes:
### Changes

Reworked the upgrade step to rely on a local file for the default translations that will be put there during the build step.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


scripts/scrape-translations.mjs
```diff
@@ -0,0 +1,100 @@
...
+import path from 'path';
+import { execSync } from 'child_process';
+
+const constantsPackagePath = path.join(import.meta.dirname, '..', 'packages', 'constants');
+
+const enumTranslationsScript = path.join(
+  constantsPackagePath,
+  'scripts',
+  'printTranslatedEnums.ts',
+);
```
passcod, line 13: At this point we could just... import enums directly and run this whole thing in ts

and maybe produce JSON directly

rohan-bes: Do we still care about generating the translations delta between a version and the previous version?

passcod: We need some way of surfacing new translation strings but I'm not sure if that's the best tbh

packages/upgrade/src/steps/1752105049000-importDefaultTranslations.ts
```diff
@@ -114,50 +104,33 @@ export const STEPS: Steps = [
...
-          args.log.error(
-            'Failed to import default translations, you will need to manually import them',
-            {
-              error,
-            },
-          );
         }
+      } catch (error) {
+        args.log.error(
+          'Failed to import default translations, you will need to manually import them',
```
passcod, line 131: Given that's no longer really possible, maybe we should hard-error?

packages/upgrade/src/steps/scripts/getTamanaPackagesPath.ts
passcod, line 1: Typo in filename

packages/upgrade/src/steps/scripts/scrapeTranslations.ts
```diff
@@ -0,0 +1,105 @@
...
+  // Remove any newlines in the default text
+  const flattenDefaultText = (defaultText: string) => {
+    return defaultText
+      .split('\n')
+      .map(line => line.trim())
+      .join(' ');
+  };
+
+  const translationRows = Array.from(translations.values())
+    .map(({ stringId, defaultText }) => ({
```
passcod, line 98:
```suggestion
  const translationRows = Array.from(translations.values(), ({ stringId, defaultText }) => ({
```

scripts/docker-build.sh
```diff
@@ -54,11 +54,11 @@ build_server() {
   rm -rf packages/*/coverage || true
   rm -rf packages/*/config/{local,development,test}.* || true
 
-  remove_irrelevant_packages "$package"
-
   # build the world
   npm run build
 
+  remove_irrelevant_packages "$package"
+
```
rohan-bes, line 61: @passcod this is the simplest way to ensure all the translations are imported. Just build everything first before removing irrelevant packages. It's a bit wasteful though obviously as we build the irrelevant packages, so I can put some time into reworking this script to just build the relevant packages if you like?

Could also split out the scrap translations from the build process. One thing that I'm a little worried about is if we ever rebuild the `@tamanu/upgrade` package later on without realising we'll replace the default-translations.json with whatever's left here

passcod (approved):
Only optionals left

rohan-bes:
Thanks for the reviews @passcod :pray: 
I think I'm in agreement with you on those last few comments, but just super strapped for time at the moment and wanna get these in before the project cycle ends, so might leave them for now.

PR #8489, opened 2025-09-23
feat(patientPortal): NASS-1798: Patient portal settings

tcaiger:
### Changes

**New Patient Portal Settings API**
A new API endpoint /settings/:facilityId has been introduced in the central server to retrieve patient portal specific settings.

**Dynamic Settings in Patient Portal**
The patient portal now dynamically fetches settings via a new useSettings hook and SettingsProvider.

**Dynamic Survey Response Defaults**
The createSurveyResponse API now uses the newly available settings to dynamically determine default locationId and departmentId for survey responses.

**Settings Exposure Control**
A new constant KEYS_EXPOSED_TO_PATIENT_PORTAL has been added to explicitly define which settings keys are safe and relevant to be exposed to the patient portal frontend.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/patient-portal/src/views/SurveyView.tsx
```diff
@@ -85,14 +87,16 @@ export const SurveyView = () => {
...
         <AuthContext.Provider value={{ facilityId }}>
-          <SurveyForm
-            patientAdditionalData={additionalData}
-            encounterType={encounterType}
-            patient={patient}
-            survey={survey}
-            onSubmit={onSubmit}
-            onCancel={onCancel}
-          />
+          <SettingsProvider facilityId={facilityId}>
```
tcaiger, line 90: As far as I can tell, the survey only uses global settings at the moment so it might be overkill to create a different settings context for each survey based on the facility but it felt risky to pass in central-server settings to the survey. @edmofro I'd be keen to get your two cents on this?

packages/settings/src/reader/ReadSettings.ts
```diff
@@ -53,6 +55,11 @@ export class ReadSettings<Path = SettingPath> {
     return settings;
   }
 
+  async getPatientPortalSettings() {
+    const allSettings = await this.getAll();
+    return pick(allSettings, KEYS_EXPOSED_TO_PATIENT_PORTAL);
```
tcaiger, line 60: I didn't bother setting up another cache for the portal settings when the allSettings is already cached.

packages/patient-portal/src/views/SurveyView.tsx
edmofro: That's a super good catch on the security of providing settings to the patient portal. Does `SettingsProvider` do the fetch from the server, so it'll only get back the settings for the facility specified, or does it do the filtering on the front end? 

I wonder if even the full facility settings is risky tbh, we might need to allowlist the specific settings that the patient portal needs so we can assess each one as safe to sit on a client browser. Then again, I imagine we're already getting the full list on the browser for the main tamanu web app, plus I think we've been careful not to move any secrets across there yet.

Sorry, a bit to think about in this comment, we can talk in person if you prefer - but great catch and totally agree we should at least be limiting it to settings for that facility.

tcaiger: Yes it only returns from the server the settings that the portal needs based on a new allow list `['features', 'fileChooserMbSizeLimit']`. These settings are actually global so if we get them by facility with  a facility id or just get them as central server settings they will be the same. But conceptually and for maintainability I set them up with the facility id.

edmofro: Oh that's perfect!

packages/central-server/app/patientPortalApi/surveys/settings.js
tcodling, line 1: Is this specifically related to surveys? Feels kind of generic and almost certainly to be used elsewhere no?

packages/patient-portal/src/contexts/SettingsContext.tsx
tcodling, line 1: Awesome getting this working without redux. Would be awesome to be able to use this for web if possible but I am almost certain would be way too much work to integrate here 🤔

tcodling (approved):
Awesome work. My only real question was about the placement of the `settings` endpoint within the surveys folder but if there is a genuine reason for that I am missing I don't mind at all :)

packages/central-server/app/patientPortalApi/surveys/settings.js
tcaiger: yes this is definitely only for surveys at the moment and each survey actually needs to have it's own version of the settings if they relate to different facilityIds which is why the settings provider was only wrapped around the survey view in the front end and not the whole app. There are no settings in the app outside of the surveys yet.

PR #8491, opened 2025-09-23
feat(auth): SAV-31: Mobile app lockout

NavarroEmilioLuis:
### Changes

Alright, we won't cover offline login so this is just a bit of a UI update to have more descriptive error messaging

### Deploys

### DIFF
--- a/packages/errors/src/getLoginErrorMessage.ts
+++ b/packages/errors/src/getLoginErrorMessage.ts
@@ -1,0 +1,21 @@
+import { ERROR_TYPE } from './constants';
+import { BaseError } from './BaseError';
+import { Problem } from './Problem';
+
+export function getLoginErrorMessage(error: Problem | BaseError): string {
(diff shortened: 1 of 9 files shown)


packages/errors/src/getLoginErrorMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function getLoginErrorMessage(error: Problem | BaseError): string {
+  if (error instanceof BaseError) {
+    return error.message;
+  }
+  let message = error.message;
+  if (error.type === ERROR_TYPE.RATE_LIMITED) {
+    message = `You have been locked out due to too many log in attempts. Please try again in ${Math.ceil(
+      error.extra.get('retry-after') / 60,
+    )} minutes.`;
```
passcod, line 13: now that i'm looking at this again, maybe we should handle the plural here?

passcod (approved):
Cool!

PR #8527, opened 2025-10-02
fix: addPatientProgramRegistrationId migration error

passcod:
### Changes

If there's no PPR for a condition, we shouldn't fail the entire migration.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -367,7 +367,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Check spelling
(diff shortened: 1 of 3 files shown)


NavarroEmilioLuis (approved):
Happy to stamp. Though it is a bit tempting to just delete those conditions: there is no way they are showing on the frontend. We would still need to find a way to resolve them to successfully run the migration, too (and that is likely to be delete, or reassign otherwise)

PR #8567, opened 2025-10-07
feat(audit): COOL-8: Filter migration produced changelogs from frontend

NavarroEmilioLuis:
### Changes

We use logs for a bunch of stuff, however, whenever we show them on the frontend, we basically want to only show specific user interaction - now that we create logs in migrations, we don't want to show these so they need to be excluded

### Deploys


tcaiger (approved):
LGTM

PR #8572, opened 2025-10-09
fix(audit): COOL-25: Remove record_change trigger from logs.migrations

NavarroEmilioLuis:
### Changes

Pretty self explanatory

### Deploys


tcodling (approved):
looks great and pretty safe to me so approving so we can unblock main. Totally fine wit me if you want to get another opinion before merge though 🙏

PR #8592, opened 2025-10-15
fix(audit): TAMOC-352: Prevent audit logs being logged against the wrong user when queries are run within a transaction

rohan-bes:
### Changes

This one was a bit of a deep dive and tbh there's still something that doesn't fully make sense to me.

Previously we were sharing the same `AsyncLocalStorage` instance for the `SequelizeCLS` behaviour and for storing the `auditUserId`. The problem with this is that sequelize will create a new `run` of that `asyncLocalStorage` instance if we start a new transaction, effectively clearing out the initial `auditUserId`. This explains why we saw this issue when changing MedicationAdministrationRecords as there's changes made within transactions there.

There's no reason for us to use the same instance for both, so the main fix here is to create a separate `asyncLocalStorage` instance for the `auditUserId`.

The other fix was to set the `auditUserId` to `SYSTEM_USER_UUID` in the database once the query has finished running, otherwise that `id` would be used for any changes that didn't come in via a request endpoint (ie. system user changes).

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

---

> [!NOTE]
> Separate CLS contexts and set/clear audit user per query with transaction-aware logic; add helpers and expand tests for concurrent and transactional scenarios.
> 
> - **Database**:
>   - Replace shared CLS with dedicated `AsyncLocalStorage` for Sequelize transactions and separate store for audit user (`getAuditUserId`).
>   - Wrap query execution to set `public.set_session_config(AUDIT_USERID_KEY, userId, inTransaction)` only when transaction context is ready; reset to `SYSTEM_USER_UUID` after non-transaction queries.
>   - Add helpers: `sequelize.isInsideTransaction` and `sequelize.isTransactionReady`.
>   - Remove `SESSION_CONFIG_PREFIX`-based session config usage.
> - **Middleware**:
>   - New `auditUserIdAsyncLocalStorage` and `attachAuditUserToDbSession` uses `run(req.user?.id, next)`; export `getAuditUserId`.
> - **Tests**:
>   - Enhance concurrency and transaction coverage: add routes for updates in/without transactions and with custom isolation; verify `updated_by_user_id` matches changes across multiple simultaneous requests, including system-initiated updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 79c5c60dfe99a98f4d84e80a02152e0b0929bc48. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/database/src/services/database.js
+++ b/packages/database/src/services/database.js
@@ -3,7 +3,7 @@ import { Sequelize } from 'sequelize';
 import pg from 'pg';
 import util from 'util';
 
-import { SYNC_DIRECTIONS, AUDIT_USERID_KEY, SESSION_CONFIG_PREFIX } from '@tamanu/constants';
+import { SYNC_DIRECTIONS, AUDIT_USERID_KEY, SYSTEM_USER_UUID } from '@tamanu/constants';
 import { log } from '@tamanu/shared/services/logging';
 import { serviceContext, serviceName } from '@tamanu/shared/services/logging/context';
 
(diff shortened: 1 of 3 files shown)


packages/database/src/utils/audit/attachAuditUserToDbSession.ts
```diff
@@ -1,14 +1,15 @@
...
-import { AUDIT_USERID_KEY } from '@tamanu/constants/database';
+export const getAuditUserId = () => auditUserIdAsyncLocalStorage.getStore();
 
 export const attachAuditUserToDbSession = async (
   req: ExpressRequest,
   _res: Response,
   next: NextFunction,
 ) => {
-  setSessionConfigInNamespace(AUDIT_USERID_KEY, req.user?.id, next);
+  auditUserIdAsyncLocalStorage.run(req.user?.id, next);
```
rohan-bes: Decided to go with just storing the `id` here rather than a `Map` with key/value logic. If we end up wanting more 'request scoped' variables then maybe we can break this out and switch to a map. But since it's just this one variable at the moment it seems nicer to store just that.

packages/database/src/services/database.js
```diff
@@ -134,17 +127,24 @@ async function connectToDatabase(dbOptions) {
...
-        const userid = getSessionConfigInNamespace(AUDIT_USERID_KEY);
-        if (userid)
-          await super.run(
-            'SELECT public.set_session_config($1, $2)',
-            [AUDIT_USERID_KEY, userid],
-          );
+        const userid = getAuditUserId();
+        // Set audit userid so that any changes in this query are recorded against it
+        if (userid) {
+          await super.run('SELECT public.set_session_config($1, $2)', [AUDIT_USERID_KEY, userid]);
```
rohan-bes: So there's still a little question mark for me on why setting this value in the session doesn't interfere with other requests and occassionally mark the wrong user in the audit logs.

This value is 'session scoped' in postgres, which means it won't leak into other concurrent sessions. But it should be possible that if a separate thread uses this same sequelize connection in between this call completing and the cleanup, then queries run in that thread will having this `auditUserId`.

I tested this a bunch though, with adding sleeps are various intervals, and playing with the connection pool size, and couldn't seem to trigger that to happen. My only guess is that sequelize doesn't paralellise `query.run()` calls on a single connection, so we can assume all the queries in this method are run in serial and nothing else re-uses this connection in the meantime.

Will dig around in the sequelize docs to see if I can find anything about this.

passcod (approved):

tcodling:
bugbot run

PR #8605, opened 2025-10-17
fix(charting): TAMOC-358: Fix crash when recording chart entries on a chart with 'rounding' enabled

rohan-bes:
### Changes

Tbf I was a bit surprised that `Number.toFixed()` returns a string, but it does make sense.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/components/FormattedTableCell.jsx
+++ b/packages/web/app/components/FormattedTableCell.jsx
@@ -38,10 +38,12 @@ const HeadCellWrapper = styled.div`
 `;
 
 function round(float, { rounding } = {}) {
-  if (isNaN(float) || !isNumber(rounding)) {
+  const floatNumber = parseFloat(float);
+  if (isNaN(floatNumber) || !isNumber(rounding)) {
     return float;
   }
-  return float.toFixed(rounding);
+
+  return floatNumber.toFixed(rounding);
 }
 
 function getTooltip(float, config = {}, visibilityCriteria = {}) {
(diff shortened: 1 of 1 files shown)


packages/web/app/components/FormattedTableCell.jsx
```diff
@@ -97,11 +99,11 @@ export const DateHeadCell = React.memo(({ value }) => (
 export const DateBodyCell = React.memo(({ value, onClick }) => {
   const CellContainer = onClick ? ClickableCellWrapper : CellWrapper;
   return (
-    <TableTooltip title={DateDisplay.stringFormat(value, formatLong)} data-testid="tabletooltip-3knb">
-      <CellContainer
-        onClick={onClick}
-        data-testid="cellcontainer-slh4"
-      >
+    <TableTooltip
```
rohan-bes: Just formatting

packages/web/app/components/FormattedTableCell.jsx
```diff
@@ -38,10 +38,12 @@ const HeadCellWrapper = styled.div`
 `;
 
 function round(float, { rounding } = {}) {
-  if (isNaN(float) || !isNumber(rounding)) {
+  const floatNumber = parseFloat(float);
+  if (isNaN(floatNumber) || !isNumber(rounding)) {
```
rohan-bes: This feels like a much more defensive approach. Calling straight `isNaN` returns `false` on a surprisingly large number of things (eg. `isNaN(null) = false`, `isNaN(true) = false`, etc.)
Doing a `parseFloat` first will mean it only works for numbers and strings of numbers

tcaiger: I was so surprised that null and true return false but yeah

tcaiger (approved):

PR #8638, opened 2025-10-22
feat(invoicing): SAV-1023: Invoice price list matching

tcaiger:
### Changes

**New Price List Matching Logic**
Implemented matchesAgeIfPresent and equalsIfPresent functions to dynamically match invoice price lists based on patient age, type, and facility rules.

**Invoice Price List Model Enhancement**
The InvoicePriceList model now includes a static method getIdForPatientEncounter to find the most suitable price list based on patient data.

**Model Reorganization**
All Invoice-related models have been moved into a new src/models/Invoice subdirectory.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/database/src/models/Invoice/invoicePriceListMatching.ts
```diff
@@ -0,0 +1,53 @@
...
+  '>': (a: number, b: number) => a > b,
+  '>=': (a: number, b: number) => a >= b,
+  '=': (a: number, b: number) => a === b,
+};
+
+/**
+ * @param condition - Age condition string (e.g., ">=18", "<65", "=21")
+ * @param dob - patient date of birth as ISO date string (e.g., "2010-10-15")
+ */
+export const matchesAgeIfPresent = (
```
rohan-bes: I think we might have a bit of a problem with the way this is set up currently. 

In FSM we need 3 separate lists for 0-14, 15-64, 65+
If we 3 lists: 
  1: `age: <=14`
  2: `age: undefined`
  3: `age: >= 65`

Then anyone who's 14 or younger, or 65+ will match with list 2 as well as their respective list, which throws an error. We need list 2 to be able to have a filter saying `> 14 AND < 65`

I'm a big fan of expressions, but this is a case I might be leaning towards just using an object to represent this? eg.
  1: `age: { max: 14 }`
  2: `age: { min: 15, max: 64 }`
  3: `age: { min: 65 }`

Could handle `=` too with just `age: 30`

rohan-bes (changes requested):

packages/database/src/models/Invoice/invoicePriceListMatching.ts
tcaiger: Good catch - I've updated it and it's ready for review

rohan-bes (approved):
Thanks!

PR #8654, opened 2025-10-23
feat(meta): KAM-429: Use random jitter to report to meta server

NavarroEmilioLuis:
### Changes

We already had a way to specify through config, though that's manual. So I decided to make it a default value instead that gets set on initialization, I think the 0-5s range should be enough.

### Deploys

### DIFF
--- a/packages/central-server/config/default.json5
+++ b/packages/central-server/config/default.json5
@@ -288,6 +288,7 @@
       "enabled": true,
       // every minute
       "schedule": "* * * * *",
+      jitterTime: "30s",
     },
     "medicationDiscontinuer": {
       // every hour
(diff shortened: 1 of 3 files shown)


passcod:
Oh, it _is_ already there. Maybe this PR should instead be to add the default to the default.json5 configs. And given this runs every minute, maybe 30s instead, so it spreads load over a wider range.

NavarroEmilioLuis:
> Oh, it _is_ already there. Maybe this PR should instead be to add the default to the default.json5 configs. And given this runs every minute, maybe 30s instead, so it spreads load over a wider range.

Right yeah that probably makes more sense - do you think it should be added to all schedules or just this one?

passcod (approved):

anilvkaran (approved):

PR #8657, opened 2025-10-27
tweak(audit): NASS-1836: Allow changing multiple encounter columns at once

tcodling:
### Changes

In order to proceed with the encounter enhancements project, we need to support changing multiple column values for an encounter in one update which is currently restricted. This is because `EncounterHistory` includes a `changeType` that assumes only one column will change at a time.

In an ideal world we would fully migrate to using logs.changes however given the short time left in the cycle we have decided to update `EncounterHistory` to support an array of types in `changeType`. This includes

- Data migration to turn update the column and existing `changeType` values into arrays containing the stored value
- Swap out places that reference changeType to treat as an array

Most of the work is actually on the Maui side updating reports

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

---

> [!NOTE]
> Enable multiple encounter fields to change in one update by converting `EncounterHistory.changeType` to an array and updating server, reports, mobile sync, and tests; remove deprecated changelog migration command.
> 
> - **Database**:
>   - Migrate `encounter_history.change_type` from `text` to `text[]` with up/down scripts.
>   - Update dbt model (`database/model/public/encounter_history.yml`) to `data_type: array` and conditional `not_null`.
> - **Backend (Central/Facility)**:
>   - `Encounter.update()` now supports multiple simultaneous changes; aggregates `changeType` values and snapshots once with an array.
>   - Initial snapshots created with `NULL` `changeType`.
>   - `EncounterHistory` model `changeType` now `string[]`; `createSnapshot` accepts `string[]`.
>   - Remove restriction/error on multi-field updates.
>   - Adjust tests to expect array `changeType` and remove deprecated multi-change error case.
> - **Reports (FHIR Medici)**:
>   - Update SQL to filter with `'<type>' = ANY(change_type)` instead of equality.
> - **Mobile**:
>   - Rename relation to `encounterHistories` on `Encounter`.
>   - `EncounterHistory.changeType` stored as CSV string; add sanitizers to convert to/from arrays during sync.
> - **CLI**:
>   - Remove deprecated `migrateChangelogNotesToEncounterHistory` command and its tests; update hidden/exports.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fb9c038ff8726a88e1c4c730021676922bc345bd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/central-server/app/subCommands/migrateChangelogNotesToEncounterHistory.js
tcodling: This is a 2 year old data migration subcommand. I can't see any reason it would be run again?

packages/mobile/App/models/EncounterHistory.ts
```diff
@@ -57,8 +57,8 @@ export class EncounterHistory extends BaseModel {
   @Column({ type: 'varchar', nullable: false })
   encounterType: EncounterType;
 
-  @Column({ type: 'varchar' })
-  changeType: EncounterChangeType;
+  @Column({ nullable: true })
+  changeType?: string;
```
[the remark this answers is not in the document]
tcodling: hmm I did this in line with other mobile models with string arrays. Will do some testing

packages/database/src/migrations/1761600704418-updateChangeTypeToArray.ts
```diff
@@ -0,0 +1,23 @@
+import { QueryInterface } from 'sequelize';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    ALTER TABLE encounter_history 
+    ALTER COLUMN change_type TYPE text[] 
+    USING CASE 
+      WHEN change_type IS NULL THEN NULL
+      ELSE ARRAY[change_type]
```
NavarroEmilioLuis: This seems a bit weird to me? I would've assumed that we were going to make `change_type` to default to an empty array as default and not accept `null`, which would allow us to always rely on it being an array. What would be the added value or the difference from `null` to an empty array (which I don't think would be possible?)

packages/database/src/models/Encounter.ts
```diff
@@ -651,38 +651,24 @@ export class Encounter extends Model {
...
-      }
-
-      // multiple changes in 1 update transaction is not supported at the moment
-      if (snapshotChanges.length === 1) {
+      // Create snapshot with array of change types
+      if (changeTypes.length > 0) {
         await EncounterHistory.createSnapshot(updatedEncounter, {
           actorId: user?.id,
-          changeType,
+          changeType: changeTypes,
```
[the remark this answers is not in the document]
NavarroEmilioLuis: This is true, wouldn't hurt? (the comment made by AI, that we are not returning the actual encounter within the transaction)

NavarroEmilioLuis (commented):
It looks good, I understand drawing from the previous model where we used null to probably specify the initial encounter? But it does read a bit weird when we are not expecting an empty array.

I'd be happy with this as it is, but yeah just to double check how you feel about it + some other minor comments

packages/database/src/migrations/1761600704418-updateChangeTypeToArray.ts
tcodling: Actually that sounds way better. Not really sure why I didnt do it like this

NavarroEmilioLuis (approved):
Sweet!

PR #8669, opened 2025-10-30
feat(invoicing): SAV-1037: Insurance contracts and items DB models

NavarroEmilioLuis:
### Changes

Very similar to the price lists

### Deploys


tcaiger (approved):
Nice one - all looks good to me

PR #8692, opened 2025-11-04
fix: Update provisioning

NavarroEmilioLuis:
### Changes

Alright so yeah this is why you don't merge if deploy is failing. Had to update the provisioning file with proper location group isBookable config.

Note how CD was successful on the first commit - the second one is just fixing the typos.

Also note that the issue here wasn't the update to the provisioning file, it was the lack of update to the provisioning file after changing the importer rules for the `isBookable` config. Plus, someone managed to merge some typos and it passed review too

### Deploys


tcodling (approved):
Thanks

PR #8738, opened 2025-11-11
feat(invoicing): SAV-1017: Automatic invoice creation

NavarroEmilioLuis:
### Changes

Alright, I didn't update this description on time. But here goes nothing: general idea was refactoring to have a common way of creating an invoice. Originally I thought about doing this inside the encounter model, but the thing is that it ends up being very awkward cause we need to access a bunch of things like settings, which doesn't really belong to the model itself.

The good thing is that everything is contained on that encounter endpoint. The other place where we create encounters is just for the vaccinations so it doesn't need to be modified there.

### Deploys


tcaiger (approved):
LGTM

packages/database/src/models/Invoice/Invoice.ts
```diff
@@ -186,4 +191,49 @@ export class Invoice extends Model {
...
+    data: any,
+  ) {
+    const insurerId = await this.sequelize.models.PatientAdditionalData.findOne({
+      where: { patientId: encounter.patientId },
+      attributes: ['insurerId'],
+    }).then(patientData => patientData?.insurerId);
+    const insurerPercentage = await facilitySettings.get(
+      SETTING_KEYS.INSURER_DEFAULT_CONTRIBUTION as SettingPath,
+    );
+    const defaultInsurer =
```
tcaiger: As we discussed on the other PR, patient insurers aren't supported at the moment so we can remove the code relating to them which will simplify this a bit.

PR #8768, opened 2025-11-17
feat(sync): TAMOC-371: Mark sync session as errored if error occurs on the facility-server

rohan-bes:
### Changes

Currently we never see the actual error from the facility-server in the sync_session, and we just rely on the session being marked as overriden.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/app/sync/FacilitySyncManager.js
+++ b/packages/facility-server/app/sync/FacilitySyncManager.js
@@ -156,11 +156,18 @@ export class FacilitySyncManager {
       startedAtTick: newSyncClockTime,
     });
 
-    await this.pushChanges(sessionId, newSyncClockTime);
+    try {
+      await this.pushChanges(sessionId, newSyncClockTime);
 
-    await this.pullChanges(sessionId);
-
-    await this.centralServer.endSyncSession(sessionId);
+      await this.pullChanges(sessionId);
+      await this.centralServer.endSyncSession(sessionId);
+    } catch (error) {
+      await this.centralServer.markSessionErrored(sessionId, error.message);
+      throw error;
+    } finally {
+      // clear temp data stored for persist
+      await dropSnapshotTable(this.sequelize, sessionId);
+    }
 
     const durationMs = Date.now() - startTime;
     log.info('FacilitySyncManager.completedSession', {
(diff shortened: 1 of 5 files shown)


packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -156,11 +156,18 @@ export class FacilitySyncManager {
...
 
-    await this.centralServer.endSyncSession(sessionId);
+      await this.pullChanges(sessionId);
+      await this.centralServer.endSyncSession(sessionId);
+    } catch (error) {
+      await this.centralServer.markSessionErrored(sessionId, error.message);
+      throw error;
+    } finally {
+      // clear temp data stored for persist
+      await dropSnapshotTable(this.sequelize, sessionId);
```
rohan-bes: So I assume this is desired? I think previously we wouldn't have cleaned this up if we hit an error during the persist phase, but that seems odd... unless we use it for debugging or something?

edmofro: No you're right to do this, can you check out central for the same issue? We've seen a big build up of snapshot tables there in some deployments (unless @passcod you've already followed up on that one)

edmofro: You were seeing it on central though hey? But yeah same pattern I imagine.

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -156,11 +157,21 @@ export class FacilitySyncManager {
...
+      await this.pushChanges(sessionId, newSyncClockTime);
+
+      await this.pullChanges(sessionId);
+      await this.centralServer.endSyncSession(sessionId);
+    } catch (error) {
+      if (!(error instanceof Problem)) {
+        // if the error is not a Problem, it occurred locally on the facility-server and we should notify the central server
+        await this.centralServer.markSessionErrored(sessionId, error.message);
+      }
+      throw error;
```
rohan-bes: @passcod could I get your input on this? My impression with the `Problem` class is that it's a wrapper around errors from the TamanuApi. So if an error is a `Problem`, we can assume it came from the TamanuApi?

Probably ideally we'd implement a specific `Problem` type for sync errors that came from central-server and check for that, but I wasn't feeling like making all those changes right now.

chris-bes (approved):
All lgtm! Thanks for this change Rohan <3

PR #8797, opened 2025-11-21
tweak(docs): reduce conventional commit topic list

passcod:
### Changes

From discussion in slack, usage numbers show that we actually use much fewer commit topics than the full list we had at the beginning. Let's slim it down.


passcod:
Noting wrt gemini: the cliff.toml has to stay with the full list for a little while because otherwise changelog generation will break.

dannash100 (approved):
💯

PR #8824, opened 2025-11-26
feat(labs): NASS-1837: Lab results interpretation

dannash100:
### Changes
- Add results interpretation on enter results modal
- Add column to mobile too

Screenshot 2026-01-05 at 1 22 42 PM

BRB: Just back adding this to results pdf

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/mobile/App/migrations/1763435712934-addLabRequestResultsInterpretation.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const tableName = 'lab_requests';
+const columnName = 'resultsInterpretation';
+
+export class addLabRequestResultsInterpretation1763435712934 implements MigrationInterface {
+  async up(queryRunner: QueryRunner): Promise<void> {
+    const tableObject = await queryRunner.getTable(tableName);
+    const existingColumns = tableObject.columns.map(col => col.name);
+
+    if (!existingColumns.includes(columnName)) {
```
NavarroEmilioLuis: Hmmmm checking this feels a bit off, migration should just run once so it's not possible the column already existed, unless I'm missing something?

I can see this approach was recently(ish) added. I can't find a proper justification for it but I guess happened because the migration errored, so their database ended up corrupted (based on renaming the migration and having to tinker with pragma).

We might want to error if the column already exists as it's an unexpected state, I don't think this would be possible in production, and if it is possible, what are we not considering?

Lastly, is this a tell that migrations aren't done inside a transaction in mobile?

NavarroEmilioLuis (approved):
This LGTM, however there is a thing with the mobile migration that I think should be addressed. In all honesty, it shouldn't cause an issue, but I'm just wary of introducing a possibly harmful pattern and/or not addressing the issue.

I'm still approving as that's just my opinion and you might feel different about it or there is something I'm missing. Everything else is good.

PR #8785, opened 2025-11-19
feat(invoicing): SAV-1051: Patient level invoice UI

tcaiger:
### Changes

This PR is to add deep link from patient level invoicing table to encounter invoices.

Including:
- Load encounter data when the encounter view mounts
- Update the calculation for the patient level invoicing table
- Fix up the finalised invoice screen

### Deploys

- [ ] **Deploy to Tamanu Internal** 

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests
- ...add UI screenshots and **testing notes** to the Linear issue
- ...add any **manual upgrade steps** to the Linear issue
- ...update the [config reference](https://beyond-essential.slab.com/posts/reference-config-file-0c70ukly), [settings reference](https://beyond-essential.slab.com/posts/reference-settings-0blw1x2q), or any [relevant runbook(s)](https://beyond-essential.slab.com/topics/runbooks-bs04ml6c)
- ...call out additions or changes to **config files** for the deployment team to take note of

---

> [!NOTE]
> Enable navigating from patient invoices to the encounter’s invoicing tab, load encounters on mount, hydrate invoices with price list and insurance data, and streamline invoice display/printing.
> 
> - **Backend (API)**
>   - Add `invoice/getInsurancePlanItems` and use it to enrich invoice items with insurance plan coverage in `encounter/:id/invoice` and patient invoice routes.
>   - Patient invoice endpoints now hydrate invoices with price list–dependent associations and compute total outstanding balance from hydrated data.
>   - Minor cleanup in encounter creation for invoice initialization.
> - **Shared**
>   - Simplify `getInvoiceSummaryDisplay` to always format numeric summary values.
> - **Web (Frontend)**
>   - Patient invoices table: clicking a row navigates to the encounter’s invoicing tab (`navigateToEncounter` + `setNavigateBackTab`).
>   - Encounter view: load encounter on mount via route param.
>   - Invoicing pane: add print button/modal for non-editable invoices; show loading state and create-invoice CTA when absent; streamline form actions.
>   - Invoice print modal: rely on encounter context and simplify loading state.
>   - Minor UI polish in dropdown button.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5118a65fb4cece1368e9594aa1a212fd3d91c861. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8826, opened 2025-11-27
fix(invoicing): Show create invoice modal

NavarroEmilioLuis:
### Changes

So the issue came from: https://github.com/beyondessential/tamanu/pull/8785/files#diff-f79905d6363c8690e631c32e44dff940796211770d1b4380eca2af590e69ffe0 essentially, we are not rendering the modal anymore. Here we retcon that change

### Deploys


tcaiger (approved):
Thanks for fixing that. I hadn't tested that use case of turning off the feature flag and starting an encounter

PR #8869, opened 2025-12-05
fix(invoicing): SAV-1029: Update the invoice item when a procedure type is changed

rohan-bes:
### Changes

Was missing an update hook for procedures

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/database/src/models/Invoice/Invoice.ts
```diff
@@ -180,20 +180,21 @@ export class Invoice extends Model {
...
-    const invoicePriceListId = await this.sequelize.models.InvoicePriceList.getIdForPatientEncounter(encounterId);
+    const invoicePriceListId =
+      await this.sequelize.models.InvoicePriceList.getIdForPatientEncounter(encounterId);
     if (!invoicePriceListId) {
       return;
     }
 
     // Confirm invoice product is not configured to be hidden for this price list
-    const invoicePriceListItem = await this.sequelize.models.InvoicePriceListItem.findOne({
+    const hiddenInvoicePriceListItem = await this.sequelize.models.InvoicePriceListItem.findOne({
```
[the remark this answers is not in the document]
NavarroEmilioLuis: Short answer: I don't know, the visibility card didn't have any explicit info about this particular scenario. But you might be right!

Long answer: Thinking about it, I guess it's possible to have invoice products imported without an entry on the price items table, previously we skipped empty cells but now we always create a price list item, but if the whole row is missing then the junction (price list item) will not be generated.

This is already something that I don't quite like because to me it sounds more like a reference data issue (data is incomplete) and/or importer issue (we should error on missing entries because automatically fill in the blanks sounds fraught too).

I can understand treating missing items as manual entry due to complexity, in the sense that you might want to look it up and add it. But for automatically added items it's weird: if you wanted to have that item automatically added it should have been on the reference data? And also seems weird to automatically add stuff that still requires manual entry anyway?

Ultimately I think you're right, these are not specified hidden, and the automatic lists are meant to save some work

packages/facility-server/__tests__/apiv1/EncounterInvoice.test.js
```diff
@@ -101,178 +103,256 @@ describe('Encounter invoice', () => {
...
+              sourceRecordType: 'Procedure',
+              productId: procedureProduct2.id,
+              orderedByUserId: user.id,
+              quantity: 1,
+              insurancePlanItems: [],
+            },
+          ],
+        });
+
+        // Switching to a procedure type that does not have a price list item should remove the item from the invoice
```
NavarroEmilioLuis: Hmmm I'm not sure this is true? I believe that if the invoice product exists, we basically don't care if it has a price list or price list item entry, we'll treat that as a manual entry price 🤔

NavarroEmilioLuis (commented):
I'm not sure I get why procedures update is different from the others where we would remove or add the item instead of both? Do we have to revisit the other models too?

Also, I left a note on a test to double check behavior and possibly a suggestion

packages/facility-server/__tests__/apiv1/EncounterInvoice.test.js
rohan-bes: Yeah good catch, this comment made based on the previous behaviour, but I didn't update it when i changed that. Will update

packages/database/src/models/Procedure/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+      previousValues.procedureTypeId &&
+      previousValues.procedureTypeId !== instance.procedureTypeId
+    ) {
+      // Ensure we remove the item from the invoice first, in case the new procedure type is not invoiceable
+      await removeFromInvoice(instance);
+    }
+
+    await addToInvoice(instance);
+  });
+};
```
NavarroEmilioLuis: Nice!

NavarroEmilioLuis (approved):
LGTM

PR #8874, opened 2025-12-08
fix(invoicing): SAV-1031: Filter invoice products properly

NavarroEmilioLuis:
### Changes

So there was a bit of confusion with the first iteration of this. Here we change the import/export rules to treat empty cells as manual price entries and keep the "hidden" text to hide the invoice product when it's being looked up for a specific price list.

Also included is to prevent automatically adding items that are also configured as hidden.

### Deploys

---

> [!NOTE]
> Replace `visibilityStatus` with boolean `isHidden` for invoice price list items, adjust import/export semantics (empty = manual price, `hidden` = isHidden), and filter hidden products from auto-adds, API, suggesters, and UI.
> 
> - **Database/Models**
>   - Replace `invoice_price_list_items.visibility_status` with boolean `is_hidden`; add migration and update model/DBT docs (`invoice_price_list_items.md/.yml`).
> - **Import/Export**
>   - Import: treat empty cells as manual (null price) and `hidden` as `{ isHidden: true }`; allow empty values; remove `manual-entry` constant.
>   - Export: output `hidden` when `isHidden` else price; include `isHidden` attribute.
> - **Server/API Logic**
>   - Prevent auto-adding invoice items when corresponding `InvoicePriceListItem.isHidden` is true.
>   - Price lookup and includes now filter by `isHidden: false`.
>   - Remove `visibilityStatus` checks from invoice item associations.
> - **Suggestions/UI**
>   - `invoiceProduct` suggester accepts `priceListId` and only returns products with `isHidden: false` for that list.
>   - Invoice form passes `priceListId` to suggester; fetch price list item price accordingly.
> - **Tests**
>   - Update/import/export tests to new semantics; add tests for hidden items not auto-adding and suggester filtering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit acfc61befdab91c79b46376f2fd4fa58243c6aef. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/database/src/models/Invoice/Invoice.ts
+++ b/packages/database/src/models/Invoice/Invoice.ts
@@ -176,6 +176,24 @@ export class Invoice extends Model {
       return;
     }
 
+    // Confirm price list exists
+    const invoicePriceListId = await this.sequelize.models.InvoicePriceList.getIdForPatientEncounter(encounterId);
+    if (!invoicePriceListId) {
+      return;
+    }
+
+    // Confirm invoice product is not configured to be hidden for this price list
+    const invoicePriceListItem = await this.sequelize.models.InvoicePriceListItem.findOne({
+      where: {
+        invoicePriceListId,
+        invoiceProductId: invoiceProduct.id,
+        isHidden: false,
+      },
+    });
+    if (!invoicePriceListItem) {
+      return;
+    }
+
     await this.sequelize.models.InvoiceItem.upsert(
       {
         invoiceId: invoice.id,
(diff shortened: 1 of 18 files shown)


packages/database/src/models/Invoice/Invoice.ts
```diff
@@ -176,6 +176,24 @@ export class Invoice extends Model {
...
     }
 
+    // Confirm price list exists
+    const invoicePriceListId = await this.sequelize.models.InvoicePriceList.getIdForPatientEncounter(encounterId);
+    if (!invoicePriceListId) {
+      return;
+    }
+
+    // Confirm invoice product is not configured to be hidden for this price list
+    const invoicePriceListItem = await this.sequelize.models.InvoicePriceListItem.findOne({
```
tcaiger: This works well in the end to have the filtering for automatic items in the model so that it's in one place

tcaiger (approved):
Nice work - it's a pretty confusing feature but all looks good to me!

PR #8885, opened 2025-12-11
fix(unifiedMove): no-issue: Triage change bug

tcodling:
### Changes

This was a bit of a weird one. I was getting really confusing behaviour whenever running `Encounter.closeTriage()`. Basically forms that triggered this were not working on the first submission. This happened on the discharge and move form.

After some investigation I discovered that all the weird behaviour went away after no longer supplying user to the `Triage.update` inside closeTriage. While I am not completely sure how this was related, I didn't actually need to supply the user here as we don't record system notes for this action so I have removed it. I also updated the `Triage` and `Encounter` model args to properly reflect the optional nature of the user arg.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

---

> [!NOTE]
> Refactors triage closing to not require a user, inlines triage lookup, makes update user param optional, and removes practitioner/closed-time change tracking in triage updates.
> 
> - **Encounter model (`packages/database/src/models/Encounter.ts`)**:
>   - Refactor `closeTriage` to inline triage lookup and remove `user` dependency; update triage via `triage.update({ closedTime })`.
>   - Update `onDischarge` and encounter type change handler to call `closeTriage(endDate)` (no user).
>   - Make `update(data, user)` accept an optional `user`.
> - **Triage model (`packages/database/src/models/Triage.ts`)**:
>   - Make `update(data, user)` accept an optional `user`.
>   - Remove change tracking for `practitionerId` and `closedTime`.
>   - Continue change tracking for complaints, arrival time/triage time, score; add arrival mode tracking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f5182fb30f4da48b8f1f83808dc876382d2cb03. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/database/src/models/Encounter.ts
+++ b/packages/database/src/models/Encounter.ts
@@ -489,7 +489,7 @@ export class Encounter extends Model {
     await triage.update({ closedTime: endDate });
   }
 
-  async update(data: any, user: any): Promise<any> {
+  async update(data: any, user?: any): Promise<any> {
     const { Department, Location, EncounterHistory, ReferenceData, User } = this.sequelize.models;
     // Track change types for encounter history snapshot
     const changeTypes: string[] = [];
--- a/packages/database/src/models/Encounter.ts
+++ b/packages/database/src/models/Encounter.ts
@@ -453,15 +453,6 @@ export class Encounter extends Model {
     });
   }
 
-  async getLinkedTriage() {
-    const { Triage } = this.sequelize.models;
-    return Triage.findOne({
-      where: {
-        encounterId: this.id,
-      },
-    });
-  }
-
   async onDischarge(
     {
       endDate,
@@ -483,20 +474,19 @@ export class Encounter extends Model {
     });
 
     await this.addSystemNote(systemNote || 'Discharged patient.', submittedTime, user);
-    await this.closeTriage(endDate, user);
+    await this.closeTriage(endDate);
   }
 
-  async closeTriage(endDate: string, user: ModelProperties<User>) {
-    const triage = await this.getLinkedTriage();
+  async closeTriage(endDate: string) {
+    const { Triage } = this.sequelize.models;
+    const triage = await Triage.findOne({
+      where: { encounterId: this.id },
+    });
+
     if (!triage) return;
     if (triage.closedTime) return; // already closed
 
-    await triage.update(
-      {
-        closedTime: endDate,
-      },
-      user,
-    );
+    await triage.update({ closedTime: endDate });
   }
 
   async update(data: any, user: any): Promise<any> {
@@ -601,7 +591,7 @@ export class Encounter extends Model {
         formatText: capitalize,
         changeType: EncounterChangeType.EncounterType,
         onChange: async () => {
-          await this.closeTriage(data.submittedTime, user);
+          await this.closeTriage(data.submittedTime);
         },
       });
       await onChangeForeignKey({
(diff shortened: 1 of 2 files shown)


packages/database/src/models/Triage.ts
```diff
@@ -129,8 +129,8 @@ export class Triage extends Model {
     });
   }
 
-  async update(data: any, user: any): Promise<any> {
-    const { Encounter, ReferenceData, User } = this.sequelize.models;
+  async update(data: any, user?: any): Promise<any> {
```
tcodling: Ok Ill be honest I am not totally sure why this fixes it. I stopped supplying user for the closeTriage method on Encounter.ts (which wouldn't generate a user note beforehand anyway).

packages/database/src/models/Encounter.ts
```diff
@@ -453,15 +453,6 @@ export class Encounter extends Model {
...
   }
 
-  async getLinkedTriage() {
-    const { Triage } = this.sequelize.models;
-    return Triage.findOne({
-      where: {
-        encounterId: this.id,
-      },
-    });
-  }
```
tcodling: Cleaned this up while playing around as felt a bit unnecessary. Can put back

tcodling:
bugbot run

packages/database/src/models/Encounter.ts
```diff
@@ -483,23 +474,22 @@ export class Encounter extends Model {
...
-      {
-        closedTime: endDate,
-      },
-      user,
-    );
+    await triage.update({ closedTime: endDate });
   }
 
-  async update(data: any, user: any): Promise<any> {
+  async update(data: any, user?: any): Promise<any> {
```
tcodling: This is actually currently treated as optional. Its just mostly used in js files. Updating this to properly reflect the current usage

packages/database/src/models/Triage.ts
```diff
@@ -141,12 +141,6 @@ export class Triage extends Model {
     );
 
     const updateTriage = async () => {
```
tcodling: I added a couple extra change columns not in the brief and dont need notes recorded

packages/database/src/models/Encounter.ts
NavarroEmilioLuis: Yeah okay this looks small enough though it does give you information about the relation and the purpose. If it had more query logic it would be more meaningful to keep

packages/database/src/models/Encounter.ts
NavarroEmilioLuis: Right that makes sense

NavarroEmilioLuis (approved):
The change looks sane. Is the error you're getting just generated by typescript complaining on `addSystemNote` method that it doesn't have a user? That also would need changing anyway.

Ultimately I think it's okay to keep the note, but I don't know about the project as a whole so your call.

tcodling:
@NavarroEmilioLuis Na that isn't the error because it is actually valid to record a system note without a user for purely system generated actions like those generated from tasks.

> The change looks sane. Is the error you're getting just generated by typescript complaining on `addSystemNote` method that it doesn't have a user? That also would need changing anyway.
> 
> Ultimately I think it's okay to keep the note, but I don't know about the project as a whole so your call.

NavarroEmilioLuis:
@tcodling yeah but it's not marked as optional, I don't know if it was throwing or something.

I think the problem here is that not knowing the cause might mean we're shooting ourselves in the foot? The triage close action is technically triggered by the user changing the encounter type, so I'd expect to see that in the note even though it's not essential to it?

I'm happy if you want to leave as is, it's just a caution warning

PR #8949, opened 2026-01-07
feat(PDFs): PHX-270: Ability to configure logo asset per facility

NavarroEmilioLuis:
### Changes

Previously logos were universal, in this way, logos used in a facility will first try to get their specific configured logo, otherwise fallback to the generic one. These assets are also used on central but they are product of certificate notifications and those will only use the generic ones per discussion.

### Deploys


tcaiger (approved):
Looks all good to me

PR #8955, opened 2026-01-08
fix(desktop): PHX-282: Improve login error message

NavarroEmilioLuis:
### Changes

This is covering a specific scenario: a failed login from an unregistered device will fail and give out a cryptic error. Here we are graciously failing and just sending back the normal invalid credentials which is what is desired.

### Deploys


edmofro (approved):
lgtm!

PR #8974, opened 2026-01-13
fix(sync): KAM-463: validate that all data pushed is both for a known and allowed model

passcod:
### Changes

Added validation logic in the `addIncomingChanges` method to reject records with disallowed sync directions (`PULL_FROM_CENTRAL` and `DO_NOT_SYNC`). This happens even before the data is stored in snapshot. When we reject, we fail the entire sync session and log in the session table which specific record was rejected.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/__tests__/sync/CentralSyncManager.addIncomingChanges.test.js
+++ b/packages/central-server/__tests__/sync/CentralSyncManager.addIncomingChanges.test.js
@@ -25,7 +25,6 @@ describe('CentralSyncManager.addIncomingChanges', () => {
   });
 
   beforeEach(async () => {
-    jest.resetModules();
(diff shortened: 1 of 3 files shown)


edmofro (approved):
I think this is a little over-coded (in the AI way)
- More debug info than is necessary for each record
- I probably would have just detected the first issue, thrown that error then bailed
- The first test (single bad record) could be dropped given it is covered by the second test (multiple bad records)

But look, none of these are bad things, they're just more code than I would have written. And I guess when writing the code is the fast bit, telling it "be lazier" is actually creating work, so...approved

Thanks for pushing this through quickly :-)

PR #8985, opened 2026-01-14
feat(deathData): PHX-273: Update death record workflow 

NavarroEmilioLuis:
### Changes

Mostly UI, but includes adding/removing some fields. Too many individual changes to be listed but essentially all changes come from Figma and also all commits are very atomic.

### Deploys


tcodling (approved):
This looks great to me! Really nice commits 😆 

I maybe have the slightest concern at just the amount of columns on this table and the direction its heading in if we keep adding them? Maybe not a real problem though.

I haven't done any death work really and I know this is supposed to be matching some international standards so maybe this is the best way. Just feels a bit hectic coming in from the outside.

PR #9028, opened 2026-01-21
tweak(cd): skip building images when we can

passcod:
### Changes

Some file changes don't affect images. In those cases, we try to look up an already-built image from the last 50 commits, and if we find one, we skip building images altogether and proceed to the deploy.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### 🦸 Review Hero

- [ ] **Run Review Hero** 
- [ ] **Auto-fix review suggestions** 
- [ ] **Auto-fix CI failures**


NavarroEmilioLuis (commented):
I'm not too familiar with these, following your comments things make sense to me. However I did ask Thomas to bugbot this, and also will post here two other AI suggestions I got when looking up syntax:

1. Improve File Pattern Matching
In the `check-image-changes` step, the script uses bash pattern matching `([[ "$file" == $pattern ]])` to check if changed files should be ignored. Patterns like docs/**/* are used to match files in subdirectories.

The Issue: The ** glob for recursive directory matching is only enabled when the globstar shell option is set. It's not on by default in most environments, including GitHub Actions runners. Without it, patterns with ** won't work as expected, and some file changes that should be ignored might incorrectly trigger a new image build.

Recommendation: Explicitly enable globstar at the beginning of the script and simplify the ignore patterns for clarity and correctness.

2. Correctly Find Parent Commits in Pull Requests
In the find-existing-image step, the script uses git rev-list ... HEAD^ to look for recent commits.

The Issue: When a workflow runs for a `pull_request` event, GitHub Actions checks out a special merge commit. For this commit, `HEAD^` points to the tip of the base branch (e.g., main), not the branch the pull request came from. As a result, the script would search for existing images on the base branch's history, not the feature branch's history, and likely fail to find a reusable image. The correct reference for the feature branch head is `HEAD^2`.

Recommendation: Add logic to differentiate between `pull_request` and `push` events to ensure you're searching the correct commit history.

LMK if these should be addressed or not

PR #9030, opened 2026-01-21
fix(lockout): PHX-299: Solve lockout feature offline issue

NavarroEmilioLuis:
### Changes

We cam across an issue when updating Aspen, we had an edge case we didn't consider and this is the most straightforward way to solve it without creating added complexity

### Deploys


tcodling:
bugbot run

passcod (approved):

PR #9038, opened 2026-01-22
fix(reporting): TAMOC-381: Ensure reporting/raw users are granted appropriate access to their database schemas

rohan-bes:
### Changes

Few different ways this could have been approached:
- Just add a migration to grant permissions to the `raw` user, as the `reporting` user is managed by Maui
- Add a post migration hook
- Add as part of the initReporting logic

I went with the third option because I don't think sites that aren't using reporting actually have the database users set up. So the  migration would fail in that scenario. Could still use the post migration hook, but the initReporting logic already had the right context, and now we can validate user/permissions on startup.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/constants/src/reports.ts
```diff
@@ -72,14 +72,19 @@ export const REPORT_DATE_RANGE_LABELS = {
 
 export const REPORT_DEFAULT_DATE_RANGES_VALUES = Object.values(REPORT_DEFAULT_DATE_RANGES);
 
-export const REPORT_DB_SCHEMAS = {
+export const REPORT_DB_CONNECTIONS = {
```
rohan-bes: Bulk rename here, as strictly these aren't `schemas` but connections. And now we want to track which database schema each of these connections has access to.

packages/database/src/services/reporting.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+  if (!username || !password) {
+    log.warn(`No credentials provided for ${connectionName} reporting schema, skipping...`);
+    return null;
+  }
+
+  await validateUser(existingStore, username);
+  await grantPrivileges(existingStore, REPORT_DB_CONNECTION_SCHEMAS[connectionName], username);
+
+  return openDatabase(`reporting-${connectionName}`, overrides);
```
rohan-bes: This is the new functionality, to both check the user exists and ensure they're granted read access to all tables in their schema. Decided to do this here as there's no guarantee those users exist in the database in other contexts.

packages/facility-server/app/database/initDatabase.js
```diff
@@ -22,48 +20,6 @@ export async function initDatabase(configOverrides = {}) {
   });
 }
 
-async function initReportStore(schemaName, credentials) {
```
rohan-bes: Moved to the `@tamanu/database` package as this logic was shared with central-server

packages/database/src/services/reporting.js
```diff
@@ -0,0 +1,70 @@
...
+    },
+  );
+
+  if (!result.length > 0) {
+    throw new Error(
+      `Reporting role "${username}" does not exist, please create the role in the database first`,
+    );
+  }
+};
+
```
passcod: While `grantPrivileges` would work idempotently as written if the main application user is able to `GRANT`, I think for robustness it would be preferable to first _check_ if `raw` user can read the tables, and only-if-not attempt to grant the permission. There could be situations where the raw user exists _and_ can `SELECT` the tables, but the _tamanu_ application user _can't `GRANT`_, and so this would fail even if the situation is actually okay.

passcod (approved):
No show stoppers, but three minors

PR #9040, opened 2026-01-22
feat(datetime): TAN-2592: Timezone aware date input and forms

dannash100:
### Changes

- Adds timezone-aware datetime input handling for web forms
- Input displays in facility timezone, persists in country timezone
- Update initial values and handle some form edge cases
  - Date only fields use facility current time as they are saved as is and are not adjusted to timezone
  - Some dateTime columns have date only fields, this can be addressed in another pr
- Removed the usage of `date-fns-tz` to use Temporal api

- I think this will be followed up by an edgecase pr as I have a few things to solve in areas of app that heavily manipulate dates like appointments.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests


packages/utils/src/dateTime.ts
```diff
@@ -362,3 +373,90 @@ export const eachDayInMonth = (date: Date) =>
...
+    if (value instanceof Date) {
+      if (!isValid(value)) return null;
+      const instant = Temporal.Instant.fromEpochMilliseconds(value.getTime());
+      return toDateTimeLocalFormat(
+        instant.toZonedDateTimeISO(displayTz ?? Temporal.Now.timeZoneId()),
+      );
+    }
+
+    if (isISO9075DateString(value)) {
+      return `${value}T00:00`;
```
rohan-bes: So the idea is that if we're just being passed a date here (without any time component) we assume that we want to display specifically that date, regardless of timezone diff between facility/central? Makes sense 👍

packages/web/app/components/ProcedureModal.jsx
```diff
@@ -108,21 +112,34 @@ export const ProcedureModal = ({
...
-    const actualDateTime = getActualDateTime(data.date, data.startTime);
-    const updatedData = {
+    const dateStr = data.date.slice(0, 10);
+    const toCountry = time =>
+      time ? toDateTimeStringForPersistence(combineDateTime(dateStr, time)) : undefined;
+    const toCountryWithRollover = (time, refTime) => {
+      if (!time) return undefined;
+      // If time < reference time, it's the next day
+      const nextDay = getTime(time) < getTime(refTime);
+      return toDateTimeStringForPersistence(
```
rohan-bes: Damn this rollover behaviour is a bit tricksy, but I can see its pre-existing. Kinda odd how we manage this... I wonder if instead we rolled over the date of the `endTime`/`timeOut` as an `onChange` hook in the actual form fields so that the form data was correct rather than fixing it here.

packages/web/app/forms/AllergyForm.jsx
```diff
@@ -24,8 +23,22 @@ export const AllergyForm = ({
...
+    if (editedObject) {
+      // Currently the recordedDate is a dateTime type in the database, so we need to convert it to date type
+      // for now to avoid timezone conversion
+      return {
+        ...editedObject,
+        recordedDate: editedObject.recordedDate?.slice(0, 10),
+      };
+    }
+    return {
+      recordedDate: getFacilityCurrentDateString(),
```
rohan-bes: I can see it getting a bit confusing that for `date` inputs we pass in the facility date, whereas for `datetime` inputs we pass in country time and let the component display in facility time. Not sure of a clean way around this tho

packages/web/app/forms/DeleteTaskForm.jsx.jsx
```diff
@@ -75,7 +75,7 @@ export const DeleteTaskForm = ({ onClose, refreshTaskTable, taskIds }) => {
               required
               saveDateAsString
               component={DateTimeField}
-              max={getCurrentDateTimeString()}
+              max={formatForDateTimeInput(new Date())}
```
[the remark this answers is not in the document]
rohan-bes: Though in general I might say that it feels a bit awkward that we pass in the `max` in facility time when the underlying value is tracked in country time.

Maybe we could convert the min/max to facility time in the `DateInput` as well?

packages/web/app/forms/MarkTaskTodoForm.jsx
```diff
@@ -71,7 +72,7 @@ export const MarkTaskTodoForm = ({ onClose, refreshTaskTable, taskIds }) => {
               required
               saveDateAsString
               component={DateTimeField}
-              max={getCurrentDateTimeString()}
+              max={formatForDateTimeInput(new Date())}
```
rohan-bes: Here too

packages/web/app/forms/PatientLetterForm.jsx
```diff
@@ -84,6 +84,7 @@ const PatientLetterFormContents = ({ submitForm, onCancel, setValues }) => {
           suggester={practitionerSuggester}
           data-testid="field-ytix"
         />
+        {/* TODO: this date goes nowhere behind scenes */ }
```
rohan-bes: Thanks for marking these out 🙏

packages/web/app/views/dashboard/components/TodayAppointmentsPane.jsx
```diff
@@ -104,13 +103,21 @@ const NoDataContainer = styled.div`
 export const TodayAppointmentsPane = ({ showTasks }) => {
   const navigate = useNavigate();
   const { currentUser, facilityId } = useAuth();
+  const { getFacilityCurrentDateString, getDayBoundaries } = useDateTimeFormat();
+  
+  // Get today's date boundaries in facility timezone, converted to country timezone for query
+  const todayFacility = getFacilityCurrentDateString();
+  const boundaries = getDayBoundaries(todayFacility);
```
rohan-bes: Will this work correctly? We're not passing in the countryTz/facilityTz?

packages/web/app/views/scheduling/outpatientBookings/useOutpatientAppointmentsCalendarData.js
```diff
@@ -26,10 +27,12 @@ export const useOutpatientAppointmentsCalendarData = ({ groupBy, selectedDate })
   const { data: usersData } = usersQuery;
 
   const { filters } = useOutpatientAppointmentsContext();
+  const dateString = toDateString(selectedDate);
+  const { start, end } = getDayBoundaries(dateString);
```
rohan-bes: Same here, maybe I'm missing something?

rohan-bes (changes requested):
Awesome work! This must've been really painstaking. But great that we're gonna have better timezone management in the future, and the switch to Temporal API is really neat.

Had questions and comments, but it's possible I'm misunderstanding things so feel free to let me know.

packages/utils/src/dateTime.ts
dannash100: Yes thats right!

packages/web/app/components/ProcedureModal.jsx
dannash100: Yeah good call Ill look into that

packages/shared/src/utils/handoverNotes/HandoverHeaderSection.jsx
```diff
@@ -2,11 +2,11 @@ import React from 'react';
 import { Box, Logo } from '../patientCertificates/Layout';
 import { H1, H3 } from '../patientCertificates/Typography';
 import { Divider } from './Divider';
-import { useDateTimeFormat } from '../pdf/withDateTimeContext';
+import { useDateTime } from '../pdf/withDateTimeContext';
```
rohan-bes: Like this rename 👍

rohan-bes (approved):
Thanks for answering my q's from last time, I think this is looking good 👍

PR #9082, opened 2026-01-29
fix(translations): NOTAM-551: 2.48 fix: Translated string admin panel field name was wrong

dannash100:
### Changes

- Incorrect data being sent: The AccessorField was adding a `labTests.` prefix to field names, causing the form data structure to be wrong. 
- English translations not showing: Since Formik was looking for field values at the wrong path (e.g., `labTests.['welcome.title'].en` instead of `['welcome.title'].en`), it couldn't find the values from initialValues, so the fields appeared empty.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/views/administration/translation/TranslationForm.jsx
+++ b/packages/web/app/views/administration/translation/TranslationForm.jsx
@@ -14,10 +14,9 @@ import {
   LANGUAGE_NAME_STRING_ID,
 } from '@tamanu/constants';
 import { useApi } from '../../../api';
-import { TextField, Form, Button, TranslatedText } from '@tamanu/ui-components';
+import { TextField, Form, Button, TranslatedText, Field } from '@tamanu/ui-components';
 import { Colors } from '../../../constants/styles';
 import { SearchInput, TableFormFields } from '../../../components';
-import { AccessorField } from '../../patients/components/AccessorField';
 import { LoadingIndicator } from '../../../components/LoadingIndicator';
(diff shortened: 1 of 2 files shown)


packages/web/app/views/administration/translation/TranslationForm.jsx
```diff
@@ -153,13 +152,11 @@ const useTranslationMutation = () => {
 };
 
 const TranslationField = ({ stringId, code }) => (
-  // This id format is necessary to avoid formik nesting at . delimiters
-  <AccessorField
-    id={`['${stringId}']`}
-    name={code}
+  <Field
+    name={`['${stringId}']['${code}']`}
```
NavarroEmilioLuis: Both string ID and code wrapper in brackets, it won't look for `['welcome.title']` either? Maybe it was meant to be:


```suggestion
    name={`['${stringId}.${code}']`}
```
?

NavarroEmilioLuis (commented):
It seems a bit weird that we want to modify the form state? Why did we need to change the field input name for it to find the proper translation?

packages/web/app/views/administration/translation/TranslationForm.jsx
dannash100: Suggestion doesn't work in relation to initial values

NavarroEmilioLuis (approved):
Thanks for clearing that out!

PR #9109, opened 2026-02-04
feat(settings): SAV-1165: settings secrets

passcod:
### Changes

Adds the ability to store secrets in config files, in local system facts, and in settings. Secrets are encrypted using the Web Crypto API, with a symmetric key. For settings, there's an additional layer so that all servers can share a key (as the secrets are encrypted once in the DB).

In config, it looks like this:

```json5
{
  sync: {
    email: "admin@tamanu.io",
    password: "S1:xK3mN9pQrStUvW:a7Bz4cDe8fGh2iJk6lMn0oPq5rSt9uVw3xYz1AbC4dEf==",
  },
  crypto: {
    keyFile: "/secrets/configKey",
    settingsPsk: "S1:f4N7yyOcqyhf==:aB3cD4eF5gH6iJ7kL8mN9oP0qR1sT2uV3wX4yZ5aB6cD7e4sT5uZ8==",
  },
}
```

In settings and local system facts, it looks similar:

```sql
INSERT INTO settings (key, value) VALUES
('superDuperSecret', 'S1:kwzzFf7NTtuh=:NL4CfIj3aHh+AgmwXBAffMThsvFK5trjcQtFwqCpPpR
CZiHKBljEo46KbXgusFEZQ==');
```

The unique thing with settings is the `settingsPsk` config secret. That's a shared key (PSK = Pre-Shared Key) common between all servers that can read secrets from settings. It's encrypted with the per-server key when it's stored in config. This key is required to be 32 bytes, encoded in hex, so it's sufficiently strong.

To start with, the per-server key will be just a file stored separately from the config, but later we can improve that security if we need, for example by using a TPM on metal or in AWS on Nitro EC2 instances.

The format of encrypted secrets is `S1` (just a version and a visual identifier that it's a secret), the IV, and the ciphertext. The algorithm used is AES-GCM.

In the admin panel UI, secret values can't be read, only written.

As an example of a migration to secrets, we have the DHIS2 credentials, which are currently in config in plain text. This PR includes a change that will first look in settings, then for an encrypted value in config, and fall back to the plain value in config.

image

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### 🦸 Review Hero

- [ ] **Run Review Hero** 
- [ ] **Auto-fix review suggestions** 
- [ ] **Auto-fix CI failures**


packages/central-server/app/subCommands/configSecret.js
```diff
@@ -0,0 +1,31 @@
...
+      process.exit(1);
+    }
+  });
+
+export const configSecretCommand = new Command('configSecret').description(
+  'Manage encrypted configuration secrets',
+);
+
+configSecretCommand.addCommand(initCommand);
+configSecretCommand.addCommand(encryptCommand);
```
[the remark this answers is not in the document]
passcod: We can't, because the commander module isn't in shared. At this point, we've abstracted all we could.

passcod:
Note that the deploy is up and shows the UI, but Kubernetes-side is not yet configured to have a config secret and so on, so it won't "actually work"

packages/central-server/app/tasks/DHIS2IntegrationProcessor.js
```diff
@@ -72,9 +73,45 @@ export class DHIS2IntegrationProcessor extends ScheduledTask {
...
+      username = await this.context.settings.getSecret('integrations.dhis2.username');
+    } catch {
+      // Settings secret not available, try config secret
+      try {
+        username = await getConfigSecret('integrations.dhis2.username');
+      } catch {
+        // Config secret not available, fall back to plain config
+        username = config.integrations?.dhis2?.username || null;
+      }
+    }
```
rohan-bes: In this specific instance might recommend just the `password` being encrypted. Nice to have the username in plaintext so we know it's set to the right user.

packages/web/app/views/administration/settings/components/SettingInput.jsx
```diff
@@ -151,11 +180,68 @@ export const SettingInput = ({
...
+            InputProps={{
+              endAdornment: (
+                <InputAdornment position="end">
+                  <IconButton
+                    aria-label="toggle secret visibility"
+                    onClick={toggleShowSecret}
+                    edge="end"
+                    size="small"
+                  >
+                    {showSecretValue ? <VisibilityOff /> : <Visibility />}
```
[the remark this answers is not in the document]
rohan-bes: I think I'm okay with not being able to view the secret setting values. Feels like it'd be hard to allow users to view in plaintext in a way that's still secure.

packages/central-server/app/tasks/DHIS2IntegrationProcessor.js
```diff
@@ -72,9 +73,45 @@ export class DHIS2IntegrationProcessor extends ScheduledTask {
...
+
+    try {
+      password = await this.context.settings.getSecret('integrations.dhis2.password');
+    } catch {
+      try {
+        password = await getConfigSecret('integrations.dhis2.password');
+      } catch {
+        password = config.integrations?.dhis2?.password || null;
+      }
```
rohan-bes: Thoughts on making this fallback logic generic in the `getConfigSecret` function? I could see it being convenient if it turns out that some deployments use secrets and others don't.

Just a suggestion tho

packages/database/src/models/LocalSystemFact.ts
```diff
@@ -142,4 +148,28 @@ export class LocalSystemFact extends Model {
...
+
+    const keyFilePath = getConfigKeyFilePath();
+    const keyBuffer = await readKeyFile(keyFilePath);
+    return decryptSecret(keyBuffer, encryptedValue);
+  }
+
+  /**
+   * Sets an encrypted secret in the local system facts table.
+   */
+  static async setSecret(key: FactName, value: string): Promise<void> {
```
rohan-bes: We're not setting any secrets in the local system facts table as yet yeah? This is just a convenience thing?

rohan-bes (changes requested):
Looks great @passcod! Awesome change.

Mostly questions/suggestions, just requesting changes around:
1. Only making the password secret in DHIS2 integration settings/config
2. The masking front-end logic, and whether we display the placeholder if a secret setting is empty

Also wasn't entirely sure why we need the patchReadSettings logic, but do feel its a a bit unexpected and might catch us out.

packages/database/src/models/LocalSystemFact.ts
passcod: We actually do, the device private key is generated and stored in plain text here. But it's not moved to encrypted values in this PR, no.

passcod:
Closing for now, need to rethink the architecture a bit. The concept is sound but the implementation is brittle.

dannash100:
bugbot run

rohan-bes (approved):
Thanks for addressing those comments :)

PR #9112, opened 2026-02-04
fix(invoicing): SAV-1058: Remove village from price list title

chris-bes:
### Changes
Screenshot 2026-02-05 at 9 14 29 am

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/shared/src/utils/patientCertificates/InvoiceRecordPrintout.jsx
+++ b/packages/shared/src/utils/patientCertificates/InvoiceRecordPrintout.jsx
@@ -263,7 +263,10 @@ const HeaderCell = ({ children, style }) => (
   </View>
 );
 
-const SectionSpacing = () => <View style={{ paddingBottom: '10px' }} />;
+const SectionSpacing = ({ minPresenceAhead = 70 }) => (
+  // Between sections, ensure we have enough space for the fixed header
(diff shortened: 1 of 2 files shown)


tcaiger (approved):
Thanks for picking this up. It was my mistake - I mis-read the design.
Just looks like there is a const to remove too.

tcaiger (approved):
Cool yep I reviewed the second batch of changes and all looks good

packages/shared/src/utils/patientCertificates/InvoiceRecordPrintout.jsx
```diff
@@ -313,13 +305,14 @@ const COLUMNS = {
       title: 'Ordered by',
       accessor: ({ orderedByUser }) => orderedByUser?.displayName,
       style: { width: '24%' },
+      CellComponent: OrderedByCellWithAdjustments,
```
chris-bes: Previously for the item adjustments, my implementation was treat them as separate sub rows, but that introduced a problem that if the details value is multi lines, it's going to create more gap between the item adjustments like below

Image


So to fix this, I remove the subrows functionality, and just include the values in the custom CellComponent. Below is the result

Image

PR #9113, opened 2026-02-04
fix: Release v2.48: Medication tweaks

tcodling:
### Changes

Overview
Standardises the “Last sent” timestamp display in medication tables by using shared formatTime/formatShortest helpers (and removing ad-hoc date-fns parsing/formatting) in both MedicationTable and PharmacyOrderMedicationTable.

Updates encounter action rendering so that when an encounter has endDate, the Encounter/Discharge summary buttons only appear if discharge data is available via useEncounterDischargeQuery

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

---

> [!NOTE]
> **Medium Risk**
> Mostly UI-level formatting and conditional rendering, but it changes when discharge/summary actions appear and adds a new discharge fetch that could affect workflows if the query behavior differs from expectations.
> 
> **Overview**
> Standardises the “Last sent” timestamp rendering in `MedicationTable` and `PharmacyOrderMedicationTable` by using shared `formatShortest`/`formatTime` helpers instead of local `Date`/`date-fns` formatting.
> 
> Updates `EncounterActions` so that when an encounter has an `endDate`, the Encounter/Discharge summary buttons only render if a discharge record is available via `useEncounterDischargeQuery` (avoiding showing discharge actions for ended-but-not-discharged encounters).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 81dc749e107d4e94d05e9b1c66da7f2a2a6c3288. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/web/app/views/patients/components/EncounterActions.jsx
+++ b/packages/web/app/views/patients/components/EncounterActions.jsx
@@ -80,7 +80,7 @@ const EncounterActionDropdown = ({ encounter, setOpenModal, setNewEncounterType 
 
   if (encounter.endDate) {
     return (
-      discharge ? (
+      discharge && (
         <ActionsContainer data-testid="actionscontainer-w92z">
           <StyledButton
             size="small"
@@ -108,7 +108,7 @@ const EncounterActionDropdown = ({ encounter, setOpenModal, setNewEncounterType 
             />
           </StyledButton>
         </ActionsContainer>
-      ) : <></>)
+      ))
   }
 
   const progression = {
--- a/packages/web/app/views/patients/components/EncounterActions.jsx
+++ b/packages/web/app/views/patients/components/EncounterActions.jsx
@@ -79,6 +79,10 @@ const EncounterActionDropdown = ({ encounter, setOpenModal, setNewEncounterType 
   const onChangeDiet = () => setOpenModal(ENCOUNTER_MODALS.CHANGE_DIET);
 
   if (encounter.endDate) {
+    // Ideally we would have a dedicated encounter type for discharged encounters and filter
+    // at the same level as the other encounter types. Because discharge uses clinic data we
+    // need this extra check here to only show encounter/discharge summary actions when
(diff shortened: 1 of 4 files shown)


packages/web/app/views/patients/components/EncounterActions.jsx
```diff
@@ -78,34 +80,35 @@ const EncounterActionDropdown = ({ encounter, setOpenModal, setNewEncounterType
...
-        >
-          <TranslatedText
-            stringId="patient.encounter.action.dischargeSummary"
-            fallback="Discharge summary"
-            data-testid="translatedtext-0hzq"
-          />
-        </StyledButton>
-      </ActionsContainer>
-    );
+      discharge && (
```
tcodling: This is a bit hackier than id like

We already have “instant close” system encounters (e.g. vaccine, survey) that are handled above this component: the parent doesn’t render EncounterActions at all for those encounter types, so we never hit this code.

Medication dispensing behaves like another instant-close system encounter and doesn't have a discharge, but it reuses the built-in clinic encounter type. That means we can’t exclude it the same way at the parent level, so EncounterActions still gets rendered for closed medication encounters.

The extra check discharge ? (...) : <></> is there so that for closed encounters without discharge data (e.g. medication dispensing), we render nothing instead of the “Encounter summary” / “Discharge summary” buttons. **Both summaries assume discharge data exists**, so showing those actions without it would be wrong.

Long term, the cleaner approach would be a dedicated “Medication dispensing” encounter type and filtering it like vaccine/survey, but that would touch a lot of logic, so it’s out of scope for this late-release fix.

dannash100: Mabye leave a comment explaining I rekon :)

dannash100 (approved):
Looks great!

PR #9172, opened 2026-02-16
fix(sync): COOL-36: Prevent duplicate patient_ongoing_prescriptions when an encounter is discharged on facility and central

rohan-bes:
### Changes

This is a common enough bug, as if sync is failing for some reason encounters often get discharged both on central by the automatic discharger, and on the facility manually. If discharge prescriptions were created, duplicates end up in the system which ends up breaking the logic to build the sync lookup table.

There's a few ways we could resolve this, but I like making the ids deterministic as its an established pattern for merging sync records, and also ensures that we now have a constraint to prevent multiple pops for the same patient/prescription. 

This does however require changing the id of live records. I've added logic to fix up the audit logs, and the sync lookup table, but still a little wary of this.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/migrations/1770250000001-makePatientOngoingPrescriptionIdDeterministic.ts
+++ b/packages/database/src/migrations/1770250000001-makePatientOngoingPrescriptionIdDeterministic.ts
@@ -1,0 +1,73 @@
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// Deterministic id from (patient_id, prescription_id), same pattern as patient_facilities.
+// REPLACE avoids collisions if ';' appears in ids. Updates logs.changes and sync_lookup so audit and sync stay consistent.
+
+export async function up(query: QueryInterface): Promise<void> {
+  const duplicates = await query.sequelize.query<{
+    patient_id: string;
+    prescription_id: string;
+    count: string;
+  }>(
+    `
+    SELECT patient_id, prescription_id, COUNT(*)::text as count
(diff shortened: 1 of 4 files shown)


packages/database/src/migrations/1770250000001-makePatientOngoingPrescriptionIdDeterministic.ts
```diff
@@ -0,0 +1,66 @@
...
+    `,
+    { type: QueryTypes.SELECT },
+  );
+
+  if (duplicates.length > 0) {
+    throw new Error(
+      `Cannot run migration: found ${duplicates.length} duplicate (patient_id, prescription_id) pair(s) in patient_ongoing_prescriptions. ` +
+        `Clean up duplicates manually before running this migration. Duplicates: ${JSON.stringify(duplicates)}`,
+    );
+  }
```
NavarroEmilioLuis: I take it these need to be sorted out case by case? Is there any strategy in place that takes care of that?

We had issues with program registries trying to pull this off, so the recommendation is to add pre deployment checks so we don't hit a wall when it happens

NavarroEmilioLuis (changes requested):
The down migration seems to need some care, otherwise looks fine.

I would only really say that it's a costly operation having to update so many places, give dev ops a really early heads up!

NavarroEmilioLuis (approved):

PR #9179, opened 2026-02-17
fix(db): COOL-48: Encounter table non nullable patient ID column

NavarroEmilioLuis:
### Changes

Well

### Deploys

### DIFF
--- a/packages/database/src/migrations/1771277622909-updateEncountersTableSetPatientIdNotNull.ts
+++ b/packages/database/src/migrations/1771277622909-updateEncountersTableSetPatientIdNotNull.ts
@@ -1,0 +1,26 @@
+import { DataTypes, QueryInterface } from 'sequelize';
+
+// Migration 2 of 2: Set patient_id NOT NULL
+// Backfill of null patient_id is in 1771277622908-backfillEncountersPatientIdToTestPatient.ts
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.changeColumn('encounters', 'patient_id', {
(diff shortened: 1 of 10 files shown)


NavarroEmilioLuis:
bugbot run

packages/database/src/migrations/1771277622909-updateEncountersTableSetPatientIdNotNull.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.changeColumn('encounters', 'patient_id', {
+    type: DataTypes.STRING,
+    allowNull: false,
+    references: {
+      model: 'patients',
+      key: 'id',
+    },
+  });
```
rohan-bes: I'd probably have combined these into a single migration. On the very off chance that some other process creates an encounter in between the previous migration and this one. But no biggie its probably fine 👍

rohan-bes (approved):
Nice stuff. Might be worth trying to get a read on how many sites don't have the test patient set up?

packages/database/src/migrations/1771277622909-updateEncountersTableSetPatientIdNotNull.ts
NavarroEmilioLuis: AFAIK we need to keep schema changes and data changes separate: https://beyondessential.slack.com/archives/GE9NHB95F/p1768427651912359

PR #9195, opened 2026-02-19
feat(integrations): PHX-297: Send transactions to open mSupply API

NavarroEmilioLuis:
### Changes

Follows a similar pattern to the DHIS integration. More information on Linear cause there are lots of details here. Note that one third of the LOC is the test file, the other third is probably boilerplate (migration, model, config, dbt) and the rest is the actual task file to send the transactions.

### Deploys


packages/facility-server/app/tasks/mSupplyMedIntegrationProcessor.js
```diff
@@ -0,0 +1,328 @@
...
+    this.context = context;
+    this.models = context.models;
+    this.serverFacilityIds = selectFacilityIds(config);
+    this.authHeader = null;
+  }
+
+  async createLog(values) {
+    const { items, ...rest } = values;
+    // Ensure items is an array if it exists
+    const payload = { ...rest, ...(Array.isArray(items) && { items }) };
```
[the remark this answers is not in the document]
NavarroEmilioLuis: @gemini-code-assist-user but what would happen if `items` is something incompatible with JSONB and creating the log fails? I think this is fine, because the API defines that it returns only an array

packages/facility-server/app/tasks/mSupplyMedIntegrationProcessor.js
```diff
@@ -0,0 +1,328 @@
...
+        pluginCode: "${INTEGRATION_PLUGIN_CODE}"
+        storeId: "${storeId}"
+        input: $input
+      )
+    }
+  `;
+}
+
+// Designed to post dispensed medications from pharmacy to an Open mSupply instance
+export class mSupplyMedIntegrationProcessor extends ScheduledTask {
```
rohan-bes: I know you're following a bit of a convention here with the `DHIS2IntegrationProcessor` but tbh I don't love that naming convention as its not very clear about what the task actually does. Thoughts on something like `MSupplyMedicationStockUpdater`?

packages/facility-server/app/tasks/mSupplyMedIntegrationProcessor.js
```diff
@@ -0,0 +1,328 @@
...
+            ],
+          },
+        ],
+      },
+    ];
+
+    // Process everything if there is no successful push
+    if (!lastSuccessfulPush) {
+      return { include };
+    }
```
rohan-bes: So whenever we turn on the integration we push the full dispense history for Tamanu? And if we turn it off and back on again, we push all dispenses that happened since the last push?

All good if this has been discussed, but seems like it might be a little unexpected. Maybe I would have thought we would only care about pushing dispenses since the integration had been turned on, and that we wouldn't want to push historical data (acknowledge that determining what's historical might not be easy).

rohan-bes (approved):
Have a few comments and optional requests for rename/refactor, but nothing blocking. Looks solid, and with all things integrations there's usually a bit of tweaking once you start actually integrating so best to get to that point sooner rather than later :)

Harnesh787 (approved):

PR #9217, opened 2026-02-23
fix(SAST): COOL-47: js/polynomial-redos

NavarroEmilioLuis:
### Changes

This should get rid of:

- https://github.com/beyondessential/tamanu/security/code-scanning/50
- https://github.com/beyondessential/tamanu/security/code-scanning/49


tcodling (changes requested):
I think we should remove parseFullLocationName completely since it isnt used 🙏

tcodling (approved):

PR #9246, opened 2026-02-26
tweak(dateTime): Location bookings draggable outside of office hours

dannash100:
### Changes
Missed a spot!
- Issue was that you could move the booking to a timeslot outside working hours when dragging
- It might become apparent that we are doing these kind of actions alot and they can become there own context utils, but for now just inline function to convert to facility date from string. There are only a few other feels that would use that same logic anyway

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


tcodling (approved):
lgtm. Only thing is it looks like maybe cursor removed some comments that should probably be kept? Idk this file is pretty hectic so the more comments the better in my opinion

PR #9374, opened 2026-03-17
db: no-issue: backport removeDeviceFkFromUserLoginAttempts migration to 2.45

passcod:
### Changes

Backport the `1768858451324-removeDeviceFkFromUserLoginAttempts` migration to the `release/2.45` branch.

This migration drops the foreign key constraint `user_login_attempts_device_id_fkey` on the `user_login_attempts` table and replaces it with a plain index on `device_id`. The `user_login_attempts` table has existed since v2.45 (migration `1757268197134`), but this FK removal was only introduced in v2.49. Backporting it here ensures older 2.45.x deployments also benefit from decoupling the `user_login_attempts` table from the `devices` table at the constraint level.

The migration uses `DROP CONSTRAINT IF EXISTS` and `CREATE INDEX IF NOT EXISTS`, so it is safe to run even if the constraint was already removed or the index already exists.

### Deploys

### Tests

- [ ] **Run E2E Tests** 

### Review Hero

- [x] **Run Review Hero** 

### DIFF
--- a/packages/database/src/migrations/1768858451324-removeDeviceFkFromUserLoginAttempts.ts
+++ b/packages/database/src/migrations/1768858451324-removeDeviceFkFromUserLoginAttempts.ts
@@ -1,0 +1,31 @@
+import { QueryInterface } from 'sequelize';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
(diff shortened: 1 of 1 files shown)


NavarroEmilioLuis (approved):
I mean, don't love the warnings but that should be okay

PR #9521, opened 2026-04-07
perf(mobile): improve survey form performance on low-end devices

tcodling:
### Changes
Fixes mobile survey form performance issues reported by IRD on low-end Android devices (Samsung Galaxy A04s/A12). Addresses slow dropdown interaction, slow form submission, and spurious auto-logout.

### Decouple `FormField` error display from `validateOnChange`
- `FormField.tsx`'s `showError` previously branched on `validateOnChange` from Formik context; now it just checks submit state (`SUBMIT_ATTEMPTED_STATUS` or `submitCount > 0`) — mathematically equivalent to the old behaviour on the survey form but independent of Formik config
- `validateOnChange` is kept enabled on the survey form: with the stable-schema change below the per-keystroke validation cost is small, and keeping it preserves pre-existing error-display timing

### Stabilise validation schema rebuilds
- `formValidationSchema` was rebuilding on every `formValues` change even when field visibility hadn't changed
- Now only rebuilds when the set of visible component IDs actually changes

### Cache visibility check internals (`fields.ts`)
- Cache `JSON.parse` results for visibility criteria (called per-component, per-render)
- Use `Map` lookup instead of `Array.find()` for component-by-code resolution (O(1) vs O(n))

### Eliminate re-renders during submission
- `setNote` (a `useState` setter) was called ~30+ times during `SurveyResponse.submit`, triggering a React re-render per answer while the note was never displayed
- Removed — `submit()` defaults to a no-op

### Remove duplicate encounter query (`FormFields.tsx`)
- `FormFields` had its own `useBackendEffect` fetching the current encounter, duplicating the same query already in `SurveyForm` — now passed as a prop

### Refactor `DetectIdleLayer` to use refs
- Old implementation used `useState` for idle tracking, causing the entire app tree to re-render every minute and on every background/foreground transition
- Replaced with refs — zero re-renders from idle tracking

### Map lookup in `SurveyResponse.submit`
- Replaced O(n²) `components.find()` loop with pre-built `Map` for answer→component matching

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 

</details>

### Tests

- [ ] **Run E2E Tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

---

> [!NOTE]
> **Medium Risk**
> Touches survey submission/value-calculation flow and idle sign-out timing; errors here could impact saved responses or unexpectedly sign users out.
> 
> **Overview**
> Improves mobile survey performance by avoiding repeated linear lookups and excessive Formik updates: survey submission now indexes components by `dataElement.code`, and calculated fields are batch-applied via `setValues` only when they actually change.
> 
> Simplifies data fetching/props in the survey form by lifting encounter loading to `SurveyForm` and passing only `encounterType` into `FormFields` for mandatory/visibility checks.
> 
> Refactors `DetectIdleLayer` to track idle time via refs + timestamp deltas (instead of state increments), ensuring background/foreground time is handled cleanly and preventing stale `signOutClient` references during auto-logout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 66086173d58da58c5a663d9ec544da4cbe284d4f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tcodling:
@cursor push 7167e241cc

tcodling:
@cursor push 4077bbce5f

packages/mobile/App/ui/components/Forms/SurveyForm/index.tsx
```diff
@@ -80,14 +78,19 @@ export const SurveyForm = ({
...
     () =>
-      getFormSchema(
-        components.filter(c => checkVisibilityCriteria(c, components, formValues)),
-        { encounterType: encounter?.encounterType },
-        getTranslation,
-      ),
-    [encounter?.encounterType, components, formValues, getTranslation],
+      getFormSchema(visibleComponents, { encounterType: encounter?.encounterType }, getTranslation),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [visibleComponentKey, encounter?.encounterType, getTranslation],
```
jaskfla: Some heebie-jeebies here. Nothing wrong with using `visibleComponents` as dependency here, seeing as you’ve memo-ised it.

If anything, `visibleComponentKey` is extra work that could be entirely omitted

packages/mobile/App/ui/components/Forms/SurveyForm/FormFields.tsx
```diff
@@ -118,15 +108,6 @@ export const FormFields = ({
...
-  }
-
-  if (isEncounterLoading) {
-    return <LoadingScreen />;
-  }
-
-  const { encounter } = encounterResult || {};
   const maxIndex = components
     .map((x) => x.screenIndex)
     .reduce((max, current) => Math.max(max, current), 0);
```
jaskfla: I dunno how big `components` gets, but here’s another opportunity to optimise. The `.map()` could be folded into the reducer:

```tsx
const maxIndex = components.reduce((max, current) => Math.max(max, current.screenIndex))
```

Omitting the initial value `0` also means that the reducer starts with the 0th rather than the 1st.

Probably the wrong call here, though. In fact, I would go the _other_ way and make this more readable at the **cost** of performance:

```tsx
const maxIndex = Math.max(...components.map(c => c.screenIndex));
```

packages/mobile/App/ui/components/Forms/SurveyForm/index.tsx
```diff
@@ -120,35 +123,49 @@ export const SurveyForm = ({
...
+          const changedCalculatedValues = Object.fromEntries(
+            Object.entries(calculatedValues).filter(([key, value]) => values[key] !== value),
+          );
 
-          // write values that have changed back into answers
-          Object.entries(calculatedValues)
-            .filter(([key, value]) => values[key] !== value)
-            .map(([key, value]) => setFieldValue(key, value, false));
+          if (Object.keys(changedCalculatedValues).length > 0) {
+            // Avoid one Formik state update per calculated field.
```
jaskfla: This will maybe save like 0.000001 ms, so obvs optional:

```tsx
const changes = Object.entries().filter(); // Don’t create Object.fromEntries()...

if (changes.length > 0) {
  // ...until we need it
  const changedCalculatedValues = Object.fromEntries(changes);
}
```

packages/mobile/App/ui/components/Forms/SurveyForm/index.tsx
```diff
@@ -120,35 +123,49 @@ export const SurveyForm = ({
...
             ...calculatedValues,
-          });
-        }, [values, setFieldValue]);
+          };
+
+          setFormValues(previousValues =>
+            Object.keys(previousValues).length === Object.keys(nextFormValues).length &&
+            Object.entries(nextFormValues).every(([key, value]) => previousValues[key] === value)
+              ? previousValues
+              : nextFormValues,
```
jaskfla: This is redundant:

```tsx
setFormValues(isEqual(prev, next) ? prev : next);
```

Instead:

```tsx
if (!isEqual(prev, next)) setFormValues(next)
```

packages/mobile/App/ui/helpers/fields.ts
```diff
@@ -154,15 +154,30 @@ function fallbackParseVisibilityCriteria(
   return compareData(comparisonDataType, expectedTrimmed, givenAnswer);
 }
 
+const parsedCriteriaCache = new Map<string, any>();
+const componentsByCodeCache = new WeakMap<ISurveyScreenComponent[], Map<string, ISurveyScreenComponent>>();
```
jaskfla, line 158: First time I’ve seen a real-world usage of `WeakMap`!!

packages/mobile/App/ui/helpers/fields.ts
```diff
@@ -154,15 +154,30 @@ function fallbackParseVisibilityCriteria(
...
+
+function getComponentsByCode(allComponents: ISurveyScreenComponent[]): Map<string, ISurveyScreenComponent> {
```
jaskfla, line 160: Just double checking we’re confident that when we call this with the “same” array, they will satisfy `===`?

`Map` keys aren’t smart enough to do deep-equals logic

packages/mobile/App/ui/helpers/fields.ts
```diff
@@ -154,15 +154,30 @@ function fallbackParseVisibilityCriteria(
...
   allComponents: ISurveyScreenComponent[],
   values: any,
 ) {
-  // nothing set - show by default
   if (!criteria) return true;
 
-  const criteriaObject = JSON.parse(criteria);
+  let criteriaObject = parsedCriteriaCache.get(criteria);
+  if (criteriaObject === undefined) {
+    criteriaObject = JSON.parse(criteria);
```
jaskfla: We’re comfortable that this will never throw `SyntaxError`?

packages/mobile/App/ui/navigation/stacks/DetectIdleLayer.tsx
```diff
@@ -19,71 +19,58 @@ const ONE_MINUTE = 1000 * 60;
...
-      // App has moved to the foreground
-      if (screenOffTime) {
-        const timeDiff = Date.now() - screenOffTime;
-        const newIdle = idle + timeDiff;
-        setIdle(newIdle);
-        setScreenOffTime(null);
-        if (newIdle >= UI_EXPIRY_TIME) {
-          handleIdleLogout();
+    const handleStateChange = (nextAppState: AppStateStatus): void => {
+      if (appStateRef.current === 'active' && nextAppState.match(/inactive|background/)) {
```
jaskfla: Probably out-of-scope, but I suspect this `.match()` would benefit from `^$`

jaskfla (approved):
Mostly just one concern with a dependency array, and a handful of optional nitpicks. Otherwise looks good!

packages/mobile/App/ui/navigation/stacks/DetectIdleLayer.tsx
dannash100: Done in 43f7bf9c — anchored both sides of the regex.

PR #9608, opened 2026-04-20
fix(sync): no-issue: include prescriptions in sync when patient_ongoing_prescription tick changes

passcod:
### Changes

A prescription could arrive at a facility after its patient_ongoing_prescription if the facility's sync cursor was already past the prescription's original tick when the POP was created. Mirror the Encounter/lab-request pattern by ORing the POP's updated_at_sync_tick into both the patient sync filter and the lookup table where clause, and rebuild the lookup table to fix existing entries.

Incident thread: https://beyondessential.slack.com/archives/C022N48Q3JQ/p1776401521213449

Manifesting as:

```
SequelizeForeignKeyConstraintError
original=error: insert or update on table "patient_ongoing_prescriptions"
violates foreign key constraint "patient_ongoing_prescriptions_prescription_id_fkey"
sql=INSERT INTO "patient_ongoing_prescriptions"
```

This is essentially a template for the upcoming sync stability project ^^

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 

</details>

### Tests

- [ ] **Run E2E Tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/migrations/1776657862375-RebuildLookupTableForPrescriptionOngoingSync.ts
+++ b/packages/database/src/migrations/1776657862375-RebuildLookupTableForPrescriptionOngoingSync.ts
@@ -1,0 +1,11 @@
+import { QueryInterface } from 'sequelize';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
+    SELECT flag_lookup_model_to_rebuild('prescriptions');
+  `);
+}
+
+export async function down(): Promise<void> {
+  // No reverse migration
+}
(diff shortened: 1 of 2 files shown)


edmofro (approved):
Looks good. Only thing I want to check is that there are two ways we do this in Tamanu:
1. Include other tables in the sync cursor filter (what you've done here)
2. Bump related tables as a side effect (I can't remember where)
Can you check with AI on the second, and think about whether it would be more appropriate here? I think the option you've gone with is good, because we're already joining anyway...but just have a quick check. That aside, approved.

passcod:
The AI did initially go for a "bump the updated_at_sync_tick" thing, but then we couldn't find a clear pattern of this existing elsewhere in the codebase for this, so revised into the sync filter approach. I've asked again and found the `bumpSyncTickForRepull` helper, but the way it's used in the codebase already seems to be for a somewhat different pattern of needing to "push out" (repull) updated linked records after a pull, whereas this one is more similar to the "Encounter including lab-request ticks in its sync filter" scenario. The fact that there's an existing join pushes it into this filter solution I think, as it's no more expensive this way.

PR #9867, opened 2026-05-21
fix(fhir): no-issue: backstop fhir queue processing from heartbeat

passcod:
### Changes

NOTIFY-triggered processing assumes the listener never misses a message. That's a fine assumption most of the time, but it isn't bulletproof: any NOTIFYs that fire during a pg-notify reconnect window are dropped on the floor, and historically a stuck `LISTEN jobs` could stall the worker indefinitely (which is the bug that motivated this stack).

On each heartbeat tick, we also call `processQueueNow` for every registered topic. The topic processor short-circuits when there's no backlog so this is cheap, and it bounds the worst-case latency for picking up a queued job to one heartbeat interval, regardless of how many NOTIFYs were missed in between.

Stacked on top of #9866 (pg-notify switch); only the heartbeat callback is touched here.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/shared/src/tasks/FhirQueueManager.js
```diff
@@ -62,6 +62,13 @@ export class FhirQueueManager {
           totalJobs: this.worker.metadata.totalJobs || 0,
         });
         await this.worker.heartbeat();
+
+        // Backstop in case a NOTIFY was missed (e.g. during pg-notify
+        // reconnect): poke every registered topic so the next heartbeat
+        // tick is the worst-case latency for picking up a queued job.
+        for (const topic of this.queueProcessors.keys()) {
+          this.processQueueNow(topic);
+        }
```
rohan-bes, line 71: Yeah I like this change!

rohan-bes (approved):

PR #9893, opened 2026-05-25
fix(web): TAM-6783: Improve back button behaviour

rohan-bes:
### Changes

Improves back button behaviour in the web app by persisting the active tab in the URL search params (`?tab=`) so navigating back from a child route (e.g. lab request, imaging request, invoice) returns the user to the same tab they were on. Introduces a `useSyncedTabSearchParam` hook that keeps the selected tab synced with the URL using replace-navigation, replacing the previous approach of manually mutating `window.history`. Removes the now-redundant `setNavigateBackTab` helper.

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/web/app/hooks/useGoBack.js
+++ b/packages/web/app/hooks/useGoBack.js
@@ -1,0 +1,22 @@
+import { useNavigate, useLocation } from 'react-router';
+
+/**
+ * Returns a callback that navigates back within the app.
+ *
+ * React Router sets location.key to 'default' for the initial history entry
+ * (i.e. the page the app first loaded on). Calling navigate(-1) from there
+ * would exit the app into whatever the browser had loaded before — potentially
+ * an external site. When we're on the initial entry we navigate to `fallback`
+ * instead.
+ */
+export const useGoBack = (fallback = '/') => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  return () => {
+    if (location.key !== 'default') {
+      navigate(-1);
+    } else {
+      navigate(fallback, { replace: true });
+    }
+  };
+};
--- a/packages/web/app/utils/useSyncedTabSearchParam.js
+++ b/packages/web/app/utils/useSyncedTabSearchParam.js
@@ -1,0 +1,60 @@
+import { useCallback, useEffect, useMemo } from 'react';
+import { useSearchParams } from 'react-router';
+
+const TAB_SEARCH_PARAM = 'tab';
+
+/**
+ * Keeps the selected UI tab in the `tab` URL search param (replace navigation).
+ * Ensures browser Back from child routes returns to the parent with the same tab selected.
+ *
+ * @param {string[]} visibleTabKeys — tab keys currently shown (e.g. from permissions / settings)
+ * @param {string | undefined} fallbackTabKey — default when URL has no tab or an unknown tab
+ * @returns {{ currentTab: string | undefined, onTabSelect: (tabKey: string) => void }}
+ */
+export function useSyncedTabSearchParam(visibleTabKeys, fallbackTabKey) {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const tabFromUrl = searchParams.get(TAB_SEARCH_PARAM);
+
+  const currentTab = useMemo(() => {
+    if (!fallbackTabKey || !visibleTabKeys.length) {
+      return fallbackTabKey;
+    }
+    if (tabFromUrl && visibleTabKeys.includes(tabFromUrl)) {
+      return tabFromUrl;
+    }
+    return fallbackTabKey;
+  }, [tabFromUrl, visibleTabKeys, fallbackTabKey]);
+
+  useEffect(() => {
+    if (!fallbackTabKey || !visibleTabKeys.length) {
+      return;
+    }
+    if (tabFromUrl === currentTab) {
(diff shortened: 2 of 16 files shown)


packages/web/app/hooks/useGoBack.js
```diff
@@ -0,0 +1,22 @@
...
+ * instead.
+ */
+export const useGoBack = (fallback = '/') => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  return () => {
+    if (location.key !== 'default') {
+      navigate(-1);
+    } else {
+      navigate(fallback, { replace: true });
```
jaskfla, line 19: Doesn’t `replace` break the ‘forward’ button behaviour? In many cases moot since Tamanu is used as a web app, and back button is often hidden; but I wonder if push is more appropriate here

packages/web/app/utils/useSyncedTabSearchParam.js
jaskfla, line 1: > [!NOTE]
> **Don’t let this comment block merge.** This is already a meaningful improvement from status quo, and we’ve discussed offline that this PR is a bit time-sensitive.)

Couple cents, none of which need to be addressed here and now.

> **TD;DR.** Tab state should usually be a path segment, not a search param.

- Generally, tabs are more appropriate as a `<nav>` of `<a>`’s, where the tab key is a path segment rather than a search param. (The alternative being a `<ul>` of `<button>` that rely on event listeners (or—god forbid—Effects) to keep the active tab state.)
- This means a reusable hook likely isn’t the “right” tool, as routing is more typically one-off code, not reusable.
- The main pitfall I see with this approach is: if a single instance of `useSyncedTabSearchParam` is mounted, then the `tab` search param is of “taken”, and will collide with anything else that may use it, **including other instances of `useSyncedTabSearchParam`**.
  - **Practical implication.** This means `useSyncedTabSearchParam` doesn’t work with nested tabs. e.g. Time complex charts within the ‘Chart’ tab of the encounter view
  - **DevX implication.** This behaviour is unintuitive. The thing that mounts `useSyncedTabSearchParam` could be in a completely different part of the codebase from where a dev is working (violating locality of behaviour). Tricky both for humans, and I suspect _also_ for agents’ first pass.[^foot]

| Path | Verdict |
|:--------|:--------|
| `…/:patientId/encounter/:encounterId?tab=charts` | Fine 🤷 |
| `…/:patientId/encounter/:encounterId/charts` | Better 😄  |
| `…/:patientId/encounter/:encounterId?tab=charts&tab=my-complex-chart` | Conflict ⚠️ |
| `…/:patientId/encounter/:encounterId/charts/my-complex-chart` | Clearer hierarchy & unambiguous 😎 |

[^foot]: Though I am confident that asking an AI “why is this happening” would quickly yield a correct answer.

packages/web/app/views/patients/EncounterView.jsx
```diff
@@ -173,31 +173,32 @@ const StyledTabDisplayDraggable = styled(TabDisplayDraggable)`
...
+  const visibleTabKeys = useMemo(() => visibleTabs.map(tab => tab.key), [visibleTabs]);
+  const fallbackEncounterTab = visibleTabs[0]?.key;
+  const { currentTab, onTabSelect } = useSyncedTabSearchParam(
+    visibleTabKeys,
+    fallbackEncounterTab,
+  );
 
   useEffect(() => {
     api.post(`user/recently-viewed-patients/${patient.id}`);
   }, [api, patient.id]);
```
jaskfla, line 190: Definitively out of scope, but really the `EncounterView` should have patient ID as its `key` prop

packages/web/app/views/patients/EncounterView.jsx
```diff
@@ -208,12 +209,6 @@ export const EncounterView = () => {
     }
   }, [userPreferences?.encounterTabOrders]);
 
-  useEffect(() => {
-    if (!currentTab) {
-      setCurrentTab(visibleTabs[0].key);
-    }
-  }, [isLoadingUserPreferences]);
```
jaskfla, line 207: Awesome to see these Effect abuses go 🧼

jaskfla (commented):
Just poking my nose in, as briefly discussed in Slack

Perhaps unsurprisingly, I do have some opinionated thoughts, but a nice improvement nonetheless. Great to see us shift away from immediately reaching for `useState` and `useContext`, and taking advantage of the free conveniences and a11y benefits that the browser gives us for free 🥳

packages/web/app/utils/useSyncedTabSearchParam.js
jaskfla: Ah, and FWIW, an existing example of nested tabs in Tamanu: https://central.main.cd.tamanu.app/admin/programs/registries/programRegistry-maternalhealthregistry/statuses

PR #10045, opened 2026-06-13
fix(sync): TAM-6858: Fix patient_invoice_insurance_plan duplicate issue (HOTFIX 2.54)

chris-bes:
### Changes

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
+++ b/packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
@@ -6,30 +6,33 @@ export async function up(query: QueryInterface): Promise<void> {
   // The released insert-new-on-re-add behaviour can leave a live row plus soft-deleted
   // tombstones for the same pair. Keep the live row if one exists, otherwise the
   // soft-deleted row with the highest id (deterministic across servers since ids are synced).
+  //
+  // Their changelog (logs.changes) and sync_lookup entries are removed by record_id, not by
+  // table_name/record_type: neither large table is usefully indexed on its type column
+  // (sync_lookup is unique on (record_id, record_type); logs.changes has a hash index on
+  // record_id), so driving both deletes off the small set of just-deleted ids stays
+  // index-assisted and avoids a sequential scan of millions of rows. Chained data-modifying
+  // CTEs share that id set without a temp table or recomputing the survivor query. Surviving
+  // rows' sync_lookup entries are remapped to the new ids in the next migration, so no
+  // rebuild/re-pull is needed — and on facilities (no sync_lookup rows) it is a no-op.
   await query.sequelize.query(`
-    DELETE FROM patient_invoice_insurance_plans
-    WHERE id NOT IN (
-      SELECT DISTINCT ON (patient_id, invoice_insurance_plan_id) id
-      FROM patient_invoice_insurance_plans
-      ORDER BY patient_id, invoice_insurance_plan_id, (deleted_at IS NULL) DESC, id DESC
-    );
-  `);
-
-  // Remove changelog entries for the hard-deleted duplicates, including the delete
-  // entries the audit trigger writes for the dedupe above.
-  await query.sequelize.query(`
+    WITH deleted_duplicates AS (
+      DELETE FROM patient_invoice_insurance_plans
+      WHERE id NOT IN (
+        SELECT DISTINCT ON (patient_id, invoice_insurance_plan_id) id
+        FROM patient_invoice_insurance_plans
+        ORDER BY patient_id, invoice_insurance_plan_id, (deleted_at IS NULL) DESC, id DESC
+      )
+      RETURNING id
+    ),
+    cleared_sync_lookup AS (
+      DELETE FROM sync_lookup
+      WHERE record_id IN (SELECT id FROM deleted_duplicates)
+        AND record_type = 'patient_invoice_insurance_plans'
+    )
     DELETE FROM logs.changes
-    WHERE table_name = 'patient_invoice_insurance_plans'
-    AND record_id NOT IN (SELECT id FROM patient_invoice_insurance_plans);
-  `);
-
-  // Remove sync_lookup entries for the hard-deleted duplicates (orphans). Surviving rows'
-  // entries are remapped to the new ids in the next migration, so no rebuild/re-pull is
-  // needed — central only runs sync_lookup, on facilities this is a no-op.
-  await query.sequelize.query(`
-    DELETE FROM sync_lookup
-    WHERE record_type = 'patient_invoice_insurance_plans'
-    AND record_id NOT IN (SELECT id FROM patient_invoice_insurance_plans);
+    WHERE record_id IN (SELECT id FROM deleted_duplicates)
+      AND table_name = 'patient_invoice_insurance_plans';
   `);
 }
 
--- a/packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
+++ b/packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
@@ -7,14 +7,23 @@ export async function up(query: QueryInterface): Promise<void> {
   // tombstones for the same pair. Keep the live row if one exists, otherwise the
   // soft-deleted row with the highest id (deterministic across servers since ids are synced).
   //
-  // Their changelog (logs.changes) and sync_lookup entries are removed by record_id, not by
-  // table_name/record_type: neither large table is usefully indexed on its type column
-  // (sync_lookup is unique on (record_id, record_type); logs.changes has a hash index on
-  // record_id), so driving both deletes off the small set of just-deleted ids stays
-  // index-assisted and avoids a sequential scan of millions of rows. Chained data-modifying
-  // CTEs share that id set without a temp table or recomputing the survivor query. Surviving
-  // rows' sync_lookup entries are remapped to the new ids in the next migration, so no
-  // rebuild/re-pull is needed — and on facilities (no sync_lookup rows) it is a no-op.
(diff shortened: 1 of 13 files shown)


packages/database/src/sync/saveIncomingChanges.ts
```diff
@@ -51,8 +51,13 @@ export const saveChangesForModel = async (
     const incoming = idToIncomingRecord[existing.id];
     idsForUpdate.add(existing.id);
 
-    // Restores only originate from central server
-    if (isCentralServer === false && existing.deletedAt && !incoming?.isDeleted) {
+    // Restores only originate from the central server, except for models that opt in
+    // to facility-originated restores (see Model.acceptsFacilityRestores)
+    if (
+      (isCentralServer === false || model.acceptsFacilityRestores) &&
```
chris-bes: This PR is to migrate `patient_invoice_insurance_plan.id` to be deterministic, hence, instead of creating a new record when a previously deleted `patient_invoice_insurance_plan` is re added, the data is restored instead 

However, the current sync model does not allow data restore originates from facility server (see original code). This was originally done by Biao which I'm not 100% sure the reason behind it. 

My guess is I can understand it coming from the idea that restoring data is more of an admin task, but I don't see any problem with allowing it globally now since `patient_invoice_insurance_plan` is a legit case that we should allow restore coming from facility server. 

However I'm also wary that any small change in sync can causes a lot of unknown side effects, so I'm playing it safe now and only allow it for `patient_invoice_insurance_plan` for now. 

What do you reckon @edmofro ?

packages/database/src/migrations/1781222400001-makePatientInvoiceInsurancePlanIdDeterministic.ts
```diff
@@ -0,0 +1,65 @@
...
+// pair offline mint the same id and dedupe on sync instead of colliding.
+//
+// logs.changes and sync_lookup are remapped to the new ids (the preceding migration already
+// removed their orphaned entries for the deduped rows). The transform is deterministic and
+// runs on every server, so remapping without bumping the sync tick keeps central consistent
+// with each facility's own migrated rows — no re-pull. The only DML is on those two tables,
+// not on patient_invoice_insurance_plans itself, so adding/dropping columns on the table
+// does not hit pending audit trigger events.
+
+export async function up(query: QueryInterface): Promise<void> {
```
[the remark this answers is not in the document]
chris-bes: This wouldn't happen as it would need DML and DDL on the same table. However, I'm making DDL change on `patient_invoice_insurance_plans` and DML change on `logs.changes` and `sync_lookup`. I've also tested running this migration locally on existing data and it works fine

packages/database/src/sync/saveIncomingChanges.ts
edmofro: I think the reason is that our sync conflict handling is last-write-wins, but with a trump card that delete-always-wins. 

Imagine that a record is on two facility servers
- Facility A deletes it, then syncs that delete up to the central server
- Before syncing down that delete, Facility B edits it, then syncs that edit up to the central server

This would create an unexpected restore. 

One option is to change how we store whether an insurance plan is deleted, and use something like `visibility_status`, or `removed_at` instead, which avoids facility restores of soft deleted records. 

Thoughts @chris-bes?

packages/database/src/models/Model.ts
```diff
@@ -35,6 +35,13 @@ export class Model<
   ) => string | null;
   declare static adjustDataPostSyncPush?: (ids: string[]) => Promise<void>;
   declare static incomingSyncHook?: (changes: SyncSnapshotAttributes[]) => Promise<SyncHookSnapshotChanges | undefined>;
+  // Restores normally only originate from the central server; models that revive
+  // soft-deleted rows in place (rather than inserting a new record) can opt in to
+  // having facility-originated restores applied on central. Note there is no tick-based
+  // conflict resolution on deletion state, so a live copy pushed by a facility that
+  // hasn't yet pulled a deletion will resurrect the record — only opt in where
+  // re-add-wins is acceptable.
```
edmofro: Oh yep this comment (presumably by claude) nailed the tradeoff

packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
```diff
@@ -0,0 +1,41 @@
+import { QueryInterface } from 'sequelize';
+
+export async function up(query: QueryInterface): Promise<void> {
+  // The next migration makes the id deterministic and the primary key composite on
+  // (patient_id, invoice_insurance_plan_id), so each pair must have exactly one row first.
+  // The released insert-new-on-re-add behaviour can leave a live row plus soft-deleted
+  // tombstones for the same pair. Keep the live row if one exists, otherwise the
+  // soft-deleted row with the highest id (deterministic across servers since ids are synced).
```
edmofro, line 8: _If_ they are in sync at the time of upgrade...?

packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
```diff
@@ -0,0 +1,41 @@
...
+      RETURNING id
+    ),
+    cleared_sync_lookup AS (
+      DELETE FROM sync_lookup
+      WHERE record_id IN (SELECT id FROM deleted_duplicates)
+        AND record_type = 'patient_invoice_insurance_plans'
+    )
+    DELETE FROM logs.changes
+    WHERE record_id IN (SELECT id FROM deleted_duplicates)
+      AND table_name = 'patient_invoice_insurance_plans';
```
edmofro: Ah ok so this has come up before, but we haven't solved it. We shouldn't delete records from `logs.changes`, it is an immutable and sacred audit log. Instead we need a way to represent that a record in `logs.changes` has had its source record deleted. 

@julianam-w do you have any suggestions here?

packages/database/src/migrations/1781222400001-makePatientInvoiceInsurancePlanIdDeterministic.ts
```diff
@@ -0,0 +1,66 @@
+import { QueryInterface } from 'sequelize';
+
+// Make the id deterministic from (patient_id, invoice_insurance_plan_id), same pattern as
+// patient_facilities / patient_ongoing_prescriptions. REPLACE avoids collisions if ';'
+// appears in a component id. The composite primary key gives each pair exactly one row for
+// its whole lifecycle, so removing a plan soft-deletes that row and re-adding it restores
+// the same row (rather than inserting a new record), and two facilities adding the same
+// pair offline mint the same id and dedupe on sync instead of colliding.
+//
+// logs.changes and sync_lookup are remapped to the new ids (the preceding migration already
```
edmofro: Again we are making changes to `logs.changes`. This is a lot less bad than removing records entirely, and we have done it before, but it's not entirely comfortable. 

Thoughts @julianam-w @passcod?

edmofro (commented):
I have comments to talk through

packages/database/src/migrations/1781222400000-dedupePatientInvoiceInsurancePlans.ts
julianam-w: Add a record in `logs.changes`with a filled `record_deleted_at`and add a note in `migration_context`/ `reason` (I'm not sure which is for what)

packages/database/src/migrations/1781222400001-makePatientInvoiceInsurancePlanIdDeterministic.ts
julianam-w: Is that what the `migration_context` field is for? We create a new record in `logs.changes` with the change? This looks similar to updating the `registration_status` in `patient_program_registrations`

packages/database/src/sync/saveIncomingChanges.ts
chris-bes: Yeah I did think of that approach, but did not have a strong preference since it would mean every time we want to allow restoring originates from facility, we would have to add an extra column there... But now after some more thinking, I don't think it's going to be a common case. Combined with the unexpected restore behaviour, I'm probably fine with your suggestion.

packages/database/src/migrations/1781222400001-makePatientInvoiceInsurancePlanIdDeterministic.ts
chris-bes: I think because we are migrating UUIDs to deterministic IDs which can be treated as deleting and recreating data, we can do both of these:
- Insert new logs.changes with record_deleted_at that reflects the old patient_invoice_insurance_plans data (data that using UUIDs for IDs) have been deleted
- Insert new logs.changes that reflects the latest patient_invoice_insurance_plans data that has got the IDs migrated to deterministic IDs (ie: patient_id;insurance_plan_id)

edmofro: @julianam-w happy?

julianam-w: Yes! 🙂

julianam-w (approved):

edmofro (approved):
lgtm!

PR #10069, opened 2026-06-16
chore: TAM-6847: upgrade Node to 26

passcod:
### Changes

🤖 Node 20 is EOL. This moves the dev/CI/container Node version straight to 26: `.node-version` (and the `.nvmrc` symlink that points at it), the root and `synthetic-tests` Dockerfiles, and the mobile `.nvmrc`.

Validated empirically with a clean install + full build + test suites on Node 26:

- Native modules (`bcrypt`, `@swc/core`, `sqlite3`) are all N-API prebuilds and ABI-independent, so they need no rebuild and load fine on 26.
- `turbo build` (tsc/vite/esbuild/swc), web + utils vitest, the mobile Metro bundle, mobile Jest, and the central-server integration harness (migrations + sequelize + bcrypt + @swc/jest) all pass.

`engines.node` is deliberately left at `>=20.9.0`: server artifacts built under Node 26 still run on Node 20 (verified by running the built central-server integration suite back on Node 20), so deployments on system Node 20 are **not** forced to upgrade in lockstep — the host Node upgrade can be sequenced separately. The only packages in the tree requiring >20 are `concurrently` and `artillery`, both dev/test tooling not present in any server's runtime.

One caveat worth flagging: once dev, CI, and containers all run 26, nothing exercises Node 20 anymore, so a future code change or dependency bump could silently raise the runtime floor without CI noticing. A temporary CI lane running the server suite on Node 20 would guard against that during the transition window — left out here on purpose, as the call is operational rather than technical.

The upgrade surfaced three independent CI failures, all fixed here (each reproduced in a clean container):

- **`npm ci` hang (cache job)** — the unused, transitive `@playwright/browser-chromium` (pulled in by `artillery` → `artillery-engine-playwright`; synthetic-tests runs no browser scenarios) downloads a browser in its postinstall, and Playwright's pinned extractor hangs during *extraction* on recent Node. An `overrides` entry aliases it to a no-op so it never downloads or extracts — also saving a pointless 167 MiB download in dev.
- **Container image build** — the package-pruning step (`scripts/docker-build.sh` → `list-packages.mjs` → `_do-with-all-packages.mjs`) derived the workspace build-order graph from `npm ls`, which exits non-zero (crashing the step) on intentional, pre-existing tree "problems" — the `overrides`-pinned react flagged "invalid" against `@material-ui` v4's stale peer range, and an extraneous-looking package. It now reads the workspace list and dependencies straight from the `package.json` files instead of `npm ls`.
- **E2E "Install Playwright browsers"** — Playwright 1.59.x's out-of-process browser installer hangs during *extraction* on recent Node (the chromium zip downloads to 100%, then the extractor stalls; the same version extracts fine on Node 20, and 1.61.0 fine on newer Node). Bumped `@playwright/test` 1.59.1 → 1.61.0.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


jaskfla (approved):
Surprisingly small diff indeed! 🗜️

NavarroEmilioLuis:
FYI the package-lock clashes with mine. I suggest to keep mine and then run npm install so it updates itself. I had to manually remove packages from the root to avoid bundling issues.

passcod:
Yep, working through it. An added complication is that Node 24+ ships with npm 11, which changes some behaviour wrt optional/peer dependencies.

PR #10082, opened 2026-06-17
feat(reporting): TAM-6862: Tamanu manages reporting/raw db roles

dannash100:
### Changes
Tamanu now **creates and manages** the reporting/raw database roles itself instead of relying on manually-provisioned login users with credentials in config. Database credentials collapse to the single core ("tamanu") connection, and reporting is **always on** (the `reportSchemas.enabled` config flag is removed).

ops:
- [tamanu: provision for self-managed, always-on reporting roles (TAM-6862)](https://github.com/beyondessential/ops/pull/235)
- [tamanu: use a single DATABASE_URL for the core db (ansible + k8s) (TAM-6862)](https://github.com/beyondessential/ops/pull/237)

**Includes the DATABASE_URL change:** [feat(database): TAM-6862: support a single DATABASE_URL connection string (#10106)](https://github.com/beyondessential/tamanu/pull/10106) was merged into this branch rather than landing on its own, so this PR also carries the collapse to a single `DATABASE_URL` core connection string (the matching infra change is ops#237).

**How it works**
- On startup Tamanu provisions the `tamanu_reporting` / `tamanu_raw` roles — `CREATE ROLE`, password, schema, `GRANT SELECT`, and `ALTER DEFAULT PRIVILEGES` so tables materialised later stay readable. Idempotent and self-healing.
- The reporting connections **log in AS the unprivileged role** (not `SET ROLE`).
- The role passwords derive (HMAC) from a **random per-server secret**, generated once and stored in `local_system_secrets` (see _Secret storage & read isolation_ below) — not synced. So they're real, instance-unique secrets regardless of the core db auth method (trust, peer or password).
- Report routes key their schema behaviour off whether the reporting connections are present, so there's no separate enable flag.
- Removed the per-connection `username`/`password` config and the redundant `createMockReportingSchemaAndRoles` test helper.

**Why log in as the role, not `SET ROLE`**
`SET ROLE` is reversible (the authenticated user stays the privileged core user), so report SQL could `RESET ROLE` — or end the surrounding transaction with `COMMIT` and continue in autocommit — to write as core. Authenticating **as** the unprivileged role removes the escape: the session has no write grants and is a member of no other role. Covered by tests that run `COMMIT; RESET ROLE; INSERT …` through the report runner and a trailing write through the central `verifyQuery` path, both rejected with no data written.

**Secret storage & read isolation**
The `raw` role is granted `SELECT` on all of `public`, so any secret stored there is readable by report SQL. To prevent that:
- Server secrets — the per-server reporting secret **and the device key** — now live in a new **`local_system_secrets`** table (mirrors `local_system_facts`; `DO_NOT_SYNC`; excluded from the sync-tick + changelog triggers so values never reach `logs.changes`). A DDL migration creates it and a DML migration moves the two rows out of `local_system_facts`; dbt source model added with `nil` masking on `value`.
- The `raw` role is **revoked** from `local_system_secrets` and from the credential/token tables (`one_time_logins`, `portal_one_time_tokens`, `refresh_tokens`, `signers`/`signers_historical`), so reports can't read the device private key, auth tokens or signing keys. It still reads `local_system_facts` (sync ticks etc.) normally.
- `LocalSystemSecret` also offers opt-in **encrypted** accessors (`setSecret`/`getSecret`, AES via `config.crypto.keyFile`) for external credentials that shouldn't be plaintext even behind the grant (e.g. a future sync_host password). Self-generated internal values (reporting secret, device key) stay on the plain path, so the always-on startup path keeps no key-file dependency.

**Concurrency**
`startAll` brings up the api, fhir workers and tasks runner concurrently, each provisioning/opening the reporting connections, so two safeguards keep them from racing on the cluster-global roles:
- role DDL runs inside a transaction-scoped advisory lock (`CREATE`/`ALTER ROLE` would otherwise hit "tuple concurrently updated");
- `openDatabase` caches the in-flight connection promise, so concurrent opens of the shared reporting connection key share one connection instead of throwing.

**Startup ordering**
Reading the per-server secret needs a migrated db, so facility opens its reporting connections **after migrate** via a new `initReportingStores()` (called from the start subcommands) rather than in `init()`. Central already inits after migrate.

**⚠️ Deployment notes**
- The core db user needs **`CREATEROLE`** to provision the roles.
- The `tamanu_reporting` / `tamanu_raw` roles need **`pg_hba`** entries permitting them to authenticate (they log in with the derived password). Both are handled by the ops PR for ansible single-servers; k8s gets `CREATEROLE` via the cluster initSql.
- **No settings PSK or extra secret config required** — the reporting secret is self-generated into `local_system_secrets`, so this works on trust/peer/minimal servers too. (The opt-in encrypted accessors need `config.crypto.keyFile`, but nothing in this PR uses them; on k8s the key file is already provisioned.)
- Existing deployments: drop the old `reportSchemas.connections.*.username/password` and `reportSchemas.enabled` from config (now removed/ignored); startup re-aligns any pre-existing roles' password/grants.

**Deferred / considered (kept simple on purpose)**
- *Provision in a migration* (to avoid the per-boot `GRANT`/`CREATEROLE` cost): reverted — roles are cluster-global, so a migration running in every parallel test/replica DB raced on the shared role catalogue. Startup provisioning + the advisory lock avoids the race. Revisit if the per-boot `GRANT ON ALL TABLES` is measured to hurt at scale.
- *Secret in settings (encrypted, scoped)*: considered — but settings are `PULL_FROM_CENTRAL` (central-owned), which fights a self-bootstrapping per-server credential, and secret settings need a PSK that minimal/trust/peer servers don't have. `local_system_secrets` (self-generated per server, like the device key) sidesteps both with no extra config.

**Testing:** `ReportSchemaRoles` (facility) and `verifyQuery` reporting-path (central) suites pass, including the sandbox-escape cases. Also smoke-tested a real local central boot (offline upgrade → provision → serve): HTTP 200, both roles log in, secret persisted. The `local_system_secrets` move + raw revoke were verified live on a k8s deploy (`raw` can read `local_system_facts` but not `local_system_secrets`/token tables); database build + lint clean, `dbt-check-todos` passes.

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Synthetic test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/database/src/services/reporting.js
```diff
@@ -1,70 +1,100 @@
...
-  }
-
-  await validateUser(existingStore, username);
-  await grantPrivileges(existingStore, REPORT_DB_CONNECTION_SCHEMAS[connectionName], username);
 
   return openDatabase(`reporting-${connectionName}`, overrides);
 };
 
 export const initReporting = async existingStore => {
+  if (!config.db.password) {
```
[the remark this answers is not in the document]
passcod: This, basically. If the connection is done over trust/peer locally, then db.password will be empty, and the password generation for reporting/raw won't work + be insecure.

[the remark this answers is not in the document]
passcod: This, basically. If the connection is done over trust/peer locally, then db.password will be empty, and the password generation for reporting/raw won't work + be insecure.

packages/database/src/models/LocalSystemSecret.ts
```diff
@@ -0,0 +1,116 @@
...
+    const newDeviceKey = EndpointKey.generateFor('ecdsa256');
+    await this.set(FACT_DEVICE_KEY, newDeviceKey.privateKeyPem());
+    return newDeviceKey;
+  }
+
+  // Encrypted-at-rest accessors, for secrets that shouldn't be plaintext even
+  // behind the table grant. Encryption uses the server key file
+  // (config `crypto.keyFile`), so only callers that store such a secret pull in
+  // that dependency; the self-generated device key / reporting secret stay on
+  // the plain get/set path.
```
passcod: Hmm, it might be more elegant to only have encrypted secrets in the table?

passcod (approved):

packages/database/src/models/LocalSystemSecret.ts
dannash100: Yeah I agree, ai pushed back on this because it would break deployments because of deviceKey if they dont have crypto keyFile, but if thats sortable thats all good.
I will QA this card and [feat(database): TAM-6862: support a single DATABASE_URL connection string](https://github.com/beyondessential/tamanu/pull/10106#top)
today

passcod: deviceKey in particular is essentially dead

dannash100: Ok cool! Should i remove that logic even?

dannash100:
Stacked on top of this: #10148 (feat: auto-rotate the reporting role secret) — needs review. It builds on the reporting-roles work here and adds age-based rotation of the per-server secret.

packages/central-server/app/admin/reports/reportRoutes.js
```diff
@@ -20,12 +19,12 @@ export const reportsRouter = express.Router();
...
   '/',
   asyncHandler(async (req, res) => {
-    const { store } = req;
+    const { store, reportSchemaStores } = req;
     req.checkPermission('read', 'ReportDefinition');
     req.checkPermission('read', 'ReportDefinitionVersion');
 
     const canEditSchema = req.ability.can('write', 'ReportDbSchema');
-    const isReportingSchemaEnabled = config.db.reportSchemas.enabled;
+    const isReportingSchemaEnabled = Boolean(reportSchemaStores);
```
passcod: Feels like `req.reportSchemaStores` should be ensured boolean upstream instead of coercing it at every use site here.

packages/database/config/dev-secret.key
passcod, line 1: This file seems different from the central-server dev-secret.key, might cause issues?

packages/database/src/services/reporting.js
```diff
@@ -1,70 +1,144 @@
...
+    // reports), and the rest hold auth tokens or certificate signing keys. Revoke
+    // SELECT on them.
+    // to_regclass skips any not present on this server (e.g. central-only ones).
+    if (schema === 'public') {
+      await sequelize.query(`
+        DO $$
+        DECLARE
+          sensitive_table text;
+        BEGIN
+          FOREACH sensitive_table IN ARRAY ARRAY[
```
passcod: Feels like this should be an array in constants or something rather than inline in the query; minor.

packages/facility-server/config/dev-secret.key
passcod, line 1: Same here

passcod (approved):
good, some nits

dannash100:
I just need to rebase to main to fix determinism I think its all good, I was tryna fix but it aint a thing.

Harnesh787 (approved):

packages/central-server/config/default.json5
```diff
@@ -520,5 +505,11 @@
     "tokenDuration": "24h",
     "loginTokenDurationMinutes": 20,
     "registerTokenDurationMinutes": 43800 // 1 month
+  },
+  "crypto": {
+    // Encryption key for local_system_secrets (encryption is mandatory).
+    // This committed key is for local dev/test ONLY; deploys override it via
+    // the CRYPTO_KEY_FILE env var pointing at a mounted secret.
+    "keyFile": "config/dev-secret.key"
```
passcod, line 513: hmmmm I think this being in default means windows deploys will start out using it?

hrazasalman (approved):

PR #10124, opened 2026-06-23
feat(sync): TAM-6863: facility sync-host setup workflow

dannash100:
### Changes

First-run setup for pointing a facility server at its sync host, held in **local system facts** instead of config (TAM-6863).

- **Boot resolution** — `serverConfig` resolves the sync host/credentials + facility ids once at boot: `SYNC_URL`/`SYNC_FACILITY_IDS` env → local system fact → legacy `config.sync.*` / `serverFacilityId(s)` (now blank by default). Runtime reads `getSyncConfig()` / `getServerFacilityIds()` / `isServerConfigured()`. When unconfigured, boot skips timesync/central/sync so the server still comes up to serve the wizard.
- **Wizard** — an unconfigured facility shows a setup wizard instead of login (gated by `setupRequired` on `GET /api/public/ping`): sync URL + central **admin** credentials + facility id(s). On submit the facility authenticates the admin against central, requires `manage:all`, has central provision a dedicated sync user, and stores the returned credentials.

Screenshot 2026-06-29 at 11 52 01 AM

- **Endpoints** — facility `POST /api/public/setup/sync` (trusted-network gated; writes config in one transaction); central `POST /api/admin/syncCredentials` (`manage:all`) find-or-rotates a `System: … sync` user.
- **Secrets** — sync password stored encrypted in `local_system_secrets`; host/username/facility ids are plain facts.
- **Sync user identity & filtering** — the provisioned account is keyed on the device, not the facility set: its email is deterministic (`sync.<sha256(deviceId)[:32]>@sync.tamanu`), so re-running setup from the same server rotates the same account rather than accumulating users, and two servers serving the same facilities can't overwrite each other's credentials. Machine accounts carry a new `users.kind` column (`sync` vs `user`, set by both the wizard provisioning and the `provision` subcommand) and are filtered out everywhere humans are expected: the central admin user list (list + direct fetch), the reference-data user exporter, and the facility user/clinician listing.
- **No restart needed** — after setup the sync/tasks process polls, re-reads the config, and wires up the sync runtime in-process (`SyncTask` no-ops until ready).
- **Ops** — `resetSyncConfig` facility subcommand clears config to re-run setup; `SYNC_URL` keeps the single connection-string form for headless/k8s.

> Deployment: `sync.host` / `sync.email` / `sync.password` and `serverFacilityId(s)` are now a fallback only — set the sync target via the wizard or `SYNC_URL` / `SYNC_FACILITY_IDS`.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [x] Companion ops branch 
- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/central-server/app/admin/syncCredentials.js
```diff
@@ -0,0 +1,39 @@
...
+  req.checkPermission('manage', 'all');
+
+  const { facilityIds } = bodySchema.parse(req.body);
+  const uniqueFacilityIds = [...new Set(facilityIds.map(id => id.trim()))].sort();
+
+  const { User } = req.store.models;
+
+  // Deterministic per-server email so re-running setup rotates the same account
+  // rather than accumulating sync users.
+  const email = `sync.${uniqueFacilityIds.join('-')}@sync.tamanu`;
```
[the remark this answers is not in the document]
passcod: Alternatively, and that would also avoid collisions for e.g. IDs `a`, `b`, and `a-b`, generate via hashing: `sha2(ids.join("\0"))` is fixed length

packages/central-server/app/admin/syncCredentials.js
```diff
@@ -0,0 +1,39 @@
...
+  const displayName = `System: ${uniqueFacilityIds.join(', ')} sync`;
+  const password = crypto.randomBytes(24).toString('base64url');
+
+  const existing = await User.findOne({ where: { email } });
+  if (existing) {
+    existing.set({ displayName, role: 'admin' });
+    existing.setPassword(password);
+    await existing.save();
+  } else {
+    await User.create({ email, displayName, role: 'admin', password });
```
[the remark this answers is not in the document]
passcod: There's other things it needs but not a bad point

packages/web/app/api/queries/useSetupStatusQuery.js
```diff
@@ -0,0 +1,26 @@
+import { useQuery } from '@tanstack/react-query';
+import { useApi } from '../useApi';
+
+// Unauthenticated check used pre-login to decide whether to show the first-run
```
passcod: Instead of having a dedicated endpoint and a new request, we could have the existing status endpoint return a special value if setup is required.

packages/facility-server/app/routes/apiv1/setup.js
```diff
@@ -0,0 +1,124 @@
...
+  const uniqueFacilityIds = [...new Set(facilityIds.map(id => id.trim()))];
+  const normalisedHost = new URL(host).origin;
+
+  // The deliberate part: prove the host is reachable and the credentials
+  // authenticate, before saving anything. Also the authorisation gate — this
+  // endpoint is unauthenticated (a fresh server has no users), so we require the
+  // supplied credentials to be a central superuser. (Next pass: central
+  // provisions a dedicated sync user once the admin has authenticated here.)
+  const probe = new TamanuApi({
+    endpoint: `${normalisedHost}/api`,
```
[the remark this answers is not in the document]
passcod: That's not viable, because all of these *could* be valid targets. That said, we should validate that the _request to this endpoint_ comes from either localhost, private ranges, or tailscale, making sure to allow both IPv4 and v6 of these ranges.

passcod (changes requested):

packages/facility-server/app/routes/apiv1/setup.js
```diff
@@ -0,0 +1,163 @@
...
+    return res.status(409).send({ error: { message: 'Server is already configured' } });
+  }
+
+  const { host, email, password, facilityIds } = setupSyncSchema.parse(req.body);
+  const uniqueFacilityIds = [...new Set(facilityIds.map(id => id.trim()))];
+  const normalisedHost = new URL(host).origin;
+
+  // Validate the host + credentials against central, and use them as the authz
+  // gate (must be a central super-admin) before saving anything.
+  const probe = new TamanuApi({
```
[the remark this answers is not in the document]
dannash100: Didnt act on this one yet

dannash100:
**Ops (k8s autodeploy):** beyondessential/ops#245 (draft)

Configures k8s autodeploy facilities via local system facts using the new `setupSync` subcommand (added here), instead of the `config.sync.*` fallback — so they boot fact-configured and never hit the setup wizard. Depends on this PR's image including `setupSync`.

packages/central-server/app/admin/syncCredentials.js
```diff
@@ -0,0 +1,49 @@
...
+});
+
+// Hash the ids so the email is deterministic (re-running rotates the same
+// account), fixed-length, and collision-free (['a','b'] vs ['a-b']).
+const syncUserEmail = facilityIds =>
+  `sync.${crypto
+    .createHash('sha256')
+    .update(facilityIds.join('\0'))
+    .digest('hex')
+    .slice(0, 32)}@sync.tamanu`;
```
passcod: Hmmm I just thought of something here: if we have two facility servers with the same set of facility IDs, they'll conflict, and worse, the second could overwrite the first's credentials. I think we need to go from the device ID instead.

packages/facility-server/app/subCommands/resetSyncConfig.js
```diff
@@ -0,0 +1,32 @@
...
+  FACT_FACILITY_IDS,
+} from '@tamanu/constants';
+
+import { ApplicationContext } from '../ApplicationContext';
+
+const SYNC_CONFIG_FACTS = [FACT_CENTRAL_HOST, FACT_SYNC_EMAIL, FACT_FACILITY_IDS];
+
+// Clears the sync facts so the server is unconfigured and can be set up again.
+// Host-level access only; restart the server afterwards to take effect.
+export const resetSyncConfig = async () => {
```
passcod: Would be good to have an interactive "Are you really sure?" confirm and a `--yes-i-am-really-sure` CLI flag for non-interactive use.

passcod (changes requested):

passcod (approved):

database/model/public/users.md
```diff
@@ -48,3 +48,9 @@ How many devices this user can register (when `features.deviceRegistrationQuota`
 
 See [`devices`](#!/source/source.tamanu.tamanu.devices) for more.
 {% enddocs %}
+
+{% docs users__kind %}
+Distinguishes machine accounts from humans: `user` for ordinary accounts, `sync` for the per-device accounts facility servers authenticate with to sync.
```
passcod: Also should have `system` for the special system user (all 0 uuid `id` column)

packages/central-server/app/admin/exporter/modelExporters/UserExporter.js
```diff
@@ -1,9 +1,13 @@
+import { Op } from 'sequelize';
+import { USER_KINDS } from '@tamanu/constants';
 import { ModelExporter } from './ModelExporter';
 
 export class UserExporter extends ModelExporter {
   async getData() {
     const modelName = 'User';
     const users = await this.models[modelName].findAll({
+      // Machine accounts for device sync are managed by facility setup, not spreadsheets
+      where: { kind: { [Op.ne]: USER_KINDS.SYNC } },
```
passcod: invert to match human users rather than match the negative, so that new user kinds have to be opted in to this flow rather than forgotten

packages/central-server/app/admin/users.js
```diff
@@ -64,6 +65,10 @@ const createUserFilters = (filterParams, models) => {
     {
       email: { [Op.ne]: ADMIN_USER_EMAIL },
     },
+    // Exclude device sync users — managed by facility setup, not this panel
+    {
+      kind: { [Op.ne]: USER_KINDS.SYNC },
```
passcod: same here

packages/database/src/migrations/1783118255000-addUserKind.ts
```diff
@@ -0,0 +1,14 @@
+import { DataTypes, QueryInterface } from 'sequelize';
+
+export async function up(query: QueryInterface): Promise<void> {
+  await query.addColumn('users', 'kind', {
+    type: DataTypes.STRING,
+    allowNull: false,
+    defaultValue: 'user',
```
passcod, line 7: this makes the system user get kind=user

passcod: and also existing sync users, but that's less obvious how to avoid

passcod (commented):
additionally, we might consider opting out sync users from the device quota checks

packages/central-server/app/admin/exporter/modelExporters/UserExporter.js
dannash100: yeah tru

Harnesh787 (approved):

passcod (approved):

julianam-w (approved):

PR #10131, opened 2026-06-23
chore: no-issue: scaffold top level specs

tcaiger:
### Changes

Add top-level `specs/` directory the skills write to.


jaskfla:
> Adds a cluster of Claude Code skills under `.claude/skills/`

Can I suggest using `.agents/skills/`? I have a PR incoming that uses that, and symlinks `.claude/skills` to it

PR #10155, opened 2026-06-26
feat(medication): TAM-6822: Update MedicationDiscontinueModal to have current user as default

sepidehdehghani992:
### Changes

Update medication discontinue modal to have current user as default

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [x] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/e2e-tests/tests/patients/medication.spec.ts
```diff
@@ -1,5 +1,70 @@
...
+
+    await medicationPane.clickFirstOngoingMedicationRow();
+
+    const detailsModal = new MedicationDetailsModal(page);
+    await detailsModal.waitForModalToLoad();
+    const discontinueModal = await detailsModal.clickDiscontinue();
+
+    await discontinueModal.changeDiscontinuedBy(currentUser.displayName);
+    await discontinueModal.fillReason('Test reason');
+    await discontinueModal.submit();
```
edmofro, line 64: @sepidehdehghani992 I notice some of these tests have no assertions at all. Do they do any testing? Are you just trying to test that nothing errors during the user flow?

sepidehdehghani992: @edmofro I was trying to test the workflow is not throwing error end to end but I think I need to add a assertion that medication is actually discontinued. will apply that. Thanks

packages/e2e-tests/pages/patients/MedicationsPage/modals/MedicationDiscontinueModal.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.discontinueButton = page.getByTestId('formsubmitbutton-discontinue-def456');
+    this.cancelButton = page.getByTestId('formcancelbutton-cancel-xyz789');
+  }
+
+  async waitForModalToLoad(): Promise<void> {
+    await this.discontinuedByField.waitFor({ state: 'visible' });
+  }
+
+  async getDiscontinuedByValue(): Promise<string> {
+    return (await this.discontinuedByInput.inputValue()).trim();
```
jaskfla: Tiniest of nitpicks, but it seems unexpected to trim here. I’d suggest maybe trimming when `getDiscontinuedByValue()` is called (only if appropriate), otherwise there could be a risk of something not working in production because of untrimmed input, but passing tests here

packages/e2e-tests/utils/apiHelpers.ts
```diff
@@ -278,6 +307,49 @@ export const createEncounterPrescriptionViaApi = async (
...
+    units: 'mg',
+    frequency: 'Immediately',
+    isOngoing: true,
+    ...overrides,
+  };
+
+  const url = constructFacilityUrl(`/api/medication/patientOngoingPrescription/${patientId}`);
+  const response = await api.post(url, { data: prescriptionData });
+
+  if (!response.ok()) {
```
jaskfla, line 370: Worth checking with agent, but I’m fairly confident `ok` is a property, not a function

Should be `response.ok`, not `response.ok()`

packages/web/app/components/Medication/MedicationDiscontinueModal.jsx
```diff
@@ -77,7 +79,7 @@ export const MedicationDiscontinueModal = ({ medication, onDiscontinue, onClose
         onSubmit={onSubmit}
         onSuccess={onClose}
         formType={FORM_TYPES.CREATE_FORM}
-        initialValues={{}}
+        initialValues={{ discontinuingClinicianId: currentUser?.id }}
```
jaskfla, line 82: Also worth checking with agent, but I think this might never initialise if the form gets rendered before the `useAuth` context has loaded?

Shouldn’t be an issue in most cases, but if Tamanu is loaded directly to this modal the form _might_ be able to load before the current user has loaded 🤔

jaskfla (commented):
Only had a cursory glance at the E2E tests—looking sensible to me

Left a question about the form‘s initial values. I’m not intimately familiar with precisely how the user and this form loads, but my hunch is that there’s an edge case where the default clinician will fail to load

packages/e2e-tests/pages/patients/MedicationsPage/modals/MedicationDiscontinueModal.ts
sepidehdehghani992: @jaskfla below is the AI comment on this. I have removed trim() completely. no trim at call site either.

The value read is a default pre-fill resolved from the current user's record — the same source as the currentUser.displayName it's compared against — so there's no whitespace to normalise. The two call-sites compare it against currentUser.displayName, which comes from the same user record that pre-fills the field, so no trimming is appropriate at the call site either. If whitespace ever did appear, that'd be a real product discrepancy worth failing on rather than masking.

packages/e2e-tests/utils/apiHelpers.ts
sepidehdehghani992: @jaskfla I checked with agent and it says below:

 "On Playwright's APIResponse — which is what api.post(...) resolves to — ok() is a method, so response.ok() is correct. The confusion is understandable: the browser Fetch API's Response.ok is a property. But this is Playwright's APIRequestContext/APIResponse, not Fetch. Calling response.ok (without ()) would return the function reference, which is always truthy, so !response.ok would be false every time and the error branch would never fire — i.e. changing it as suggested would actually introduce a bug."

packages/web/app/components/Medication/MedicationDiscontinueModal.jsx
sepidehdehghani992: @jaskfla Agent is suggesting to leave it as it is because of below points:

- Tamanu only shows the logged-in app when state.auth.user exists — otherwise it shows the login screen (App.jsx:73).
- This discontinue modal lives deep inside that logged-in app, and only opens after you click Discontinue.
- So by the time the form mounts, currentUser is always there — currentUser?.id will never be empty.

The concern would only apply if the form could load before the user did, but the login gate makes that impossible. Adding enableReinitialize to "fix" it would risk wiping a half-typed reason for no real benefit.

jaskfla: Happy with that explanation. Let’s leave it as-is 🙂

jaskfla (approved):
Looks good—thanks 🙌

packages/e2e-tests/pages/patients/MedicationsPage/modals/MedicationDiscontinueModal.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.cancelButton = page.getByTestId('formcancelbutton-cancel-xyz789');
+  }
+
+  async waitForModalToLoad(): Promise<void> {
+    await this.discontinuedByField.waitFor({ state: 'visible' });
+  }
+
+  async getDiscontinuedByValue(): Promise<string> {
+    return this.discontinuedByInput.inputValue();
+  }
```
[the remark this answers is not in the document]
sepidehdehghani992: I have applied this review comment

edmofro:
Hey @sepidehdehghani992, have just removed from the merge queue as you made a few more changes since @jaskfla's last review, so need to re-request review (see screenshot for how)
image

jaskfla (approved):
Latest changes look good 👌

sepidehdehghani992:
@edmofro It passed rereview merging again

PR #10179, opened 2026-06-30
E1: Add padding after invoice items list

edmofro:
Adds bottom padding to the invoice items list in `InvoiceForm.jsx` so that when the invoice is finalised and the "Add item" button (via `FormFooter`) is hidden, there is still adequate visual spacing between the last item and the "Record payment" button below. Previously, the "Add item" button was incidentally providing that breathing room, so its absence on finalised invoices caused the layout to feel cramped, especially when a long item name wrapped to two lines.

The diff also includes a collection of agent skill and documentation files (`.agents/`, `AGENTS.md`) added to the repository, unrelated to the UI fix.

### Review Hero

- [ ] **Run Review Hero**

### DIFF
--- a/.agents/docs/spec-format.md
+++ b/.agents/docs/spec-format.md
@@ -1,0 +1,109 @@
+---
+workhorse-version: 0.1.0
+---
+
+# Spec format and information architecture
+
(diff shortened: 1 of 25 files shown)


edmofro:
Superseded by #10180 (workhorse setup only — padding change dropped).

PR #10202, opened 2026-07-04
feat(settings): TAM-6864: server-scoped settings

dannash100:
## Changes

experiment

Adds a fourth settings scope, `server`, for machine-level settings on a facility server — settings that make no sense keyed by facility when one server serves several facility ids (TAM-6863's device identity is the natural key). Server-scope rows are **centrally managed and device-routed**: central is the single author, and sync delivers each row only to the facility server it belongs to.

**Routing.** `settings.device_id` + `sync_lookup.device_id` (with a load-bearing partial index — see benchmark below). The snapshot predicate (both lookup and legacy paths) gains a device guard: broadcast/facility rows require `device_id IS NULL`; device rows sync only to the session pulling as that device (`OR device_id = :deviceId`, already a replacement param). Mobile never receives device rows; its settings table gains the mirrored column (TypeORM migration) and a defensive `deviceId IS NULL` read guard, and the admin device picker lists facility servers only.

**First tenants.** Sync pull tuning (`sync.persistedCacheBatchSize`, `sync.pauseBetweenCacheBatchInMilliseconds`, `sync.dynamicLimiter.*`) — never centrally manageable before (ops edit config on the box). The sync manager resolves tuning once per run via the `settings.server` reader (device rows → config fallback → schema defaults) and threads it into pull/stream.

**Migration.** A fact-gated facility upgrade step snapshots non-default config values into the `facility_setting_migrations` carrier keyed by deviceId (`facility_id` now nullable); central applies them as device-keyed server settings (skip-if-exists), which sync back down. Round-trip covered in `settingsMigrationRoundTrip.test.js`.

**Admin.** Settings GET/PUT accept `deviceId` (required for server scope); new `GET /admin/devices` feeds the scope picker's device dropdown ("Server (Single Facility Server)").

**Benchmark** (3M-row synthetic sync_lookup, production-shaped chunk query): old predicate ~28ms/chunk (BitmapOr on the tick index). New predicate **without** the device index: planner falls to parallel seq scan, ~110ms/chunk — 4x, far worse at production scale. With the partial index (`WHERE device_id IS NOT NULL`): **~25ms/chunk, plan and cost identical to old** (device arm joins the BitmapOr at 0.003ms). The index ships in the migration.

**Tests**: device routing on both snapshot paths (own device receives, other device / null-device sessions excluded, globals still broadcast), lookup build carries device_id, carrier round trip incl. operator-value-wins, upgrade step unit tests, settings schema/reader tests. dbt models regenerated for the three table changes.

Review focus: the predicate change in `snapshotOutgoingChanges.js` + `Setting.buildSyncFilter`, the lookup column plumbing, and the carrier change.

**Follow-up in this PR — fhir moves to server scope.** `fhir.worker.resourceMaterialisationEnabled` was facility-overridable with boot-time union-merging across a server's facilities (and an infoBanner explaining that enabling for one facility enabled it server-wide). It's now a server-scope override: the server reader cascade falls through to global (server row → global row → defaults, mirroring the facility cascade), `initFhirSettingsFromDb` collapses to a single reader with the merge logic deleted, and the banner is gone. A fact-gated central upgrade step replays the old merge once — facility rows grouped by the device that serves them (latest non-mobile sync session), any-true values become one device row (never clobbering an operator's server value), migrated facility rows soft-deleted so the removal syncs down; unmapped facilities are left untouched with a warning (their rows are inert). Covered in the round-trip test.

### DIFF
--- a/database/model/public/facility_setting_migrations.md
+++ b/database/model/public/facility_setting_migrations.md
@@ -19,3 +19,7 @@ JSON value from the legacy config file.
 {% docs facility_setting_migrations__facility_id %}
 The [facility](#!/source/source.tamanu.tamanu.facilities) the setting is scoped to.
 {% enddocs %}
+
+{% docs facility_setting_migrations__device_id %}
+For machine-level (server scope) carrier rows: the device id of the facility server the value came from. Null for facility-scoped rows, which use facility_id instead.
+{% enddocs %}
--- a/database/model/public/facility_setting_migrations.yml
+++ b/database/model/public/facility_setting_migrations.yml
@@ -49,11 +49,16 @@ sources:
             data_type: character varying(255)
             description: "{{ doc('facility_setting_migrations__facility_id') }}"
             data_tests:
-              - not_null
               - relationships:
                   arguments:
                     to: source('tamanu', 'facilities')
                     field: id
+          - name: device_id
+            data_type: text
+            description: "{{ doc('facility_setting_migrations__device_id') }}"
+            config:
+              meta:
+                masking: text
           - name: updated_at_sync_tick
             data_type: bigint
             description: "{{ doc('generic__updated_at_sync_tick') }} in facility_setting_migrations."
--- a/database/model/public/settings.md
+++ b/database/model/public/settings.md
@@ -27,4 +27,9 @@ One of:
 - `global`
 - `central`
 - `facility`
+- `server` (machine-level, keyed by device_id)
+{% enddocs %}
+
+{% docs settings__device_id %}
+For server-scope (machine-level) settings: the device id of the facility server this row belongs to. Null for all other scopes.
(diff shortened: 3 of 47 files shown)


passcod:
hmm why server settings and not facts?

dannash100:
> hmm why server settings and not facts?
@passcod 

Good point, I was following a line of thought that came about in FHIR facility side when thinking how do the settings work for multiple facility ids on one machine. Theres a few new ones like this in the migration like facility rateLimiter settings. 

I guess this method means it can be managed by central server, im not attached tho was really a play

No one uses omniserver right now tho.

dannash100:
@passcod u right tho wud make more sense to have a fact gui on facility side ay, no need to manage on central, i guess then it cud be a bit confusing.

PR #10254, opened 2026-07-09
ci(cd): build Windows-native VHDX and pack as OCI artifact

passcod:
### Changes

Alongside the linux container images, build **Windows-native** Tamanu VHDX disk images and pack them as **OCI artifacts** in `ghcr.io`, folded into the existing **multi-arch manifest**.

- **Servers** (`central`, `facility`): reduced to the target package with the same `docker-build.sh` the linux image uses — native `npm install` on a Windows runner, so `node_modules` are Windows-native — then the Windows Node runtime is embedded and the tree copied into a fresh **NTFS VHDX** (`scripts/build-windows-vhdx.ps1`, diskpart + robocopy).
- **Frontends** (`web`, `patient-portal`): built via the same script, packed into a self-contained VHDX embedding the Windows **Caddy** binary + Caddyfile, mirroring the linux frontend container.
- Each VHDX is zstd-compressed and `oras push`ed with a **BES-specific config media type** carrying a **runtime config blob** (Docker-style `WorkingDir`/`Env`/`Entrypoint`/`Cmd`/`ExposedPorts`, plus a `PATH` for the embedded runtime), and appended to the multi-arch index with an **`unknown`/`unknown` platform** so container runtimes never select it while a purpose-built BES client discovers it by artifact type.

`docker-build.sh` and `precompress-assets.sh` are now cross-platform (linux + Windows/Git Bash); precompression uses Node's built-in zstd/brotli/gzip.

**Build gating:** autodeploys build **linux/arm64 only**; **Build all images** (`%allimages`), a `%arch=amd64` deploy, or a release builds the full set — linux amd64 + arm64 and the Windows VHDX.

_Follow-up: confirm the VHDX boots/attaches and Tamanu runs on a target Windows host._

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [x] Build images only (don't deploy) 
- [x] Build all images (amd64 + Windows VHDX; default is arm64 only) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KCxZSrCdwVFsi35gZ9Pput


passcod:
## Windows VHDX experiment — `ntfs-compression` (aborted: impractically slow)

- Filesystem: **NTFS** | Compression: **true** (transparent, compress-on-write) | copy-based
- Runs: [#70968-ish](https://github.com/beyondessential/tamanu/actions/runs/28994899349) and [rerun](https://github.com/beyondessential/tamanu/actions/runs/28996121611) — both cancelled

**Finding: NTFS transparent compression is not viable for this workload.**

- First attempt: the `Build VHDX` step ran **21 min+** and never finished. Root cause was a redundant post-copy `compact /C /S` pass re-walking the whole tree; removed it.
- Second attempt (compress-on-write only, via the volume root's compressed attribute inherited by `robocopy`-written files): `Build VHDX` **still ran ~14 min and hadn't finished** when cancelled — vs **1:22** for the uncompressed baseline copy.

The bottleneck is per-file NTFS LZNT1 compression across `node_modules`' ~100k+ tiny files. And the payoff would be marginal: the artifact is already `zstd`-compressed **>10×** at upload (baseline 900 MiB raw → 83 MiB uploaded), so filesystem-level compression mostly duplicates work that `zstd` does far faster on the wire/at rest.

**Verdict:** skip NTFS compression. Size wins come from (a) `zstd` on the artifact (already in place) and (b) reclaiming dynamic-VHDX slack via compaction — evaluated next as the `optimize` experiment.

Proceeding to the ReFS experiment.


---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
## Windows VHDX experiment — consolidated scoreboard & recommendation

All variants build the two server packages (central/facility) and pack the VHDX as a BES OCI artifact (`application/vnd.au.bes.tamanu.windows-vhdx.v1+zstd`) folded into the multi-arch index with an `unknown/unknown` platform (never selected by a container runtime). Numbers are for `central` (facility tracks within a few %).

| # | Variant | VHDX raw (MiB) | Uploaded zst (MiB) | CD→upload | Notes |
|--|--|--:|--:|--:|--|
| 1 | **baseline** — NTFS, copy from linux image | **900** | **83** | 21:03 | reference; tree already pruned by `docker-build.sh` |
| 2 | direct-build — in-place on VHDX, full workspace | 1860 | 167 | **10:57** | parallel to linux ⇒ ½ wall-clock, but full-workspace deps |
| 3 | ntfs-compression | — | — | — | ✗ aborted: `Build VHDX` >14 min (per-file LZNT1 over ~100k files). `zstd` already gets >10× |
| 4 | refs | 1412 | 104 | 19:22 | larger artifact, no speed gain |
| 5 | direct-build + optimize | 1860 | 167 | 9:00 | compaction reclaimed **0** — `Optimize-VHD -Mode Full` present, "reclaimed 0.0 MiB" (npm cache is off-volume; deleted blocks aren't zeroed) |
| 6 | direct-build-reduced — **in-place** reduction | **2308** | **254** | 12:53 | ✗ **worse**: a dynamic VHDX keeps written-then-deleted blocks (full install → prune) and they're unzeroed, so uncompactable |
| 7 | direct-reduced-**copy/stripped** — reduce off-VHDX, copy into fresh VHDX | ~2180 | see auto-report | ~13–15 | correct approach (no in-place churn); size still above baseline — see below |

### Actual tree contents (measured, not inferred)

For `direct-reduced-stripped`, `du` of the built tree:

- **total: 1393 MiB**, of which **`node_modules`: 927 MiB**
- **node-gyp build intermediates: none** — `find` for `*.pdb/*.obj/*.iobj/*.ipdb/*.lib/*.exp` matched nothing; the strip reclaimed **0 MiB**. (I had hypothesised these dominated — that was wrong.)
- Largest `node_modules` entries: `@mui` 88, `ace-builds` 55, `@opentelemetry` 49, `hermes-compiler` 47, `@react-native` 40, `@napi-rs` 37, `lucide-react` 35, `@swc` 26, `typescript` 23 … and `jest`/`@playwright`/`cypress`/`@artilleryio`/`metro-*` are present.

**So the real bloat is unpruned dependencies, not build cruft.** Those are frontend (`@mui`, `ace-builds`, `recharts`, `lucide-react`), mobile (`react-native`, `hermes-compiler`, `metro`), and dev (`jest`, `playwright`, `cypress`, `artillery`) deps that a server doesn't need. The reduction removed the *packages* but the hoisted `node_modules` still carries their deps — i.e. the Windows path did **not** prune to the production server set the way the linux `docker-build.sh` does (that's why baseline is 900 and this is ~2180). The lever to close the gap is fixing the production prune, not stripping files.

### Findings

- **Never build in-place on a dynamic VHDX.** Any write-then-delete churn (npm temp, reduction) stays allocated and unzeroed; compaction can't reclaim it (confirmed 0 even with `Optimize-VHD`). Build the tree off-VHDX, then copy the *final* tree into a *fresh* VHDX.
- **NTFS, not ReFS** (ReFS: bigger artifact, no gain).
- **Rely on `zstd`, not filesystem compression** (NTFS compression is impractically slow here; `zstd` already >10×).
- **Skip compaction** — 0 reclaim on the runners.
- **Legacy tar derives from the built tree with no second `npm install`** — tar+zstd the release tree (VHDX re-mount was unreliable on GH runners across three approaches; the workspace tree is byte-identical, and it's what a release path already has). This is the transition win: one Windows build feeds both the OCI VHDX and the legacy `-windows.tar.zst`.
- **Speed vs size is a real trade-off, not a free win:** source/direct builds ~halve wall-clock (decoupled from the ~14-min linux image build) but produce a **larger** Windows-native artifact than the linux-derived baseline — *and that gap is unpruned deps, addressable by matching the linux production prune*, not an inherent Windows tax as I'd guessed.

### Recommendation

- Adopt **NTFS + build-off-VHDX-then-copy-into-fresh-VHDX** as the VHDX build; keep `zstd`; drop NTFS-compression, ReFS, and compaction.
- **Follow-up to reach baseline size:** make the Windows reduction prune `node_modules` to the production server set (as the linux image does) — the measured `node_modules` still contains frontend/mobile/dev deps.
- **Transition:** derive the legacy `-windows.tar.zst` from the built tree (no second `npm install`), keeping the VHDX build on the hot/tested path.

### Before merge

This is an experimental branch: the CI-muting (`workflow_dispatch`-only on `ci`/`e2e`/`dast`/`non-deterministic`/`review-hero`/`auto-fix`) and the experiment scaffolding (`.github/vhdx-experiment.env`, the report/metrics wiring, the aborted-variant knobs) must be reverted/cleaned before this could be a real change.


---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
### Correction to the scoreboard — `direct-reduced-stripped` regressed on both axes

Final measured numbers (central), which are worse than I projected above:

| | VHDX raw | Uploaded zst | CD→upload | of which `compress` (zstd -19) | legacy tar.zst | legacy derive |
|--|--:|--:|--:|--:|--:|--:|
| direct-reduced-stripped | 2180 MiB | **266.9 MiB** | **21:09** | **11:50** | 254 MiB | 5:57 |

So this variant is **not** the "best of both" — it's closer to the worst of both:

- **Larger than plain direct-build** (267 vs 167 MiB zst). The ported `docker-build.sh` reduction *injects build-tooling and does a full-workspace install first*, and (as the content breakdown showed) never pruned the hoisted `node_modules` back down — so it ends up bigger than the simpler exp-2 direct-build, which at least dropped `packages/mobile`.
- **No wall-clock win** (21:09 ≈ baseline's 21:03). The decoupling from the linux build was erased by `zstd -19` taking **~12 min** on the 2.1 GiB image. Compression cost scales with image size, so a bloated image is slow to pack regardless of how it was built.

**Revised takeaways:**
- The real speed result stays **plain direct-build** (167 MiB, 10:57) — ~½ wall-clock at ~2× baseline size.
- **baseline** remains smallest (83 MiB) and is the reference.
- My "reduction" as implemented is **counterproductive** — it must actually prune `node_modules` to the production server set (what linux `docker-build.sh` achieves) to be worthwhile; right now it only adds a full install.
- **`zstd -19` is a first-order cost** for large VHDXs (~12 min for 2.1 GiB). Any large-image variant should drop the zstd level (e.g. -10, as the linux server tar uses) or the packing time dominates.

Net recommendation is unchanged in principle (build off-VHDX → copy into a fresh VHDX; NTFS; zstd not FS compression; skip compaction; derive the legacy tar from the built tree) but the **size/speed win depends entirely on getting the production prune right** — without it, the direct/reduced path loses on both size and time. This closes the experiment series.


---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
## Experiment: `direct-reduced-pruned` — the reduction now actually prunes

The prior `direct-reduced-*` runs were building a Windows-native tree but **never
shrinking it** — the `remove_irrelevant_packages` port was silently a no-op.
Diagnosed from the run's own logs (`node_modules` stuck at 927 MiB, 2nd
`npm install` still auditing ~2974 packages) and fixed.

### Root cause (two bugs, stacked)

1. **jq `-= $unwanted` is a no-op.** `docker-build.sh` uses `jq --slurpfile unwanted … '.workspaces.packages -= $unwanted'`, which wraps the file's array in an extra array so nothing is subtracted. Production gets away with it because it runs under `NODE_ENV=production` (dev deps never install) and reconciles against the removed dirs. Our path does a **full** install (so `prepare`/patch-package runs) then `npm prune --omit=dev`, so the workspaces list must be *genuinely* trimmed → `-= $unwanted[0]`.
2. **Windows path mangling (the real killer).** On the runner, `list-packages.mjs` emits **backslash-separated** paths (`globSync` returns `packages\web`, and its forward-slash regex cleanup misses it) plus a stray CR. So `rm -rf` and the workspaces diff both operated on `packages\web\r`, which matches nothing → dirs kept, workspaces untrimmed. Fix: normalise each entry to posix `packages/x` before diffing.

### Result (commit `942229f`, NTFS, zstd -19, `#deployopt %imagesonly`)

2nd install now reports **`removed 1400 packages`** (was a no-op) → prune → tree more than halves.

| package | tree | node_modules | VHDX raw | VHDX zst (registry) | legacy tar | total (CD→upload) |
|---|---|---|---|---|---|---|
| **central** | 573 MiB | 434 MiB | 932 MiB | **95.1 MiB** | 88.3 MiB | **11m35s** |
| **facility** | 478 MiB | — | 804 MiB | **86.7 MiB** | 81.0 MiB | ~11m36s |

**central phase breakdown:** setupFetch 16s · npmInstall 317s · embedNode 7s · buildVhdx 71s · compress 260s · upload 7s

### vs the broken `direct-reduced-stripped` (central)

| metric | stripped (no-op prune) | **pruned (fixed)** | Δ |
|---|---|---|---|
| tree | 1393 MiB | 573 MiB | −59% |
| node_modules | 927 MiB | 434 MiB | −53% |
| VHDX raw | 2180 MiB | 932 MiB | −57% |
| VHDX zst | 267 MiB | 95.1 MiB | −64% |
| total time | 1471s | 695s | −53% |

### Takeaway

A **Windows-native** VHDX (real `npm install` on Windows) now lands at **~95 MiB compressed / 932 MiB raw**, essentially matching the linux-copied plain baseline (~83 MiB / 900 MiB) — and cheaper zstd (smaller input) cut compress time from ~11m to ~4m, halving the whole build to ~11.5m. This is the best result of the series: full Windows-native validity at baseline size on the hot-tested path, plus the legacy tar (~88 MiB) derived from the same finished tree in one build.


---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
## Consolidation + native frontend VHDXs — green

Run for `36203da` passed end-to-end, including the change with production-Docker blast radius:

- **All 8 linux image jobs succeeded** (central/facility/frontend/patient-portal × amd64/arm64) with the refactored cross-platform `docker-build.sh` + `Dockerfile` — so routing the Windows builds through the shared script did **not** regress the production image build.
- **Server VHDXs** (central/facility) now build via `bash scripts/docker-build.sh` — the ~90-line inline reduction duplicate is gone.
- **Frontend VHDXs** (web/patient-portal) build **Windows-native in parallel with the linux images** (needs only `config`), embedding the Windows Caddy binary + Caddyfile — no copy-from-linux, no wait on the image job.
- Precompression uses Node's built-in zstd/brotli/gzip (no CLI dependency on any platform; `apk add zstd brotli` dropped from the Dockerfile).
- All 4 multi-arch jobs appended the VHDX entries.

### Native frontend VHDX sizes

| package | VHDX raw | compressed (registry pull) | npm build+bundle | total (CD→upload) |
|---|---|---|---|---|
| **patient-portal** | 164 MiB | **17.6 MiB** | 352s | 7m13s |
| **web (frontend)** | — | **31.4 MiB** | ~5–6m | ~7m |

Both run concurrently with the linux images, so they add no serial time to CD.

### Landing next (commit `5d9c526`)

- **Runtime config blob** on each VHDX artifact (Docker-image-style `WorkingDir`/`Env`/`Entrypoint`/`Cmd`/`ExposedPorts`, BES config media type) — verifying on the current run.
- Fixed a cosmetic metrics bug where frontend jobs (no embed-node step) reported garbage `embedNode`/`buildVhdx` phase times.


---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
## Runtime config blob — attached & verified

Each Windows VHDX artifact now carries an OCI **config blob** describing how to run it, mirroring the linux Docker image config. Verified from the registry (`oras manifest fetch`):

```json
"config": {
  "mediaType": "application/vnd.au.bes.tamanu.windows-vhdx.config.v1+json",
  "digest": "sha256:23cdf4f3…", "size": 635
}
```

`oras push --config <blob>:<bes-type>` works and ghcr accepts the custom media type; the VHDX layer keeps its `…windows-vhdx.v1+zstd` type, and the multi-arch index entry still gets the `unknown/unknown` platform — so container runtimes never treat it as a runnable image, while a BES client reads the blob to learn how to run it.

**Server blob** (mirrors the Docker image — `docker-entrypoint.sh`'s effective run, not pm2):
```json
{
  "WorkingDir": "release-v2.60.0/packages/central-server",
  "Env": ["NODE_ENV=production", "PORT=3000", "NODE_CONFIG_DIR=config/"],
  "Entrypoint": ["node", "--import", "tsx", "app"],
  "Cmd": ["serve"],
  "ExposedPorts": {"3000/tcp": {}}
}
```

**Frontend blob** (mirrors the caddy container):
```json
{
  "WorkingDir": "tamanu-frontend-static-v2.60.0",
  "Env": ["NODE_ENV=production"],
  "Entrypoint": ["caddy"],
  "Cmd": ["run", "--config", "Caddyfile", "--adapter", "caddyfile"],
  "ExposedPorts": {"3000/tcp": {}}
}
```

`vhdx.rootDir` records the top-level directory the bundle sits under on the mounted volume (the analogue of the container's `/app`). Confirming these exact values land on the current run.


---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+      # Reduce to the target server via the SAME cross-platform docker-build.sh
+      # the production Docker image uses.
+      - name: Build reduced source (Windows-native)
+        shell: bash
+        run: |
+          set -euxo pipefail
+          pkg='${{ matrix.package.path }}'   # central-server | facility-server
+          cd "release-v${VERSION}"
+
+          # Point pm2's cwd at the target package (mirrors cd-package-servers),
```
passcod: This is still referencing pm2 even though the vhdx doesn't use it?

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+          $root = "release-v$env:VERSION"
+          if (Test-Path $root) { Remove-Item -Recurse -Force $root }
+          New-Item -ItemType Directory -Force -Path $root | Out-Null
+          # robocopy is native (handles the repo's symlinks); exclude .git,
+          # node_modules (installed fresh below), and the dest dir itself.
+          robocopy . $root /E /XD .git node_modules "$root" oras-bin /XF *.vhdx *.zip /R:2 /W:2 /NFL /NDL /NJH /NJS /NP | Out-Null
+          if ($LASTEXITCODE -ge 8) { throw "robocopy staging failed ($LASTEXITCODE)" }
+          $global:LASTEXITCODE = 0
+
+      # Reduce to the target server via the SAME cross-platform docker-build.sh
```
passcod: Remove or reword, since this will become A production build

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
     outputs:
       version: ${{ steps.version.outputs.version }}
 
+  # Windows-native release VHDXs, packed as OCI artifacts in ghcr.io and folded
+  # into the multi-arch manifest we already generate. Built from source on a
+  # Windows runner (real npm install / native build) in parallel with the linux
```
passcod: remove or reword comment: don't explain what is already obvious from the code

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+          # per-package *.config.* files.
+          sed -E "s|const cwd =.+|const cwd = 'packages/${pkg}';|" \
+            "packages/${pkg}/pm2.config.cjs" > pm2.config.cjs
+
+          # No DOCKER_BUILD, so the /app cache and self-delete are skipped.
+          bash scripts/docker-build.sh "$pkg"
+
+      - name: Embed Windows Node runtime
+        shell: bash
+        run: scripts/embed-node-runtime.sh win-x64 "release-v${VERSION}"
```
passcod: the release-v... folder here feels completely unnecessary: put it in an unversioned node folder or the root or something

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+        with:
+          package: ${{ matrix.package.name }}
+          package-path: ${{ matrix.package.path }}
+          kind: server
+          root-dir: release-v${{ env.VERSION }}
+          version: ${{ env.VERSION }}
+          sha: ${{ needs.config.outputs.sha }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  # Frontends (web / patient-portal): static builds, so no reduction — build the
```
passcod: again, remove this as it describes what the code does

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: '.node-version'
+          cache: npm
+
+      - name: Extract version
+        shell: bash
+        run: echo "VERSION=$(jq -r .version package.json)" >> "$GITHUB_ENV"
+
+      # Build the frontend natively with the same docker-build.sh the linux image
```
passcod: useless comment

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+      # uses (npm build + precompress; precompress-assets.sh uses Node's built-in
+      # zstd/brotli/gzip, so no extra CLI tooling is needed).
+      - name: Build frontend (Windows-native)
+        shell: bash
+        run: bash scripts/docker-build.sh '${{ matrix.package.path }}'
+
+      # Assemble the self-contained bundle: dist under app/, the Windows Caddy
```
passcod: useless comment

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+      # binary, a bundle-relative Caddyfile, and a launcher.
+      - name: Assemble bundle (dist + Windows Caddy)
+        shell: bash
+        run: |
+          set -euxo pipefail
+          stage="tamanu-${{ matrix.package.name }}-static-v${VERSION}"
+          mkdir -p "$stage/app"
+          cp -r "packages/${{ matrix.package.path }}/dist/." "$stage/app"
+
+          # Embed the Windows Caddy binary (matches the container's embedded caddy).
```
passcod: useless comment

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+            "https://github.com/caddyserver/caddy/releases/download/v${caddy_version}/caddy_${caddy_version}_windows_amd64.zip"
+          unzip -o caddy.zip caddy.exe -d "$stage"
+          rm caddy.zip
+
+          # Same Caddyfile as the container, but serve from the bundle-relative
+          # ./app rather than the absolute /app.
+          sed -E 's|root \* /app|root * app|' \
+            "packages/${{ matrix.package.path }}/Caddyfile.docker" > "$stage/Caddyfile"
+
+          # Launcher so an operator/service can start it from the mounted VHDX.
```
passcod: why is this needed when the config blob exists

.github/workflows/cd.yml
```diff
@@ -251,10 +251,202 @@ jobs:
...
+          sha: ${{ needs.config.outputs.sha }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
   multi-arch:
     needs:
       - config
       - images
+      - windows-vhdx
+      - windows-vhdx-frontend
+    # always() so a failure/skip of the windows-vhdx jobs never blocks publishing
```
passcod, line 425: that was useful experimentally but now we're moving into prod we shouldn't anymore: instead we should have the cd-config only build linux-arm64 for autodeploys, opt into the full builds with an option checkbox, and require every image for releases

.github/workflows/cd.yml
```diff
@@ -269,12 +461,25 @@ jobs:
...
         run: |
           set -x
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+          echo "${{ secrets.GITHUB_TOKEN }}" | oras login ghcr.io -u '${{ github.actor }}' --password-stdin
           image="ghcr.io/beyondessential/tamanu-${{ matrix.repo }}"
-
+          sha="sha-${{ needs.config.outputs.sha }}"
+          vhdx_type="application/vnd.au.bes.tamanu.windows-vhdx.v1+zstd"
+
+          # The Windows VHDX artifact exists for the servers (windows-vhdx*) and
```
passcod, line 453: again, not anymore

.github/workflows/cd.yml
```diff
@@ -269,12 +461,25 @@ jobs:
...
+          # the frontends (windows-vhdx-frontend), but only when that job
+          # succeeded. Detect it rather than assume it.
+          vhdx_ref=""
+          if oras manifest fetch --descriptor "$image:$sha.windows-vhdx" >/dev/null 2>&1; then
+            vhdx_ref="$image:$sha.windows-vhdx"
+          fi
+
+          # 1) Build the normal linux multi-arch index (unchanged behaviour).
```
passcod: useless comment

.github/workflows/cd.yml
```diff
@@ -283,15 +488,53 @@ jobs:
             || true
           }
 
-          retag sha-${{ needs.config.outputs.sha }}
+          # 2) Append the Windows VHDX as an extra index entry. It carries the
```
passcod: reword to only explain why and not all this useless fluff

Dockerfile
```diff
@@ -6,6 +6,9 @@ COPY package.json package-lock.json COPYRIGHT LICENSE-GPL LICENSE-BSL ./
 COPY patches/ patches/
 
 FROM base AS build-base
+# docker-build.sh keys Docker-only behaviour (the /app npm cache carried between
+# stages, self-deleting the script) off this flag.
+ENV DOCKER_BUILD=1
```
passcod: this adds this env permanently into the image rather than only for the script call

Dockerfile
```diff
@@ -36,9 +39,9 @@ ARG PACKAGE_PATH
 # copy all packages
 COPY packages/ packages/
 
-# do the build, which will also reduce to just the target package
+# do the build, which reduces to just the target package and prunes to a
+# production dependency set (docker-build.sh's build_server runs npm prune)
```
passcod: useless comment, revert

Dockerfile
```diff
@@ -68,7 +71,7 @@ ENV NODE_CONFIG_DIR=/config:/app/packages/${PACKAGE_PATH}/config
 ## Build the frontend packages (web or patient-portal)
 FROM build-base AS build-frontend-base
 ARG PACKAGE_PATH
-RUN apk add zstd brotli
+# precompress-assets.sh uses Node's built-in zlib, so no zstd/brotli CLIs needed.
```
passcod: useless comment, documents something that no longer exists

scripts/build-windows-vhdx.ps1
```diff
@@ -0,0 +1,97 @@
+<#
+.SYNOPSIS
```
passcod: useless extensive description for a script that will only be used by this machine process it is custom-made for

scripts/docker-build.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 
-### This expects to be run in the production docker build in /Dockerfile.
+### Reduce a Tamanu workspace to a single deployable package (server or frontend).
```
passcod: revert to the previous simpler comment, add like two or three words for the new usage

scripts/docker-build.sh
```diff
@@ -11,17 +18,21 @@ common() {
...
-  npm install --no-interactive --package-lock
+  # install dependencies. Force bash as the lifecycle script shell so the
+  # shell-syntax `prepare` hook (patch-package) parses on Windows, where npm
+  # defaults to cmd.exe. bash is present in the Docker build image and Git Bash.
+  npm install --no-interactive --package-lock --script-shell bash
 }
 
 remove_irrelevant_packages() {
-  # remove from npm workspace list all packages that aren't the ones we're building
+  # remove from the npm workspace list every package that isn't the one we're building
```
passcod: useless rewording, revert

scripts/docker-build.sh
```diff
@@ -33,10 +44,12 @@ remove_irrelevant_packages() {
       exit 1
     fi
 
-  # erase from the package.json
+  # erase from the package.json. --slurpfile wraps the file's array in an extra
+  # array, so subtract [0] to actually trim the workspace entries (otherwise the
```
passcod: useless rewording, revert. the fix is useful, but this comment now documents the negative space of a behaviour that no longer exists

scripts/docker-build.sh
```diff
@@ -75,23 +88,24 @@ build_server() {
     package.json.working > package.json
   rm package.json.working
 
-  # remove build dependencies
-  npm install --no-interactive --package-lock
+  # remove build dependencies, then prune to a production dependency set
```
passcod: useless comment change, revert. the original comment existed because it wasn't obvious what the npm install did, the added prune is by itself obvious

scripts/precompress-assets.sh
```diff
@@ -1,16 +1,37 @@
...
-); do
-  echo "precompressing $file"
-  zstd -f -T0 -19 -o $file.zst $file
-  gzip -f -9 -k $file
-  brotli -f -Z -k $file
-done
+
+# Precompress static assets so Caddy can serve them with `precompressed zstd br
+# gzip`. Uses Node's built-in zlib (zstd, brotli and gzip are all available in
+# the Tamanu Node version), so no external compression CLIs are required on any
```
passcod: comment documents things that no longer exist

.github/workflows/cd.yml
```diff
@@ -147,6 +163,11 @@ jobs:
       go-down: ${{ steps.config.outputs.go-down == 'true' }}
       go-mobile: ${{ steps.config.outputs.go-mobile == 'true' }}
 
+      # whether to build the full image set (amd64 + Windows VHDX), vs arm64 only
+      all-images: ${{ steps.scope.outputs.all }}
+      # platform matrix for the images job
+      platforms: ${{ steps.scope.outputs.platforms }}
+
```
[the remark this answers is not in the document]
passcod: Good catch. Fixed in 53a03e6: `check-images-exist` now checks the arch that distinguishes the requested scope (`amd64` when the full set is requested, else `arm64`), so a widened scope (`%allimages`/`%arch=amd64`) after an arm64-only run still rebuilds instead of short-circuiting on the existing `sha-<sha>` tag.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #10263, opened 2026-07-10
feat(medication): TAM-2044: Alow edit prescription

chris-bes:
### Changes

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [x] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/database/src/migrations/1783404000001-backfillDispensedDetailsFromPrescriptions.ts
```diff
@@ -0,0 +1,47 @@
+import { QueryInterface } from 'sequelize';
+
+// Backfill the dispensed-detail columns on historical medication_dispenses from the prescription
+// each dispense was made against (via its pharmacy_order_prescription). Historical dispenses were
+// all dispensed as prescribed, so modified_* columns stay null.
+export async function up(query: QueryInterface): Promise<void> {
+  await query.sequelize.query(`
```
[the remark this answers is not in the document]
chris-bes: this is fine as run migrations when deployment is down

packages/ui-components/src/components/Field/NumberField.jsx
jaskfla, line 1: Thanks for linking me here!

Sorry to be a bit of a bummer, but I will voice a strong opposition to this in favour of the native spin buttons:

- Even though this implementation uses semantic `<button>`s, it breaks convention and user expectation for keyboard navigation. Now, in order to tab from a `<NumberField>` to the next field or element, it takes _three_ tabs instead of one, because there are two spin buttons that have been injected into the tab order.
- <kbd>↑</kbd> and <kbd>↓</kbd> handlers are missing.
- The custom buttons don’t respect the `<input>`’s `min`, `max`, `step` or `list` attributes.
- Custom handling present for a `disabled` `<NumberField>`, but not `readonly`
- ~~This component is expected to render in a `<form>`, in which case `<button>`s normally default to `type="submit"`.~~ Looks like agent addressed this, but no need to reinvent the wheel
- …probably other free behaviours I haven’t thought of (or even know about!)

Even though all of these are things that can be overcome, it really blows code out of proportion. Sometimes a custom element is truly warranted, but there needs to be a strong case for it. Otherwise it’s naïve to think we can to better than browser vendors imo.

Moot for this PR, but I’ll also take this opportunity to highlight `HTMLInputElement`s have [`stepUp`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp) and [`stepDown`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepDown) methods[^foot], which _do_ honour `min`/`max`/`step` attributes for that `<input>` instance. For the (rare!) case where we _really_ need to hijack control over the `<input>`’s value with event listeners, we should use `inputRef.stepUp()`/`inputRef.stepDown()`

[^foot]: Even then, it seems Firefox has been lagging in support

packages/web/app/components/Medication/ModifyPrescriptionModal.jsx
```diff
@@ -0,0 +1,373 @@
...
+  const retainedDoseAmount = useRef(null);
+
+  const defaultPharmacyNote = getTranslation(
+    'medication.modify.defaultPharmacyNote',
+    'This prescription has been modified by pharmacy when dispensing. Please check for details if unsure.',
+  );
+
+  const drugSuggester = useSuggester('drug', {
+    formatter: ({ name, id, ...rest }) => ({ ...rest, label: name, value: id }),
+  });
```
jaskfla, line 120: Bit of a micro-optimisation, but the `useSuggester` options object here can be defined at module-level to avoid reinstantiating a `new Suggester()` every render

hrazasalman (approved):

julianam-w (approved):

PR #10276, opened 2026-07-12
fix(invoicing): TAM-6901: convert inpatient-bundled medication qty to dispensing units

passcod:
### Changes

Fixes a high-severity medication-invoicing overcharge (from the logic-bug audit, #10266).

In `Prescription.recalculateAndApplyInvoiceQuantity`, the normal path converts administered dose totals from dosing units to dispensing units:

```ts
marQty = Math.ceil(totalDosingAmount / unitConversion);
```

But the inpatient-fee-bundled branch (which re-computes the billable quantity as only the pre-admission doses) summed raw `doseAmount` values with **no `unitConversion` division and no ceil**. The result was in dosing units (e.g. mg) while the invoice item quantity is defined in dispensing units (e.g. tablets), and it was then added to `totalSentQty` (dispensing units) and written to the invoice item.

Example: a facility bundling medication into the admission fee; a drug with `unitConversion = 250` (mg/tablet) given as two 250 mg pre-admission doses should bill `ceil(500 / 250) = 2` tablets, but the invoice quantity became `500` — a 250× overcharge.

- `Prescription.ts` — apply the same `Math.ceil(sum / unitConversion)` conversion in the bundled branch as the normal path.

The existing `EncounterInvoice.test.js` only exercises `unitConversion = 1` for the bundled case, which masks this; a bundled case with `unitConversion != 1` has been added.

### Auto-Deploy

- [ ] **Deploy** 

### Tests

- [ ] **Run E2E tests** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37796,6 +37796,7 @@
       "devDependencies": {
         "@casl/ability": "^6.8.0",
         "@tamanu/build-tooling": "*",
+        "@tamanu/fake-data": "*",
         "@types/bcrypt": "^5.0.2",
         "@types/express": "^4.17.13",
         "@types/jsonpath": "^0.2.4",
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -31,6 +31,7 @@
   "devDependencies": {
     "@casl/ability": "^6.8.0",
     "@tamanu/build-tooling": "*",
+    "@tamanu/fake-data": "*",
     "@types/bcrypt": "^5.0.2",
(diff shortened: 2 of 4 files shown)


passcod:
Need a product call here

tcaiger (approved):
Good catch - thanks

PR #10340, opened 2026-07-13
fix(centralServer): allow template name conflicts with historical templates

jaskfla:
### Changes

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/central-server/app/admin/template.js
+++ b/packages/central-server/app/admin/template.js
@@ -18,11 +18,16 @@ const checkUniqueName = asyncHandler(async (req, res, next) => {
     return;
   }
 
-  const conflictingRecords = await req.models.Template.count({
-    where: { name, id: { [Op.ne]: id ?? null } },
+  const conflict = await req.models.Template.findOne({
+    where: {
+      name,
+      id: { [Op.ne]: id ?? null },
+      visibilityStatus: VISIBILITY_STATUSES.CURRENT,
+    },
+    attributes: ['id'], // Arbitrary projection, just checking existence
   });
 
-  if (conflictingRecords) {
+  if (conflict) {
     throw new ValidationError('Template name must be unique');
   }
 
--- a/packages/central-server/app/admin/template.js
+++ b/packages/central-server/app/admin/template.js
@@ -18,7 +18,7 @@ const checkUniqueName = asyncHandler(async (req, res, next) => {
     return;
   }
 
-  const conflict = await req.models.Template.findOne({
+  const conflictingRecord = await req.models.Template.findOne({
     where: {
       name,
       id: { [Op.ne]: id ?? null },
@@ -27,7 +27,7 @@ const checkUniqueName = asyncHandler(async (req, res, next) => {
     attributes: ['id'], // Arbitrary projection, just checking existence
   });
 
-  if (conflict) {
+  if (conflictingRecord) {
     throw new ValidationError('Template name must be unique');
   }
 
--- a/packages/central-server/app/admin/template.js
+++ b/packages/central-server/app/admin/template.js
@@ -1,8 +1,9 @@
 import express from 'express';
 import asyncHandler from 'express-async-handler';
-import { Op, ValidationError } from 'sequelize';
+import { Op } from 'sequelize';
 
 import { VISIBILITY_STATUSES } from '@tamanu/constants';
+import { EditConflictError } from '@tamanu/errors';
 import { simpleGetList, simplePost, simplePut } from '@tamanu/shared/utils/crudHelpers';
 
 export const templateRoutes = express.Router();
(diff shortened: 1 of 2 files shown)


packages/central-server/app/admin/template.js
jaskfla, line 1: Minor note: `findOne` short-circuits whereas `count` does not. (In absence of `.exists()` Sequelize method 🙄)

More importantly: This is not watertight. It still lets us take a historical template with an identical name as current one, and flip its visibility status back to current.

Points to consider:

- API allows a historical → current change via a PUT request, but there is no GUI in Tamanu that allows this;
- easy to enforce at DB level with `UNIQUE (name) WHERE visibility_status = 'current'`.

My question: **is it worth the migration** to add the partial unique index?

NavarroEmilioLuis: A couple of different notes:

1. findOne is 100% the right call.
2. Watertight is not always the right call, so this is good. From a Tamanu perspective, AFAIK our endpoints are usually not watertight and that is part of a flexible API design - in the future, this would allow us to reuse the endpoint and update the visibility status, so we only need to do frontend work.

I think this is almost a given for a facility server endpoint. On central we might want to be more strict, but then again, this needs an authenticated client deliberately sending manual requests, which is not only unlikely, but in this scenario pretty harmless:
-  That user _must_ know what they are doing (fun fact, I actually used this technique a couple of years ago for support which we couldn't remote into the server).
- If it's meant to be an attack vector it's... funny.

3. Enforcing at a DB level is not always the right call, so this is good. This is usually a grey area? It depends on a lot of things we probably don't need to get into. I will only list the cons I see for this approach:
- DB level logic is usually "hidden" whereas code logic is right there on the endpoint.
- This has added complexity: we might need to cleanup dupes before adding the index.
- More constraints means less flexibility and future changes would always need a code change.
- The constraint is mixed and depends on two variables rather than one, which isn't bad by itself, but it leads to...

4. This approach feels a bit off. I don't know the business logic requirements, nor what these template records are or why we hit such scenario. I think that uniquely identifying by name shouldn't depend on visibility status, it's a bit of a weird state to be in: why do we want to keep the historical template (not medical data?) why not just overwrite it? If we cannot turn a historical into a current one via the GUI, then that operation isn't meant to be and I assume historical won't show up, so why don't we just get rid of the record?

The DB state could have 30 templates named "XYZ" historical and just one current, which sounds like just version control, except we don't care about the other 30 templates (maybe we do, I really don't know enough).

TLDR: nah

packages/central-server/app/admin/template.js
```diff
@@ -18,11 +18,16 @@ const checkUniqueName = asyncHandler(async (req, res, next) => {
     return;
   }
 
-  const conflictingRecords = await req.models.Template.count({
-    where: { name, id: { [Op.ne]: id ?? null } },
+  const conflict = await req.models.Template.findOne({
```
NavarroEmilioLuis: This variable name is thought provoking 😂  alas I'd favor something simpler (even if it has longer length)

NavarroEmilioLuis (commented):
I'll start with a disclaimer: because we're not doing much reviews, I really enjoyed doing this one and ended up commenting more than what was really asked, so feel free to disregard anything as this is good enough as is.

The last overall review comment in the same light would be: because the PR doesn't have a linked card and/or intent, I can't suggest a different pattern for the fix.

Non AI summary: if you have minutes to spare and consider, read this review, otherwise go ahead and merge

packages/central-server/app/admin/template.js
jaskfla: > because the PR doesn't have a linked card and/or intent

Sorry, I should’ve linked [the release issue](https://bes-international.slack.com/archives/C03KJSEPV9A/p1783908705513589)!

> This approach feels a bit off.

This raises a good point. What’s going on here is that when you “delete” a template it actually sends `PUT` request and updates the visibility status.

Since this isn’t actually a regression—we just caught this UX issue in this release—I might bring it to PO for an opinion. Not too much harm in deferring it to a future release.

jaskfla: [Resolved in Slack](https://bes-international.slack.com/archives/C03KJSEPV9A/p1783985450703829?thread_ts=1783908705.513589&cid=C03KJSEPV9A) to keep the unusual delete behaviour (i.e. change `visibitliy_status` instead of `delete_at`); and stick with the code change proposed here

Will add an inline comment to explain why the uniqueness is based on visibility status, though

packages/central-server/app/admin/template.js
jaskfla: d6783bc03b6208f8f88ae6378ba1a66bf42d6af7

PR #10402, opened 2026-07-15
perf(labs): no-issue: exclude sensitive labs with NOT EXISTS anti-join in worklist query

dannash100:
### Changes

The lab worklist (`GET /api/labRequest`) is intermittently very slow on production facilities — measured on Palau BNH: bimodal at p50 24ms / p90 92ms but **p99 10.7s, max 10.9s**. Load is ruled out (every other endpoint is p99 ~448ms across 53k requests on the same box).

`EXPLAIN (ANALYZE, BUFFERS)` of the worklist count query pinned the cause: **the `sensitive_labs` CTE**. It is a single-reference CTE, so Postgres inlines it and pushes it into a nested loop, **re-evaluating the whole-table `lab_requests ⋈ lab_tests ⋈ lab_test_types` sensitive-lab scan once per matched row** (`Group … loops=11`, 318,707 buffers = 97% of the query; the `lab_requests` scan itself is only ~20ms / 7,542 buffers). 234ms × 11 ≈ 2.57s.

Fix (first pass): mark the CTE `MATERIALIZED` so it is computed **once** per query instead of per matched row (~2.5s → 32ms validated on the affected facility).

Fix (final): drop the CTE and its `LEFT JOIN` entirely and express the exclusion as a per-row `NOT EXISTS` anti-join inside the existing `makeFilter(!canListSensitive, …)`. This subsumes both follow-ups from the first pass: privileged users (`canListSensitive`) get **no sensitive machinery at all** (the filter is simply not added), and non-privileged users get an index probe on `lab_tests_lab_request_id` per candidate row instead of any whole-table `lab_tests` scan. Semantics are unchanged — a request is excluded when any of its tests has a sensitive type — and covered by the existing `should exclude sensitive lab requests` test.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

---
**Validated on the affected facility (Palau BNH) with `EXPLAIN (ANALYZE, BUFFERS)` of the count query — cost of the sensitive-labs exclusion:**

| | Inlined CTE (before) | `MATERIALIZED` (first pass) | `NOT EXISTS` (this PR) |
|---|---|---|---|
| Sensitive-side cost | ~234 ms **per matched row** (318,707 buffers) | ~234 ms once per query | **0.26 ms / 63 buffers** |

The anti-join plan is a `Parallel Hash Right Anti Join`: the sensitive `lab_test_types` set is hashed first (empty on this facility), so the `lab_tests` probe short-circuits after reading a single row — the 587k-row `lab_tests` table is never scanned. The remaining query time is the base worklist scan (`lab_requests` + encounter lookups), unrelated to the sensitive filter.

### DIFF
--- a/packages/facility-server/app/routes/apiv1/labs.js
+++ b/packages/facility-server/app/routes/apiv1/labs.js
@@ -144,12 +144,21 @@ labRequest.get(
   '/',
   asyncHandler(async (req, res) => {
     const {
-      models: { LabRequest },
+      models: { LabRequest, LabTestType },
       query,
       settings,
     } = req;
     req.checkPermission('list', 'LabRequest');
     const canListSensitive = req.ability.can('list', 'SensitiveLabRequest');
+    // With no sensitive test types (most deployments) the anti-join below is vacuous, so skip
+    // it. paranoid: false mirrors the raw filter, which doesn't exclude soft-deleted types.
+    const mustExcludeSensitive =
+      !canListSensitive &&
+      (await LabTestType.findOne({
+        where: { isSensitive: true },
+        attributes: ['id'],
+        paranoid: false,
+      })) !== null;
 
     const {
       order = 'ASC',
@@ -235,7 +244,7 @@ labRequest.get(
       ),
       makeDeletedAtIsNullFilter('encounter'),
       makeFilter(
-        !canListSensitive,
+        mustExcludeSensitive,
         `NOT EXISTS (
           SELECT 1
           FROM lab_tests
--- a/packages/facility-server/app/routes/apiv1/labs.js
+++ b/packages/facility-server/app/routes/apiv1/labs.js
@@ -304,8 +304,11 @@ labRequest.get(
         ${whereClauses && `WHERE ${whereClauses}`}
     `;
 
+    // MATERIALIZED is load-bearing: without it Postgres inlines this single-reference CTE and
+    // re-evaluates the whole-table sensitive-lab scan once per matched row (measured ~2.5s / 11
+    // rows on a prod worklist). Forcing a single computation keeps it to one scan.
     const queryCte = `
-      WITH sensitive_labs AS (
+      WITH sensitive_labs AS MATERIALIZED (
         SELECT lab_requests.id as id, TRUE as is_sensitive
         FROM lab_requests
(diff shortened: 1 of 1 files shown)


packages/facility-server/app/routes/apiv1/labs.js
```diff
@@ -234,7 +234,18 @@ labRequest.get(
...
+        `NOT EXISTS (
+          SELECT 1
+          FROM lab_tests
+          INNER JOIN lab_test_types
+            ON (lab_test_types.id = lab_tests.lab_test_type_id)
+          WHERE lab_tests.lab_request_id = lab_requests.id
+            AND lab_test_types.is_sensitive IS TRUE
+        )`,
+        () => {},
+      ),
```
rohan-bes, line 257: Oh this is clever, so now we only calculate this if the user doesn't have access to sensitive facilities? Maybe we could improve the `!canListSensitive` check to be something like?

```suggestion
      const sensitiveTestsExist = await LabTests.findOne({ isSensitive: true }) !== null;
...
      makeFilter(
        !canListSensitive && sensitiveTestsExist,
        `NOT EXISTS (
          SELECT 1
          FROM lab_tests
          INNER JOIN lab_test_types
            ON (lab_test_types.id = lab_tests.lab_test_type_id)
          WHERE lab_tests.lab_request_id = lab_requests.id
            AND lab_test_types.is_sensitive IS TRUE
        )`,
        () => {},
      ),
```

That way we wouldn't even include this filter if there are no sensitive tests (which I think would be quite common)

rohan-bes (approved):
Happy with this change but had a suggestion so that we don't use this filter at all on deployments with no sensitive tests

PR #10586, opened 2026-07-27
fix(sync): TAM-6962: record sync credentials offline, pull the settings PSK on sync

dannash100:
### Changes

Four parts of one problem, all TAM-6962: an upgrade step whose failure mode is "keep using the old thing" must not depend on the network, because the fallback it leans on is what 2.63 deletes.

**1. Forward-port of #10584**, which went to `release/2.61`. `provisionSyncUser` wrote nothing unless it could reach central, so any facility that upgraded on a flaky link is on the config fallback right now. Now an offline copy that can't fail, plus the existing central round-trip. Cherry-picked with `-x`, one test-file conflict resolved by pointing main's PSK test at `provisionStep`.

**2. The settings PSK pulls on sync rather than on upgrade.** Same shape, worse consequence: the old step needed central, caught, skipped, and retried only next upgrade, so removing the config fallback while a facility sits there costs it every secret setting for a whole release. `SyncTask` already needs central and runs on a schedule, so `pullSettingsPsk` runs after a completed session, gated on the PSK being absent and wrapped so it can't fail the sync. `1785200000000-provisionFacilitySettingsPsk.ts` is deleted. Env-credential k8s facilities are covered for the first time, since that step gated on `FACT_SYNC_EMAIL` which they don't have.

No offline copy of `crypto.settingsPsk` into the store, deliberately: the store wins over config, so it would pin a facility to its per-host value and the pull's `setIfAbsent` would no-op forever after.

**3. Logging** (was #10583, folded in). Nothing showed which deployments still resolve sync settings or the PSK from config, so the 2.63 removal would go in blind.

**4. The PSK is served to `kind: sync` accounts only** (was #10587, merged into this branch). `manage:all` alone doesn't establish the caller is a facility server: mobile and facility servers request the same `sync_client` device scope, `isMobile` arrives in the request body, and admin routes don't check scope at all. `kind` is the one part of an account central sets itself. That forced the credential swap onto sync as well (`convergeSyncUser`), because an un-converged facility authenticates as the human admin its credentials were copied from, and the gate alone would 403 exactly the facilities part 1 exists to serve.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


passcod:
I think this is conflating two auth related mechanisms where one is what is allowed to sync and the other is what is allowed to be a facility. The first covers all mobile devices but the second covers only essentially facility servers — I believe that the settings psk was meant to be only for facilities servers, not for mobile. So if we are doing this, then… like, only facility will be pulling the settings psk, well behaved mobile will not pull it, but because the authentication is riding the sync session and mobile is on the same session auth, that means that mobile is theory if they were hacked, could pull this settings psk.

Whichhich we don't want, I think. So we either need to change the security model and such that mobiles **can** pull the settings psk here, or we need to change the authentication for that route in particular so that only facility can access it.

dannash100:
> I think this is conflating two auth related mechanisms where one is what is allowed to sync and the other is what is allowed to be a facility. The first covers all mobile devices but the second covers only essentially facility servers — I believe that the settings psk was meant to be only for facilities servers, not for mobile. So if we are doing this, then… like, only facility will be pulling the settings psk, well behaved mobile will not pull it, but because the authentication is riding the sync session and mobile is on the same session auth, that means that mobile is theory if they were hacked, could pull this settings psk.
> 
> Whichhich we don't want, I think. So we either need to change the security model and such that mobiles **can** pull the settings psk here, or we need to change the authentication for that route in particular so that only facility can access it.

New branch for auth angle https://github.com/beyondessential/tamanu/pull/10587

passcod (approved):

PR #10648, opened 2026-07-31
fix(sync): no-issue: make streaming pull paging work

passcod:
### Changes

🤖 Streaming pull (`sync.streaming.enabled`, off by default) does not page. The stream route read its batch size and poll interval from `sync.databasePollBatchSize` / `sync.databasePollInterval`, but those settings live under `sync.streaming.*`, so both reads returned undefined: the page limit fell back to `Number.MAX_SAFE_INTEGER`, pulling the whole snapshot in one query — which is the thing streaming exists to avoid — and the two wait loops slept for ~1ms per iteration, polling the database in a tight loop while waiting for a snapshot that can take hours.

Behind that, the paging cursor was wrong on both ends. The dependency-ordered pull query orders by `(sort_order, id)` and expects a cursor carrying both parts, but the stream route and the streaming client each passed the bare snapshot id — the route throws on its second page, the client silently restarts from the beginning of the snapshot instead of resuming. The unbounded limit masked the route side by never needing a second page. The streaming client also wrote pulled records to the snapshot table without stripping `sortOrder`, which is not a column on it; the polling client deletes it, and the two paths had drifted apart.

The cursor format now has a single definition shared by the query, both pull routes and both clients, and the facility’s two pull paths share one record-preparation step so they cannot diverge again. Mobile keeps its own copy of the encoding, as it cannot depend on `@tamanu/database`.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/database/src/sync/snapshotCursor.ts
+++ b/packages/database/src/sync/snapshotCursor.ts
@@ -21,11 +21,24 @@ export const decodeSnapshotCursor = (cursor?: string | null): Partial<SnapshotCu
 
   // the cursor arrives as a query parameter, so a malformed one is a bad request rather than a
   // server fault, and it says which parameter to look at
+  let decoded;
   try {
-    return JSON.parse(atob(cursor));
+    decoded = JSON.parse(atob(cursor));
   } catch (error) {
     throw new InvalidParameterError(
       `fromId is not a valid pull cursor: ${(error as Error).message}`,
     );
   }
+
+  // Both parts go into a SQL comparison against (sort_order, id), so a cursor that decodes but
+  // carries the wrong types is still the client's mistake — checked here rather than left to fail as
+  // a query error, which reaches the client as a server fault and says nothing about the cause.
+  const { sortOrder, id } = decoded ?? {};
+  if (!Number.isInteger(sortOrder) || !['string', 'number'].includes(typeof id)) {
+    throw new InvalidParameterError(
+      'fromId is not a valid pull cursor: expected a sort order and a record id',
+    );
+  }
+
+  return { sortOrder, id };
 };
--- a/packages/database/src/sync/snapshotCursor.ts
+++ b/packages/database/src/sync/snapshotCursor.ts
@@ -1,0 +1,18 @@
+/**
+ * Outgoing pull pages are ordered by (dependency sort order, snapshot id), so a resume cursor has
+ * to carry both parts: snapshot ids ascend within a table, but a later table's ids can be lower
+ * than an earlier table's. Paging on the id alone silently skips records.
+ *
+ * The encoded form is opaque to clients — it is only ever produced and consumed here, so both the
+ * polling and streaming pull endpoints stay in step with what the query expects.
+ */
+export type SnapshotCursor = {
+  sortOrder: number;
+  id: string | number;
+};
+
+export const encodeSnapshotCursor = ({ sortOrder, id }: SnapshotCursor): string =>
+  btoa(JSON.stringify({ sortOrder, id }));
+
+export const decodeSnapshotCursor = (cursor?: string | null): Partial<SnapshotCursor> =>
+  cursor ? JSON.parse(atob(cursor)) : {};
--- a/packages/database/src/sync/snapshotCursor.ts
+++ b/packages/database/src/sync/snapshotCursor.ts
@@ -1,3 +1,5 @@
+import { InvalidParameterError } from '@tamanu/errors';
+
 /**
  * Outgoing pull pages are ordered by (dependency sort order, snapshot id), so a resume cursor has
  * to carry both parts: snapshot ids ascend within a table, but a later table's ids can be lower
@@ -14,5 +16,16 @@ export type SnapshotCursor = {
 export const encodeSnapshotCursor = ({ sortOrder, id }: SnapshotCursor): string =>
   btoa(JSON.stringify({ sortOrder, id }));
 
-export const decodeSnapshotCursor = (cursor?: string | null): Partial<SnapshotCursor> =>
-  cursor ? JSON.parse(atob(cursor)) : {};
+export const decodeSnapshotCursor = (cursor?: string | null): Partial<SnapshotCursor> => {
+  if (!cursor) return {};
+
+  // the cursor arrives as a query parameter, so a malformed one is a bad request rather than a
+  // server fault, and it says which parameter to look at
+  try {
+    return JSON.parse(atob(cursor));
+  } catch (error) {
+    throw new InvalidParameterError(
+      `fromId is not a valid pull cursor: ${(error as Error).message}`,
+    );
+  }
+};
--- a/packages/facility-server/app/sync/pullIncomingChanges.js
+++ b/packages/facility-server/app/sync/pullIncomingChanges.js
@@ -1,8 +1,9 @@
 import config from 'config';
-import { chunk } from 'es-toolkit/compat';
+import { chunk, omit } from 'es-toolkit/compat';
 import { log } from '@tamanu/shared/services/logging';
 import { SYNC_STREAM_MESSAGE_KIND } from '@tamanu/constants';
 import {
+  encodeSnapshotCursor,
   insertSnapshotRecords,
   SYNC_SESSION_DIRECTION,
   SYNC_TICK_FLAGS,
@@ -12,6 +13,15 @@ import { sleepAsync } from '@tamanu/utils/sleepAsync';
 import { calculatePageLimit } from './calculatePageLimit';
 
 const { persistedCacheBatchSize, pauseBetweenCacheBatchInMilliseconds } = config.sync;
+
+// sortOrder comes from the dependency ordering central pages by; it belongs to the cursor, not to
+// the snapshot table, which has no such column
+const toSnapshotRecord = record => ({
+  ...omit(record, ['sortOrder']),
+  // mark as never updated, so we don't push it back to the central server until the next local update
+  data: { ...record.data, updatedAtSyncTick: SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER },
+  direction: SYNC_SESSION_DIRECTION.INCOMING,
+});
 
 export const pullIncomingChanges = async (centralServer, sequelize, sessionId, since) => {
   const start = Date.now();
@@ -42,21 +52,13 @@ export const pullIncomingChanges = async (centralServer, sequelize, sessionId, s
       break;
     }
 
-    const { id, sortOrder } = records[records.length - 1];
-    fromId = btoa(JSON.stringify({ sortOrder, id }));
+    fromId = encodeSnapshotCursor(records[records.length - 1]);
     totalPulled += records.length;
     const pullTime = Date.now() - startTime;
 
     log.info('FacilitySyncManager.savingChangesToSnapshot', { count: records.length });
 
-    const recordsToSave = records.map(r => {
-      delete r.sortOrder;
-      return {
-      ...r,
-      data: { ...r.data, updatedAtSyncTick: SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER }, // mark as never updated, so we don't push it back to the central server until the next local update
-      direction: SYNC_SESSION_DIRECTION.INCOMING,
-    };
-  });
+    const recordsToSave = records.map(toSnapshotRecord);
 
     // This is an attempt to avoid storing all the pulled data
     // in the memory because we might run into memory issue when:
@@ -87,16 +89,7 @@ export const streamIncomingChanges = async (centralServer, sequelize, sessionId,
 
   const writeBatch = async records => {
     if (records.length === 0) return;
-    await insertSnapshotRecords(
-      sequelize,
-      sessionId,
-      records.map(r => ({
-        ...r,
-        // mark as never updated, so we don't push it back to the central server until the next local update
-        data: { ...r.data, updatedAtSyncTick: SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER },
-        direction: SYNC_SESSION_DIRECTION.INCOMING,
-      })),
-    );
+    await insertSnapshotRecords(sequelize, sessionId, records.map(toSnapshotRecord));
   };
 
   log.info('FacilitySyncManager.pulling', { since, totalToPull });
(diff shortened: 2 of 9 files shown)


packages/database/src/sync/snapshotCursor.ts
```diff
@@ -0,0 +1,18 @@
...
+export type SnapshotCursor = {
+  sortOrder: number;
+  id: string | number;
+};
+
+export const encodeSnapshotCursor = ({ sortOrder, id }: SnapshotCursor): string =>
+  btoa(JSON.stringify({ sortOrder, id }));
+
+export const decodeSnapshotCursor = (cursor?: string | null): Partial<SnapshotCursor> =>
+  cursor ? JSON.parse(atob(cursor)) : {};
```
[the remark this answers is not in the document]
passcod: 🤖 Fixed in bde464d — `decodeSnapshotCursor` now catches the decode failure and raises `InvalidParameterError`, naming `fromId` as the offending parameter, so a malformed cursor reads as a bad request rather than a server fault. Added unit coverage for a non-base64 cursor and for base64 that is not JSON.

packages/facility-server/app/sync/pullIncomingChanges.js
```diff
@@ -123,7 +116,7 @@ export const streamIncomingChanges = async (centralServer, sequelize, sessionId,
       case SYNC_STREAM_MESSAGE_KIND.PULL_CHANGE:
         records.push(message);
         totalPulled += 1;
-        fromId = message.id;
+        fromId = encodeSnapshotCursor(message);
```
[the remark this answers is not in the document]
passcod: 🤖 Fixed in bde464d — the stream loop now keeps the last record and `endpointFn` encodes a cursor only on the reconnect that actually needs one, so a first sync of hundreds of thousands of records does one encode instead of one per record.

packages/facility-server/app/sync/pullIncomingChanges.js
```diff
@@ -13,6 +14,15 @@ import { calculatePageLimit } from './calculatePageLimit';
 
 const { persistedCacheBatchSize, pauseBetweenCacheBatchInMilliseconds } = config.sync;
 
+// sortOrder comes from the dependency ordering central pages by; it belongs to the cursor, not to
+// the snapshot table, which has no such column
+const toSnapshotRecord = record => ({
+  ...omit(record, ['sortOrder']),
```
edmofro: @claude is this the same performance as our previous `delete` then spread?

edmofro (approved):
lgtm

packages/facility-server/app/sync/pullIncomingChanges.js
passcod: Close, but slightly worse — and both are much worse than the obvious third option.

Benchmarked on a realistic pulled record (11 snapshot columns + `sortOrder`, `data` a nested object), 50k records per round, 20 rounds, median, each variant in its own process so JIT ordering can't skew it:

| variant | ns/record | vs old |
| --- | --- | --- |
| old: `delete r.sortOrder` then spread | ~4100–4700 | 1.00x |
| this PR: `{ ...omit(record, ['sortOrder']) }` | ~5100–5600 | ~1.1–1.35x |
| `({ sortOrder: _, ...record }) => ({ ...record, … })` | ~570 | ~0.13x |
| spread with no key removal at all | ~275 | ~0.06x |

So the answer to the question as asked: same ballpark, ~10–35% slower.

The interesting part is that both are ~8x slower than they need to be, for the same reason. `delete` demotes the object out of its hidden class into dictionary mode, and the spread that follows then has to copy a dictionary-mode object. `omit` doesn't avoid that — `es-toolkit/compat`'s `omit` shallow-clones every key (`keysIn` + `getSymbolsIn`) and *then* `unset`s the key, which is a `delete` on the clone. So it pays the same dictionary-mode penalty on the outer spread, plus one extra full copy and a couple of array allocations. Removing the key without ever calling `delete` (rest destructuring, or a copy loop that skips it) keeps both objects in fast mode.

In absolute terms none of this is likely to matter: at 1M records it's ~5s of CPU vs ~4s vs ~0.5s, spread across a pull dominated by HTTP and the snapshot inserts. But the fast version is the same number of lines:

```js
const toSnapshotRecord = ({ sortOrder: _sortOrder, ...record }) => ({
  ...record,
  // mark as never updated, so we don't push it back to the central server until the next local update
  data: { ...record.data, updatedAtSyncTick: SYNC_TICK_FLAGS.INCOMING_FROM_CENTRAL_SERVER },
  direction: SYNC_SESSION_DIRECTION.INCOMING,
});
```

One thing worth stating explicitly either way: going back to `delete` is no longer just a style choice now that `streamIncomingChanges` shares this helper. There, `lastRecord` and the batched `records` hold the *same* message objects, and `lastRecord.sortOrder` is what `endpointFn` encodes into the resume cursor. A mutating `delete` inside `toSnapshotRecord` would strip `sortOrder` off the cursor record as a side effect of writing a batch, `encodeSnapshotCursor` would emit a cursor with no `sortOrder`, and the reconnect would silently restart from the beginning of the snapshot — precisely the bug this PR is fixing. Whatever this ends up as, it has to stay non-mutating; a comment saying so would be worth having.

---
_Generated by [Claude Code](https://claude.ai/code)_

edmofro: Ah right, that reuse elsewhere is a very good reason not to use `delete`! It is always a bit of a code smell mutating a shared object.

@passcod if this is 8x faster, let's do it 
```
const toSnapshotRecord = ({ sortOrder: _sortOrder, ...record }) => ({
  ...record,
```
I hear what it's saying about 5s not being much on a 1mil sync pull, but I'd like us to move towards "kill every delay" philosophy as we try to get more out of sync (and the rest of Tamanu)

passcod:
CI went red on my commit (`dc89bed`), but not because of it — flagging the cause so it doesn't read as a regression from this branch.

The only failure is `Test @tamanu/central-server 6/8`, on `__tests__/hl7fhir/materialised/Encounter.test.js › search › filters › by lastUpdated=gt` (expected total 1, got 0). My commit touches `packages/facility-server/app/sync/pullIncomingChanges.js` and its test only, so there's no path from it to FHIR Encounter search.

It's a race in the test itself — the update isn't awaited, so the search can run before it commits:

```js
newMat.update({ lastUpdated: addDays(new Date(), 5) });   // ← no await
const response = await app.get(`...Encounter?_lastUpdated=gt...`);
```

A one-line `await` fixes it. Out of scope for this branch, so I've left it alone rather than smuggling a central-server test fix into a sync PR — now its own fix PR: https://github.com/beyondessential/tamanu/pull/10657

I've re-run the failed job; will report back if it comes up red again, in which case the diagnosis above is wrong and it's mine to chase.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/facility-server/app/sync/pullIncomingChanges.js
passcod: Already done — that's `dc89bed` on this branch, which is why this thread now shows as outdated.

It's the rest-destructure version verbatim, plus the non-mutation requirement written down at both ends (on `toSnapshotRecord`, and on `lastRecord` where the shared objects and the cursor meet), and a test that fails if the helper ever mutates again. CI on that commit is green.

---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
Worth a look before re-queueing: the E2E failure reproduces, and it looks specific to this branch rather than flaky.

Both queue attempts failed on the **same** test, on all three of its attempts each time — `[AT-6829] flags an out-of-range numeric result in the lab request results table`, timing out after 80s on `waiting for getByTestId('container-pag3')`, i.e. `LabRequestDetailsPage` never renders:

| run | base | shard 3/4 result |
| --- | --- | --- |
| [30866236883](https://github.com/beyondessential/tamanu/actions/runs/30866236883) (queue attempt 1) | `fec948d` | **AT-6829 failed** 3/3, 55 passed |
| [30867273574](https://github.com/beyondessential/tamanu/actions/runs/30867273574) (queue attempt 2) | `53ed336` | **AT-6829 failed** 3/3, 54 passed (`note.spec` flaky, passed on retry) |
| [30866860526](https://github.com/beyondessential/tamanu/actions/runs/30866860526) — #10657 | `fec948d` | passed |
| [30864032082](https://github.com/beyondessential/tamanu/actions/runs/30864032082) — #10572, which added AT-6829 | `45d1f31` | passed |

So it isn't the new test being broken on main, and it isn't the runner: a different PR queued against the identical base passed, and the PR that introduced the test passed on its own.

I can't pin a cause from the diff — nothing here goes near lab requests or the web frontend. The only mechanism I can see is via sync: if the facility's pull drops or skips records, the reference data behind the request (ALT and its 5–40 range) wouldn't be there and the details page would never finish rendering. That's a guess, not a diagnosis — `facility-server.out` in the [server-logs-2 artifact](https://github.com/beyondessential/tamanu/actions/runs/30867273574/artifacts/8876906563) should settle it quickly, and the trace for the failing test is in the blob report.

For what it's worth, my commit (`dc89bed`) is not a plausible cause: it swapped `omit(record, ['sortOrder'])` for rest destructuring, which is behaviourally identical — same key removed, same non-mutation — so it can't change what reaches the snapshot table. Happy to dig further into the sync side if that would help.

---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
Ruling out my own guess: **it isn't sync.** I ran the sync suites against this branch on a local postgres, and they're green.

| suite | branch (`dc89bed`) |
| --- | --- |
| `facility-server __tests__/sync/` | 14 suites, 87 tests, all pass |
| `central-server __tests__/sync/` | 23 suites, 173 tests, all pass |

One caveat so the record is complete: on my first full-directory central run, `PatientMergeSync › does not pull child records (notes) … IF feature flag is false` failed (expected 0 notes, got 2). It does not reproduce — it passes in isolation on this branch, at `bde464d`, and on `main`, and the full directory re-ran clean on the branch. Running 23 suites concurrently against one database locally is the likely culprit; CI shards them, so it shouldn't bite there. Mentioning it only in case it shows up again.

There's also a second reason to doubt the sync theory that I should have weighed earlier: if the facility weren't getting its data, the other 54 tests in that shard — creating patients, encounters and notes — would have failed too, and they passed. A broad sync break doesn't fit a single failing spec.

So the AT-6829 failure still needs an explanation, and the reproducibility (3/3 retries on two queue attempts, against a base where other PRs pass) says it's real rather than flaky. The trace in the blob report and `facility-server.out` are still the fastest way in — the test never gets past `LabRequestDetailsPage` rendering at all, so whatever goes wrong happens before any of the out-of-range assertions. Happy to keep digging if it's useful, but I don't want to guess again in public.

---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
Correcting myself: I said the AT-6829 failure looked "specific to this branch rather than flaky". The first half was wrong, and your re-queue proved it.

[Run 30886…/`gh-readonly-queue/main/pr-10648-d29945d5e6efe3`](https://github.com/beyondessential/tamanu/actions/runs/30886556093) passed E2E with this branch unchanged at `dc89bed` — same code, newer base. So the variable was the base, not the diff:

| queue attempt | base | E2E |
| --- | --- | --- |
| 1 | `fec948d` | AT-6829 failed 3/3 |
| 2 | `53ed336` | AT-6829 failed 3/3 |
| 3 | `d29945d` | **passed** |

My inference was sound for the evidence I had — two failures on two bases, while other PRs passed on one of them — but I over-narrowed "not the runner, not the new test on main" into "therefore this branch". A third base was the case that separated them, and I didn't have it.

Between attempt 2 and 3 main picked up several web fixes, including `ac9d7ee` (TAM-6839) which touches `LabRequestView.jsx` — the very view whose container never rendered — and `88a4c8a`, which stops patient data being requested with a null patient id. I'm not claiming either is the fix; the point is only that the failure tracked the base and the base has since changed.

Nothing here needs anything from you — just didn't want a wrong call of mine sitting on the thread while the PR goes through.

---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
This latest E2E failure is not yours either — it's hitting other PRs in the same queue wave.

[Run 30967527311](https://github.com/beyondessential/tamanu/actions/runs/30967527311) (this PR, base `0d8270e`) failed on **shard 2/4**, `note.spec.ts › Edit Note Tests`, all three cases timing out on the same locator — `waiting for getByTestId('formsubmitcancelrow-confirmButton') to be detached`, i.e. the note edit modal never closes after submit. 1 hard failure, 2 flaky.

[Run 30967527813](https://github.com/beyondessential/tamanu/actions/runs/30967527813) — **#10684**, queued at the same minute, different base — failed on the same shard with the *same two note tests* on the same locator (flaky there), plus a hard failure in `labRequest.spec.ts › [AT-0069] Cancel lab request`. #10668 has also failed three times on base `0d8270e`.

So: different failing tests than before, appearing across at least two unrelated PRs simultaneously. AT-6829 did not recur — consistent with the run that passed at 01:37Z.

I'm not proposing a cause this time. Last time I inferred "specific to this branch" from two failures and it didn't survive a third data point, so I'll stick to what's directly comparable: these note failures are reproducing on someone else's PR too, so re-queueing this one is unlikely to be what fixes them.

---
_Generated by [Claude Code](https://claude.ai/code)_

passcod:
🤖 Correcting my earlier comment on this PR: **the AT-6829 E2E failure is not this branch**, and my reading of it was wrong.

I said it "reproduces, and it looks specific to this branch rather than flaky". The two controls I cited were runs that happened to pass — the successful merge-queue run at `00:50` sits between failures at `00:38` and `00:58`.

What it actually is: the test landed on `main` in `fec948d` at `2026-08-03 23:57`, the first merge-queue failure was 41 minutes later, and since then it has failed **8 of 16** merge-queue runs while PR-head runs pass. It took down a docs-only PR (#10683) with the identical failure. The cause is in `waitForTableToLoad`: the table renders a status row ("Loading…", or the no-data message) that is a `tr` with no click handler, so clicking the first row hits something inert and the details page never opens.

Fix is up as #10684, with a follow-up sweep in #10685. Not being worked on further from here — flagging it so nobody re-diagnoses it from my wrong comment.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #10649, opened 2026-07-31
feat(sync): no-issue: run a facility's first sync in phases

passcod:
### Changes

🤖 A facility server’s first sync pulls its whole share of the central database from the beginning of the sync timeline. For a facility with a long history of patients that is hours of work, and until it finishes the server cannot even be logged into: a user’s facilities are resolved against the local `facilities` table, which is still empty, so every login is refused as though the user had no access to any facility on the server.

The first sync now runs in three phases, each its own session pulling only its own tables, so the central snapshot for a phase completes in proportion to that phase’s data rather than the facility’s whole share:

- **boot** — facilities, users, roles and permissions, settings, departments, locations, translated strings, and reference data
- **catalogue** — programs and surveys, scheduled vaccines, templates, reference data relations, and patient records themselves plus which patients are marked for sync here
- **records** — everything recorded against a patient or an encounter

Login works once boot lands, patient search once catalogue does, and clinical history when records does. Reference data sits in boot because `facilities.catchment_id` points at it and login needs the facilities table.

Each model declares its phase next to its sync direction. A model that declares none arrives in the last phase, so a model added without thinking about phases holds up nothing that comes before it. A test enforces that no model sits in an earlier phase than something it holds a foreign key to — each phase is saved on its own, and would otherwise hit a reference to a record that has not been pulled yet. It walks the same associations the dependency-order save walks, so it cannot drift from what persist actually does.

Each phase pulls from the beginning of the timeline but is snapshotted at its own tick, so the earliest of the three becomes the pull cursor when the last phase lands — the only tick every model has been pulled up to. Changes to an early phase’s models made while the later phases ran fall after that tick and arrive with the next sync. A phase records its progress in the same transaction that saves its records, so a failed phase is retried on its own and the phases before it stand; today a first sync that dies at 90% restarts from nothing. Because the cursor stays unset until the last phase, the facility also keeps its place at the front of the sync queue for the whole initial sync.

Central needed one change: a snapshot pass whose model set is empty now contributes nothing instead of building a `record_type IN ()` clause. The boot phase reaches that because it carries no patient-linked tables — mobile could reach it today with the right `tablesToInclude`.

Also here: `specs/sync/phases.md` for the phased behaviour, and `specs/sync/overview.md` filled out with the session and pull-cursor behaviour it builds on.

Stacked on #10648.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


specs/sync/phases.md
```diff
@@ -0,0 +1,55 @@
+---
+id: PHASES
+---
+
+# Initial sync phases
+
+A facility server's first sync pulls the whole of its share of the central database, which for a facility with a long history of patients is hours of work.
+It runs in three phases, each a sync session of its own, so that each capability of the facility becomes available as soon as the data it needs has arrived.
+Every sync after the first is a single unphased session pulling the changes since the facility's pull cursor.
```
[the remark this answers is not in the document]
passcod: 🤖 They did disagree — thanks. Fixed in bfa7a56 by updating the spec, because boot is the phase that is forced: `Facility.belongsTo(models.ReferenceData, { foreignKey: 'catchmentId' })` (`packages/database/src/models/Facility.ts:79`), and login resolves a user's facilities against the local facilities table, so `facilities` has to be in boot and `reference_data` cannot land later than it. Moving `ReferenceData` to catalogue instead would fail the dependency-ordering test. `ReferenceDataRelation` stays in catalogue, and the spec now says why reference data is in boot.

packages/facility-server/app/routes/sync/sync.js
```diff
@@ -54,6 +56,10 @@ sync.get(
     const isSyncRunning = syncManager.isSyncRunning();
     const currentDuration = isSyncRunning ? new Date().getTime() - syncManager.currentStartTime : 0;
 
+    // null once the first sync has completed all its phases
+    const phase = await models.LocalSystemFact.get(FACT_INITIAL_SYNC_PHASE);
+    const initialSyncPhase = phase === null ? null : SYNC_PHASE_LABELS[parseInt(phase, 10)];
```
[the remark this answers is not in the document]
passcod: 🤖 Fixed in bfa7a56 — added `__tests__/sync/syncStatusRoute.test.js`, which builds the sync app and asserts the mapped label while a phase is set and `null` once it is cleared. Testing the sync app's route directly rather than the `apiv1` proxy, since the proxy mocks the connection and would not exercise this code.

edmofro:
@passcod two questions before I go into the review proper:
1. Why boot vs. catalogue? Seems it's only useful to be able to log in once catalogue has finished?
2. I would think later phases should include record types from earlier phases that have changed since the sync tick they were pulled up to, otherwise we could end up with FK errors (e.g. new ref data added that an encounter depends on). Unlikely but with sync, any hole will be fallen down at some point.

passcod:
agreed for 2, ~~I will check on that~~ addressed.

for 1, the benefit is a bit marginal, but the expectation is that boot will always be super fast (<1MB data) while catalogue could be a bunch more time (hundreds of MBs for large patients tables), so we get a "sync works, we can move on and check back later" signal earlier than having to wait for a larger sync. then we can hook that first phase into the wizard with a progress bar or something, so once that's cleared, someone can immediately log in, instead of having an undefined amount of time where the facility is set up but can't be checked on until the catalogue has landed.

edmofro:
I actually think it's more confusing for a user to be able to log in before the catalogue lands. Even if you keep it as three phases (and now it's done, I think that's kinda nice and we might use it for other reasons), please gate login on the catalogue completing sync.

passcod:
Landed on something in between: login works, but until initial sync finishes you get dumped into a splash screen that makes you wait until catalogue lands. Once catalogue lands you can proceed to the app with a click, and once the full initial sync is done, the splash screen disappears entirely as normal.

This also makes the wizard experience part built-in: once the wizard has set up the facility, you immediately get to the splash screen, so you know what's happening without Tamanu being available before it's ready.

packages/central-server/__tests__/sync/CentralSyncManager.setupSnapshotForPull.test.js
```diff
@@ -780,6 +806,52 @@ describe('CentralSyncManager.setupSnapshotForPull', () => {
         );
       });
 
+      it('does not sync all lab requests during an initial sync, even when enabled', async () => {
+        // The widened filter reaches every lab-bearing record in the deployment, so an initial sync
+        // would carry the whole lab history. A phase of an initial sync resumes from the tick the
+        // phase before it reached, so this cannot be inferred from `since` - the client says so.
```
edmofro, line 812: The can't be inferred from since thing is one of those comments that is about what it used to do, whereas there's nothing unexpected about the isInitialSync mechanism if you don't know what it used to do.
```suggestion
        // The entire lab history is not part of initial sync, because a.) it's a lot of data, and b.) that data is for old requests that probably don't need to be processed. 
```

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -565,10 +575,19 @@ export class CentralSyncManager {
           where: { facilityId: facilityIds },
         });
 
-        // regular changes
+        // regular changes, leaving out the tables about to be snapshotted in full from the start of
+        // the timeline: a record in one of those changed since `since` would otherwise be snapshotted
+        // by both passes, and the persist creates a row per snapshot record, so the second copy is a
+        // duplicate primary key
```
edmofro, line 581: Turns out it's really confusing having `isInitialSync` and `since` diverge. We've gotta always keep sync as conceptually simple as we can (hard!) so any time a comment is getting unwieldy like this it's a smell to put in more thought about how to untangle things. 

I have considered two options: 
1. Set an `until` cursor that carries through all phases of initial sync, meaning the later snapshot tables are not going to get ahead of earlier tables and cause inconsistencies. Unfortunately this is broken: a record might have been updated in the meantime and would be excluded from the sync by the cap, even if it might have other records depending on it.
2. Resync all earlier phases in full during later phases. Really simple! Quite silly in terms of duplicate work. But maybe resyncing the catalogue isn't the worst thing in the world for the sake of "this is stupid but I can understand it immediately". In a country with a million patients, or ten million patients, yeah, it is probably too much double work.
3. Do a single snapshot at the start, and have the phases pull subsets from that snapshot. The snapshot phase is likely the longest phase of most initial syncs though, so it kills the concept a bit.
4. Hold open the repeatable read transaction across the snapshot part of all sync phases. Kick off all three snapshots at the same time, within that transaction, but let the client know the earlier phases are ready as soon as their snapshot is up. This one is interesting to me...it might involve just as much complexity, but worth playing with.
5. Restart earlier phases if we find that an outgoing record in a phase relies on a foreign record that has been updated since that phase synced down. This is complicated, slow, and not all our FKs are set up as proper FKs (i.e. polymorphic relationships aren't). Disregard.
6. Just drop this comment lol...it's actually not that hard to understand if you just read the code, and there's nothing unexpected about "regular changes filters out tables for full resync". Maybe just do that 😂

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -578,8 +597,9 @@ export class CentralSyncManager {
           sessionConfig,
         );
 
-        // any tables for full resync from (used when mobile needs to wipe and resync tables as
-        // part of the upgrade process)
+        // any tables to pull from the beginning of the sync timeline rather than from `since`: mobile
+        // wiping and resyncing tables as part of an upgrade, and a phase of a facility's first sync
+        // pulling its own tables for the first time while catching the earlier phases' up from `since`
```
edmofro, line 602: This can be simplified too
```suggestion
        // any tables to pull from the beginning of the sync timeline rather than from `since`, i.e. 
        // - any phase of initial sync from a facility server (boot tables, then catalogue tables, then the rest)
        // - mobile wiping and resyncing tables as part of an upgrade
```

packages/database/src/models/Model.ts
```diff
@@ -85,6 +92,9 @@ export class Model<
       );
     }
     this.syncDirection = syncDirection;
+    // a model that doesn't nominate a phase arrives in the last one, so a model added without
```
edmofro, line 95:
```suggestion
    // a model that doesn't nominate a phase arrives in the final "records" phase, so a model added without
```

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -528,9 +533,14 @@ export class CentralSyncManager {
...
-        // no need for historical ones on initial sync, and no need on mobile
-        syncAllLabRequests: syncAllLabRequests && !session.parameters.isMobile && since > -1,
+        // For facilities with a lab, need ongoing lab requests. This widens the snapshot past both
+        // the marked-for-sync patients and the facility itself, to every lab-bearing record in the
+        // deployment, so it stays off for the whole of an initial sync - however many sessions that
+        // takes - and off on mobile. The client says whether it is doing an initial sync: it can't be
+        // inferred from `since`, because a phased initial sync resumes each phase from the tick the
+        // one before it reached.
+        syncAllLabRequests:
+          syncAllLabRequests && !session.parameters.isMobile && !isInitialSyncSession,
```
edmofro, line 543: Again I think this comment overexplains in a way that becomes more confusing. If it wasn't Claude, I apologise for the offence.

I also think we can just use the presence of a non-empty `tablesForFullSync`, because currently the only places that is used is:
- Initial sync 
- Mobile 
The same two places we never synced all labs. If it's used in more places in future...well I think that use case will probably also mean it's fine not to have historic labs synced, and if it's not, we can find out then! Doing this retires a whole boolean flag, and an early mentor was very strong on avoiding the build up of boolean flags to solve problems. In this case it's a boolean flag I found a bit unexpected too, because on initial read I thought it was going to hold more weight than it does - it's only used for this labs thing hey? - but sounds like it's much more load bearing.

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -441,13 +441,18 @@ export class CentralSyncManager {
 
   async setupSnapshotForPull(
     sessionId,
-    { since, facilityIds, tablesToInclude, tablesForFullResync, deviceId },
+    { since, facilityIds, tablesToInclude, tablesForFullResync, isInitialSync, deviceId },
```
edmofro, line 444:
```suggestion
    { since, facilityIds, tablesForNormalSync, tablesForFullSync, isInitialSync, deviceId },
```
I don't know about what the best names are (`tablesForNormalSync` I've been through about 5 options before posting this), but I do think the two should align, and it shouldn't be `resync` now it's used for initial sync too.

edmofro (changes requested):
Have reviewed central side, will review the larger facility side tomorrow

Codeowner got opinions...

All of those opinions are: it's so easy for sync to get conceptually too complicated for us to maintain it, that's what happened to sync v1, so let's do everything we can to keep it simple.

packages/central-server/__tests__/sync/CentralSyncManager.setupSnapshotForPull.test.js
```diff
@@ -604,7 +602,7 @@ describe('CentralSyncManager.setupSnapshotForPull', () => {
         {
           since: 1,
           facilityIds: [facility.id],
-          tablesToInclude: ['programs'],
+          tablesToPull: { incremental: ['programs'] },
```
edmofro, line 605: Ohhh yeah this is the right call re. naming

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -528,9 +536,14 @@ export class CentralSyncManager {
...
       const sessionConfig = {
-        // for facilities with a lab, need ongoing lab requests
-        // no need for historical ones on initial sync, and no need on mobile
-        syncAllLabRequests: syncAllLabRequests && !session.parameters.isMobile && since > -1,
+        // For facilities with a lab, need ongoing lab requests. This widens the snapshot past both
+        // the marked-for-sync patients and the facility itself, to every lab-bearing record in the
+        // deployment, so it stays off for the whole of an initial sync - however many sessions that
+        // takes - and off on mobile. The client says whether it is doing an initial sync: it can't be
+        // inferred from `since`, because a phased initial sync resumes each phase from the tick the
+        // one before it reached.
```
edmofro, line 544: This overcomment remains

packages/central-server/app/sync/CentralSyncManager.js
```diff
@@ -441,13 +441,18 @@ export class CentralSyncManager {
...
     sessionId,
-    { since, facilityIds, tablesToInclude, tablesForFullResync, deviceId },
+    { since, facilityIds, tablesToPull, isInitialSync, deviceId },
     unmarkSessionAsProcessing,
   ) {
     let transactionTimeout;
     try {
       const { models, sequelize } = this.store;
 
+      // A client that doesn't say tells us by pulling from the start of the timeline, which is what an
```
edmofro, line 451: If we change mobile to pass `isInitialSync`, we can remove this conditional assignment and confusing comment

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -257,11 +285,44 @@ export class FacilitySyncManager {
     log.debug('FacilitySyncManager.updatedLastSuccessfulPush', { currentSyncClockTime });
   }
 
-  async pullChanges(sessionId) {
-    // syncing incoming changes happens in two phases: pulling all the records from the server,
+  // If it fails to start, the next SyncTask run picks the same phase up.
+  startNextPhaseInBackground(phase) {
+    log.info('FacilitySyncManager.startingNextPhase', { phase: SYNC_PHASE_LABELS[phase] });
+    // not urgent: a facility mid-initial-sync already keeps its place at the front of the queue by
+    // reporting itself as never-synced, and urgent is for syncs a person is waiting on
```
edmofro:
```suggestion
    // reporting itself as never-synced, whereas urgent is for syncs a person is waiting on
```

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -139,7 +157,15 @@ export class FacilitySyncManager {
       startTime,
     });
 
-    const pullSince = (await this.models.LocalSystemFact.get(FACT_LAST_SUCCESSFUL_SYNC_PULL)) || -1;
+    // an unfinished first sync leaves this unset, so the facility reports as never-synced and keeps
+    // its place at the front of the sync queue
+    const pullSince = (await this.models.LocalSystemFact.get(FACT_LAST_SUCCESSFUL_SYNC_PULL)) ?? -1;
+
+    // a facility that hasn't finished its first sync performs one phase of it per run
+    const phase = await getInitialSyncPhase(this.models);
```
edmofro, line 165:
```suggestion
    const isInitialSync = pullSince === -1;
    const initialSyncPhase = isInitialSync ? await getInitialSyncPhase(this.models) : null;
```
These don't add anything logically, but I think will make it cleaner to read everywhere below in this class. We can use `isInitialSync` everywhere we're testing presence of `phase` as the condition, so if you come across that line in isolation it's immediately clear what it means.

packages/facility-server/app/sync/FacilitySyncManager.js
```diff
@@ -257,11 +285,44 @@ export class FacilitySyncManager {
...
+    const pullParams = phase
+      ? {
+          since: await getPhaseCatchUpSince(this.models),
+          tablesToPull: {
+            incremental: tableNamesOf(modelsForIncrementalPull),
+            full: tableNamesOf(modelsForFullPull),
+          },
+          isInitialSync: true,
+        }
+      : { since: pullSince };
```
edmofro, line 325: Would extract this into two functions `getInitialSyncPullParams` and `getPullParams`. Less DRY but each one will be nice and clear, and read less dense than this series of ternaries.

packages/web/app/App.jsx
```diff
@@ -71,6 +79,24 @@ export function App({ sidebar, children }) {
   if (!isServerAlive) return <UnavailableStatusPage />;
   if (serverStatus?.setupRequired) return <SetupWizardView />;
   if (!isUserLoggedIn) return <LoginView />;
+
+  // A facility part-way through its first sync can be logged into but not worked in. Sits ahead of
+  // facility selection: there may be no facility to select yet, and selecting one wouldn't help.
+  const initialSyncPhase = serverStatus?.initialSyncPhase;
+  const canEnterDuringInitialSync = initialSyncPhase === 'records';
+  if (initialSyncPhase && !(canEnterDuringInitialSync && hasEnteredDuringInitialSync)) {
```
edmofro, line 87:
```suggestion
  if (initialSyncPhase && !hasEnteredDuringInitialSync) {
```
Feels unnecessary to have that gate - if they have entered, they can. This is simpler to read.

specs/sync/overview.md
```diff
@@ -5,3 +5,24 @@ id: SYNC
...
+
+Facility servers and mobile devices each hold a subset of the central database and sync with central on a schedule.
+A sync is a session: the client pushes the changes it has made, then pulls the changes central holds for it.
+This spec covers the parts that apply to every sync; `lookup-table.md` covers how central assembles what a client can pull, and `phases.md` covers the phases a facility's first sync runs in.
+
+## Sessions
+
+- [ ] A client syncs within a session, and runs one session at a time.
+- [ ] Central runs a limited number of concurrent sessions, and a client that requests one while central is at capacity is queued rather than refused.
+- [ ] The sync queue is ordered by how far behind each client is, so the client that has pulled up to the earliest tick is admitted first.
```
edmofro, line 17: Ah wow this is nice to have highlighted. It probably acts against us in situations like Samoa, where a major facility (like TTM) has a healthier internet connection, so is more likely to be up to date, so is therefore leapfrogged by poorly connected facilities. This explains why we see Samoa get stuck behind non-syncing facilities, which was a mystery to me!

@rohan-bes we should change this...

specs/sync/phases.md
```diff
@@ -0,0 +1,74 @@
...
+- [ ] The boot phase carries the data a facility server needs to authenticate a user and start serving requests: facilities, users, roles and their permissions, settings, departments, locations and location groups, and translated strings.
+- [ ] The boot phase also carries reference data, because a facility record names its catchment as a reference data record and so cannot arrive before it.
+- [ ] The catalogue phase carries the rest of the data that is not scoped to a patient: reference data relations, programs and surveys, scheduled vaccines, and templates.
+- [ ] The catalogue phase also carries patient records themselves, without the data recorded against them, together with the records marking which patients are synced to this facility.
+- [ ] The records phase carries the data scoped to a patient or an encounter: encounters, notes, vitals, lab requests and tests, imaging, invoices, documents, appointments, and program responses.
+- [ ] A user can log in once the boot phase has completed, search for and open patients once the catalogue phase has completed, and see a patient's clinical history once the records phase has completed.
```
edmofro:
```suggestion
- [ ] A user can log in once the boot phase has completed (but sees only a progress bar), search for and open patients once the catalogue phase has completed, and see a patient's clinical history once the records phase has completed.
```

edmofro (changes requested):
As per earlier reviews, always working hard to make sync code as easy to parse as possible.

The UI aspects will need review by Flic, can you put up a card for this and attach screenshots for her? 

Looking good though!

PR #10656, opened 2026-08-03
perf(facility): TAM-7029: use faith for the central server connection

dannash100:
### Changes

The facility server's connection to central now goes through [faith](https://github.com/passcod/faith), a rust fetch with a DNS cache and HTTP/2. Sync is thousands of small requests over links that are often slow, and node's built-in fetch gives us neither. `TamanuApi.fetchImplementation` was already the swap point (tests use it), so the wiring is one line in `CentralServerConnection`.

Two things that might read as mistakes:

- faith reports a network failure as the entire rust error chain, a couple of thousand characters of debug output. Left alone, every one of the 15 retry attempts writes that to the log and the sync error record, so the wrapper keeps the first segment capped at 200 characters, which is where the diagnosis is ("dns error", "tcp connect error", certificate failures). Error classification is untouched: still `RemoteUnreachableError`, still recoverable, same backoff.
- `TAMANU_DISABLE_FAITH_FETCH` is an env var rather than a setting because facility settings only arrive by sync. A setting can't be flipped for the one facility whose sync this breaks.

Checked by driving the real `TamanuApi` against a local fake central: the framed sync-pull stream decodes identically to node fetch, including a frame split across two writes, the error-response path still yields a `Problem`, and an unreachable host yields the trimmed `RemoteUnreachableError`. HTTP/2 negotiates against public hosts, a `Headers` instance and our extra config keys are accepted. Attachments upload as base64 JSON, so faith having no `FormData` doesn't reach us. Prebuilds cover win32 x64/arm64 and linux gnu/musl, and `timesimp` in this package already ships the same napi layout.

Not measured against a real sync yet, and not tried against a facility with a private CA. faith's docs say it trusts the platform store; worth confirming before this goes anywhere near a deployment that has one.

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [x] Build all images (amd64 + Windows; default is arm64 only) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/facility-server/app/sync/faithFetch.js
```diff
@@ -0,0 +1,20 @@
+import { fetch } from '@passcod/faith';
+
+const MAX_ERROR_LENGTH = 200;
+
+// faith reports a failure as the whole Rust error chain, a few thousand characters of
+// debug output that would land in the log line and the sync error record. The head of
+// it holds the diagnosis: "dns error", "tcp connect error", a certificate failure.
+export const summariseFaithError = message => {
```
passcod: I… guess but note the last section of faith's readme, which has an error classification already and exposes it on one property with a stable ID

packages/facility-server/app/sync/faithFetch.js
```diff
@@ -0,0 +1,20 @@
...
+  const [head] = message.split(' -> ');
+  return head.length > MAX_ERROR_LENGTH ? `${head.slice(0, MAX_ERROR_LENGTH)}…` : head;
+};
+
+export const faithFetch = async (url, options) => {
+  try {
+    return await fetch(url, options);
+  } catch (error) {
+    if (!error.code) throw error;
```
passcod: Yeah, see, here you detect error.code and then… do nothing with it. What?

packages/facility-server/app/sync/faithFetch.js
dannash100: Oh true!

packages/facility-server/app/sync/faithFetch.js
dannash100: lol random

passcod (approved):
marking passed but ofc the item is mostly testing :)

PR #10659, opened 2026-08-04
tweak(vitals): TAM-7036: serve vitals edit history from the changelog

dannash100:
### Changes

`vital_logs` duplicated what the changelog records: every vitals edit was stored twice, and the table is ~330k changelog rows a year of pure duplication on a mid-size deployment. This serves vitals edit history from `logs.changes`, per `specs/vitals/overview.md` (added here). Dropping the table is deferred to a follow-up.

Found on the way and fixed first: `updateWithReasonForChange` cleared the audit reason before commit, but changelog triggers are deferred to commit, so **chart edit reasons have been silently landing as NULL**. Proven by a failing test before the fix; the fix is deleting the premature clear (the var is transaction-scoped).

The pieces:

- Vitals PUT joins charts on `updateWithReasonForChange`; POST and calculated answers get `createWithReasonForChange` so insert entries carry the reason too. The answer's insert entry replaces the "initial vital log" (web already treats `logs.length > 1` as edited, so shapes line up).
- `getAnswersWithHistory` uses the changelog select for vitals and charts alike, now with `migration_context IS NULL` and deterministic ordering, which the chart path previously lacked.
- DML migration converts `vital_logs` rows to changelog entries, reusing the vital log's id so the copies central and facility each author dedupe on sync; original user/time/reason preserved, `migration_context` null so history views show them. Where the changelog trigger was already live the entry exists but lacks the reason (the cleared-var bug), so the migration restores those reasons instead of duplicating: matched on record + body + nearest preceding entry within 30 minutes, since the trigger stamps transaction-start time while Sequelize stamped vital logs seconds later (verified against live data). Entries are synthesised only for history predating the answer's first real entry.
- Writers are removed (mobile only ever wrote initial logs at submission), but the table stays for now: the DDL drop, the sync-lookup removal and the mobile TypeORM migration are deferred, so the server and mobile models and the dbt models are still here.
- 🤖 Mobile answers never got an initial vital log (its writer is a submission hook, which sync doesn't run), so their provenance relied on running `generateInitialVitalLogs` by hand: ~126k such rows on one production deployment, last batch 2025-04-30. Their changelog insert entry can't stand in, because central authors it at sync time with no audit user, so history showed the system user as the recording clinician. The original recording now reads from the response, which fixes that and removes the manual dependency.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/database/src/migrations/1785811717000-migrateVitalLogsToChangelog.ts
```diff
@@ -0,0 +1,116 @@
...
+// synthesised only for vital logs older than their answer's first trigger-written entry.
+//
+// Synthesised entries reuse the vital log's id, so the copy a facility authors and the
+// copy central authors from the same synced row dedupe on sync, and carry a sentinel
+// device id (the authoring device was never recorded), which also makes the down
+// self-contained. migration_context stays null so history views read them as
+// operational. Their record_data is the answer's current row with body and updated_at
+// overlaid, not a point-in-time snapshot: only body is reconstructable from a vital log.
+//
+// Only answers that were actually edited get entries synthesised: a lone initial
```
julianam-w, line 20: ~~This is different to how logs.changes is expected to work. I'm going to push back on this.~~

Since vital logs are captured in survey_response_answers, prior backfill would already have captured final state in logs.changes.

So only including vitals with edit is right but to include logs for change that is not the final state.

Do we need to update surveys, yes(?) since we have info on who made the change.

dannash100: Oh true that! No im glad for push back this one is a complicated situ

dannash100:
Postponing

PR #10686, opened 2026-08-04
tweak(labs): TAM-7044: serve lab request status history from the changelog

dannash100:
### Changes

`lab_request_logs` duplicates history the changelog already holds: every writer changes `lab_requests.status` in the same transaction, and the log table is itself audited, so one status change was stored three times. Follows TAM-7036 (vital_logs). Spec: `specs/labs/status-history.md`.

A status transition is now an entry whose status differs from the one before it, which is a `LAG` over the request's changelog entries. Both endpoints keep their response shapes, so the status log modal and the printout's `publishedBy` are untouched.

**`encounter_history` is deliberately not here.** It looked like the same job, but its `date` is a *clinical* timestamp (the initial snapshot backdates to `encounter.startDate`, moves pass a user-supplied `submittedTime`) and the Medici FHIR report exposes it as `assignedTime`. A changelog entry has no field for that, so migrating would silently turn "moved ward at 3pm yesterday" into "recorded this morning". Split to TAM-7045 for a product decision.

Two things worth a reviewer's attention:

- **Attribution needed a new tool.** The FHIR results ingest credited its log to the FHIR requester and the auto-publisher to the request's requester, neither being the session user. Since the changelog trigger is deferred and reads the audit user at commit, one transaction can only have one audit user, so `runWithAuditUser` wraps those two whole transactions rather than individual writes. Without it the status log would start naming the integration or nobody.
- **The migration only synthesises pre-changelog history.** Audited-era log rows already have a matching entry, so inserting for them would double every transition. Entries reuse the log's id (so a facility's copy and central's dedupe on sync) and carry a sentinel device id, which also makes the down self-contained.

Historical attribution for FHIR/auto-published transitions from the audited era will now read as whoever the session was, not the requester; existing entries aren't rewritten. Draft until the data team items below are confirmed and the pre-audit row count is sized on a real deployment.

### Data team requirements

This drops a table their models read, so their changes need to land with (or before) the release that carries this:

- **`tamanu-source-dbt`: `models/bases/lab_request_logs.*`** is a curated base model, so downstream models break on drop. Rewrite against `logs.changes` (`record_type = 'lab_requests'`, status at `record_data->>'status'`, migrated rows carry `device_id = 'lab-request-log-migration'`) or retire it.
- **Palau lab-tally report** queries `lab_request_logs` directly; same rewrite or retire call.
- **Per-deployment custom reports** can also reference the table. Before a deployment upgrades, run: `SELECT id, version_number FROM report_definition_versions WHERE query ILIKE '%lab_request_logs%'` and rewrite any hits.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


dannash100:
Might come back to this

PR #10694, opened 2026-08-05
feat(blobTransfer): F2: add fetch-by-hash blob transfer subprotocol

dannash100:
Implements the hash-keyed blob transfer channel between facility and central servers, building on the blob store primitive.

- Adds a `BlobTransferChannel` on the facility server (probe/fetch/offer-push, streaming, resume-from-delivered-bytes) and corresponding transfer routes plus error serialisation on the central server, wired into both `ApplicationContext`s
- Introduces availability-state constants distinguishing upload-pending from fetch-pending content, and a `BlobHashMismatchError` for content verification on receipt
- Generalises the previous attachment-GET flow into a reusable `faithFetch` helper on the sync connection
- Adds the `blobs` table/model and migration (central and mobile), blob store integration, related config defaults, and utility helpers for hash handling
- Adds specs under `specs/blob-storage/` (including `transfer.md`) capturing the workshopped design, plus test coverage for `BlobStore`, `BlobTransferChannel`, and the central transfer routes
- Does not yet touch any consumer table — records still sync carrying only the hash

### Review Hero

- [ ] **Run Review Hero**


packages/central-server/config/default.json5
```diff
@@ -459,5 +459,11 @@
     // This committed key is for local dev/test ONLY; deploys override it via
     // the CRYPTO_KEY_FILE env var pointing at a mounted secret.
     "keyFile": "config/dev-secret.key"
+  },
+  "blobStorage": {
+    // Root directory of the content-addressed blob store, resolved relative to
+    // the server's working directory when not absolute. May be pointed at a
+    // separate volume to remove disk contention with the database.
+    "root": "data/blobs"
```
passcod: hmm.... settings, env?

packages/facility-server/app/sync/CentralServerConnection.js
```diff
@@ -8,6 +8,7 @@ import { log } from '@tamanu/shared/services/logging';
 
 import { version } from '../serverInfo';
 import { getSyncConfig, getServerFacilityIds } from '../serverConfig';
+import { faithFetch } from './faithFetch';
```
passcod: hmm this looks like the fáith branch is merged into this one?

packages/database/src/blobStore/BlobStore.ts
passcod, line 1: We may want to create a new sans-io style package so we can reuse the blob shared code (store management, transfer, etc) between servers and mobile?

passcod (changes requested):

packages/facility-server/app/sync/CentralServerConnection.js
dannash100: Yeah because spec mentioned it, so I told it to merge it in 🤔

packages/central-server/config/default.json5
dannash100: Yeah it looks like the first pr thats merged into epic also added a config so addressing now. Need to work out why that isn't being enforced properly, I added to rules im sure

passcod (approved):

PR #10697, opened 2026-08-05
feat(blobs): G2: Facility blob outbox and LRU cache

dannash100:
### Review Hero

- [ ] **Run Review Hero**


database/model/public/blobs.md
```diff
@@ -23,3 +23,23 @@ The blob's integrity state: `verified` when its content matched its hash at the
...
+When the blob's content was last read (or admitted, whichever is later), used
+for least-recently-used eviction ordering on facility and mobile servers.
+Recency updates may be coalesced, so this is a lower bound on the true last
+access.
+{% enddocs %}
+
+{% docs blobs__sync_cycles_unpushed %}
+How many successful sync cycles this blob has survived in the outbox while
+eligible for push and not actively transferring. The outbox dysfunction
+measure: a working connection with an undelivered blob. Zero for cache blobs.
```
passcod: I'm not sure it makes sense to have this as a counter? The more robust thing would be to record the created at or the sync tick or something like that (and created at would be independently useful too), and then compare that against sync state?

passcod (changes requested):

database/model/public/blobs.md
dannash100: Oh yes ok good idea

passcod (approved):

PR #10709, opened 2026-08-06
chore(blobs): L2: specify mobile blob storage and lazy fetch

dannash100:
### Changes

Specs for aligning the mobile app with the blob transfer subprotocol. Spec and
planning only, no code.

Mobile already keeps attachment bytes on the filesystem, but reads them back into a
Buffer on load purely so they ride inside the sync record. This card replaces that
with hash-carrying records plus lazy fetch, and puts the device store under a bounded
cache.

The decision worth a reviewer's attention is how mobile learns a hash. Attachment
records sync down and mobile fetches by hash, rather than keeping the id-keyed path
and having central resolve id to hash. The id-keyed option looks like the safer one
because attachments today carry no patient association and so cannot be sync-scoped,
but J2 gives them their owning record's linkage, which removes that constraint. It
costs a round trip before every cold read and keeps a mobile-only path alongside the
general channel, so it loses once J2 lands.

Two consequences: mobile's attachment sync becomes bidirectional, and records are
retained rather than deleted after push, since the record is what makes content
refetchable.

Capture also becomes offline-capable. The `health/canUploadAttachment` pre-check
against central moves to push time, which falls out of the outbox model. Worth
flagging to testers as a real behaviour change.

New spec `blob-storage/mobile.md`, plus folds into the cache, disk usage, and
integrity specs for the device budget, the free-disk floor, and mobile running no
scheduled scrub. Plan and 38 test cases included.

Held pending J2: the spec cross-references J2's attachments spec, and whether mobile
behaviour folds into that file or stays its own is a merge-time call.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


specs/blob-storage/mobile.md
```diff
@@ -0,0 +1,57 @@
...
+  server; a record is not removed to reclaim space, since the record is what makes
+  its content refetchable.
+- [ ] A device holds no binary column for attachment content. The bytes live in the
+  device's blob store and are reached through the record's hash.
+- [ ] Records synchronise at their own pace whether or not the bytes have moved, so a
+  large photo never holds up the survey response that references it.
+
+## Capturing content
+
+- [ ] Content captured on the device, a survey photo among it, is admitted to the
```
passcod:
```suggestion
- [ ] Content captured on the device (like photos, uploads, etc) is admitted to the
```

passcod (approved):
one reword of an awkward phrasing

PR #10713, opened 2026-08-07
feat(blobs): P2: blob integrity scrub and self-heal

dannash100:
### Changes

Implements the blob integrity spec (SCRUB) over the store, transfer and cache tiers already on `workhorse/b2`.

Bit rot on bare metal and NTFS is silent: nothing under the store notices a flipped bit, so a corrupt attachment is found the day a clinician opens it. This adds detection on every path and repair where a good copy exists.

**Detection.** A `BlobScrubber` runs three passes: verification (least-recently-scrubbed first, bounded per pass by blob count and bytes), reconciliation (walks the fan-out for bytes no registry row names), and referential integrity. `BlobStore.get()` now hashes while streaming and fails a whole-blob read that does not match.

**Repair.** The scrubber only detects; each server supplies a healer, because identical bytes mean different things. A facility grades on the existing tier: cache is dropped and refetched, outbox is quarantined and escalated since it may be the only copy. Central treats every copy as authoritative.

Two decisions worth a look:

- **Ranged reads are not verified.** Part of a blob cannot be checked against a hash of the whole, so they rely on receipt verification and the scrub. That is what keeps Bao deferred.
- **Central's peer healing is the offer route answering `wanted` for a quarantined hash**, not central fetching. It cannot reach a facility on demand and the spec rules out indexing what facilities hold, so repair rides a connection the facility makes anyway.

The spec left "delivered" undefined for central's referential check; this defines it as a reference standing longer than a delivery grace, since push is sync-first. It finds nothing until J2 or K2 register a reference source.

Migration is numbered clear of J2 (`1785830000000`) and K2 (`1785860000000`). No mobile migration: a device runs no scrub.

### Auto-Deploy

- [ ] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/.agents/skills/draft-card-breakdown/SKILL.md
+++ b/.agents/skills/draft-card-breakdown/SKILL.md
@@ -1,0 +1,45 @@
+---
+name: draft-card-breakdown
+description: "Workshop how to break this card into smaller spawned cards and capture entries in the card breakdown"
+label: "Draft card breakdown"
+pill-order:
+  not-started: 8
+  specifying: 13
+  implementing: 7
+  reviewing: 7
+surface: both
+jockey-hint: "Always available but low-traffic — surface as an available pill, not a top suggestion. Most cards are implemented as one PR; only suggest prominently when the conversation has already revealed the card is too large to ship in one piece."
+workhorse-version: 0.3.0
+---
+
+## Your task: Draft card breakdown
+
+If you don't already have this card's context (title, identifier, description) — for instance when running outside Workhorse — establish it first by following `.agents/docs/card-context.md`.
+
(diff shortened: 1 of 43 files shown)


dannash100:
bugbot run

PR #10753, opened 2026-08-11
feat(blobs): TAM-7037: collect orphaned blobs on central

dannash100:
### Changes

`specs/blob-storage/reclamation.md` shipped with the epic and nothing implements its central half. The facility half (evict cache under an LRU budget, never evict the outbox) is built and covered; central orphan collection is not. No card in the plan claimed it, and there is not one `spec: RECL` tag anywhere in the codebase, which is how the gap survived: every other spec area tags its implementation.

`findOrphanedBlobHashes` anti-joins the registry against every registered reference source, the changelog, and the quarantine table. A bounded pass deletes what it finds through `BlobStore.delete`, so the parity sidecar and the registry row are handled as they already are. Central only: there is no facility setting for it, so it cannot be scheduled there, and a test asserts that.

Reference rows are matched without a `deleted_at` filter, so a soft-deleted record still holds its content, which is the whole point of the spec's "live or soft-deleted". Only `verified` blobs are eligible: corrupt content is retained for investigation and `absent` is the record of a fault, not spare disk.

Four decisions a reviewer might otherwise read as mistakes:

- **The safety window is 7 days**, on `blobs.created_at`. What it has to clear is the gap between admitting content and writing the reference naming it, which on central is a single request (push is sync-first, so a synced reference is always ahead of its bytes). A week is orders of magnitude past that, and long enough that a misbehaving pass is noticed before it takes anything recent.
- **`maxBlobsPerPass` bounds what a pass takes, not what it scans.** A scan window over the registry would need a persisted cursor; without one, an oldest-first window sticks behind old-but-referenced blobs and never reaches a younger orphan. Full coverage every pass, bounded destruction per pass.
- **The changelog index is a partial expression index**, not GIN. GIN over `record_data` cannot serve `->>`, which is why `dropUnusedChangelogIndexes` removed the old one. The predicate is on the extraction rather than on today's table names, so registering another source later does not silently cost the index.
- **One race remains and is inherent to an age-based rule.** If long-orphaned content is re-uploaded, `put` dedupes onto the existing row without touching `created_at`, so it stays eligible for the moment between `put` returning and the referencing row being inserted. Closing it needs something other than age.

Ten cases in `blobReclamation.test.js`, including the superseded-asset one that proves the changelog keeps replaced content alive. 65 tests green across reclamation, referential integrity and transfer.

### Auto-Deploy

- [ ] **Deploy** 

### Review Hero

- [ ] **Run Review Hero**


dannash100:
🤖 Withdrawing this rather than hardening it before QA. The convergent finding across this round is right: the safety window is keyed on `blobs.created_at`, and `put` on content already held leaves that alone, so a blob first seen a month ago can be deduplicated onto a moment before its new reference commits. A pass in that gap hard-deletes the row and the file, and the reference lands pointing at content that is gone from the authoritative copy.

That is fixable (re-check liveness under `SELECT … FOR UPDATE` in the delete transaction, refresh recency on a dedupe hit, plus the missing `blobs` scan index, schema-qualified identifiers, and a registration guard that fails loudly). But the asymmetry says don't rush it: not collecting rare orphans costs disk, collecting the wrong one costs a clinical attachment no database restore recovers.

#10756 records the deferral in `reclamation.md`, including the dedupe window and the never-pruned changelog, so the next attempt starts from both traps rather than rediscovering them. The branch stays at `feat/b2-reclamation` if this is picked up as a card.

PR #10756, opened 2026-08-11
chore(blobs): TAM-7037: record central orphan collection as deferred

dannash100:
### Changes

`specs/blob-storage/reclamation.md` shipped with the epic and nothing implements its central half. No card in the plan claimed it, and there is not one `spec: RECL` tag in the codebase, which is how it went unnoticed: every other spec area tags its implementation.

An implementation was written (#10753) and withdrawn. Review found that an age-based safety window structurally cannot cover the case it most needs to: admission is content-addressed and idempotent, and re-admitting content already held does not refresh the existing row's admission time. So a blob first seen a month ago, deduplicated onto by an upload a moment before that upload's reference commits, reads as a long-standing orphan. A pass in that gap hard-deletes the row, discards its parity and unlinks the file, and the reference commits pointing at content that is gone from the authoritative copy.

Rather than harden that on the eve of QA, the decision recorded here is to defer. The asymmetry decides it: not collecting orphans costs disk, collecting the wrong one costs a clinical attachment that restoring the database alone does not recover. Genuine orphans are also rarer than they look, since content repointed away stays named by its changelog entry, which RECL calls deliberate retention rather than a leak. That leaves interrupted operations, a failure-rate-bounded trickle, and the scrub's reconciliation pass adopts unregistered content into the registry so it stays visible in the store-size queries even while nothing reclaims it.

The spec section now says it is deferred and carries the two properties whatever implements it has to answer: the dedupe window above, and that the changelog is never pruned, so a liveness check reading it can only ever collect content that never acquired a reference at all. Both were found the hard way; writing them down is the point of this change.

The epic's test-case document is updated to match.

### Auto-Deploy

- [ ] **Deploy** 

### Review Hero

- [ ] **Run Review Hero**


dannash100:
🤖 Pushed straight to `workhorse/b2` as 4ba108898a instead.

PR #10777, opened 2026-08-12
fix: no-issue: fetch all medication preset labels, not just the first 25

rohan-bes:
### Changes

Fixes a silent-failure bug in medication dispensing: staff picking a "Preset label" outside the first 25 (sorted by code) would see the Label field quietly keep its auto-generated default instead of the preset's approved wording. Reported on FSM Demo (v2.61.1) with preset "SCABY"; present since the feature shipped in 2.60, surfacing once a deployment has more than 25 presets.

Root cause: the dropdown searches the backend live and lets staff select any preset, but the code that fills in the Label text (`resolvePresetLabelText`) looked the selected preset up against a locally cached list (`usePresetLabelsQuery`) that was fetched with the suggester's default page size of 25. Presets outside that first page weren't in the cache, so the lookup silently fell through to the fallback text.

Fix: pass `baseQueryParameters: { noLimit: true }` to the `medicationPresetLabel` suggester options, matching the existing pattern used by other suggesters that need their full result set (e.g. `PatientHistorySearch.jsx`). No backend changes needed — `noLimit` was already a supported parameter on the generic suggester route.

Added a unit test for `resolvePresetLabelText` with a 30-item list (asserting presets at index 25+ resolve), and a facility-server integration test confirming `?noLimit=true` returns the full preset list rather than the default-capped 25.

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests

### DIFF
--- a/packages/facility-server/__tests__/apiv1/Suggestions.test.js
+++ b/packages/facility-server/__tests__/apiv1/Suggestions.test.js
@@ -542,6 +542,35 @@ describe('Suggestions', () => {
         'W17',
       ]);
     });
+
+    it('should return every preset label when noLimit is set, bypassing the default page size', async () => {
+      await models.ReferenceData.destroy({
+        where: { type: REFERENCE_TYPES.MEDICATION_PRESET_LABEL },
+        force: true,
+      });
+
+      const presetCount = 30;
+      for (let index = 0; index < presetCount; index++) {
+        const code = String(index).padStart(3, '0');
+        await models.ReferenceData.create({
+          id: `medicationPresetLabel-${code}`,
+          type: REFERENCE_TYPES.MEDICATION_PRESET_LABEL,
+          code,
+          name: `Preset ${code}`,
+          visibilityStatus: VISIBILITY_STATUSES.CURRENT,
+        });
+      }
+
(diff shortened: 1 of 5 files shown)


rohan-bes:
Pushed a follow-up commit that removes the redundant lookup entirely, rather than just fixing its data cap.

The change handlers previously re-looked-up the selected preset's name in a separately fetched/cached list (the thing capped at 25). That lookup was actually redundant: `AutocompleteInput`'s change event already carries the full matched suggestion — it just collided with the component's own `name` prop (used for the Formik field name) and got silently overwritten. Renaming the formatter's field from `name` to `presetName` avoids that collision, so the handlers can read the selected preset's name straight off the change event. This removes the second data path (and cache) entirely, rather than just making it complete.

PR #10798, opened 2026-08-14
feat(settings): TAM-7059: suggest Anthropic models from the live API

dannash100:
### Changes
https://github.com/user-attachments/assets/0527ff5a-a0eb-4388-9964-f16f45680779

The two Anthropic model settings are free-text, so a typo or a retired model ID only surfaces when an AI feature fails at runtime. This turns both into dropdowns backed by the live Models API.

- `anthropicModel` and `anthropicVisionModel` suggesters on central only. They read `ai.anthropicApiKey` with `getSettingSecret` and call the API server-side, so the key never reaches the browser.
- Mounted in `central-server/app/suggestions.js`, not the shared router: the shared one runs on facility servers too, and the setting is central-only. That also keeps it out of `SUGGESTER_ENDPOINTS`, whose consistency check governs the shared router.
- The fast model field uses the vision-filtered endpoint. It serves `FORM_BUILDER_IMAGE`, so a model without `image_input`/`pdf_input` would fail only at the point of use.
- Gated on `read`/`Setting` rather than the open `flagPermissionChecked()` that `timeZone` uses, since each call spends the deployment API quota.
- The `/:id` lookup is deliberately unfiltered, so a saved model still renders its own name once it stops being offered rather than showing an empty field.
- The cache holds the pending request, not the result, so concurrent callers share one call; an empty result is never cached, or saving the key would leave the dropdown empty for the whole TTL.

Unreachable API or unsaved key gives an empty list rather than an error. 12 endpoint tests cover permissions, filtering, both failure modes, and that the key is not serialised to the client.

### Auto-Deploy

- [x] **Deploy** 

<details>
<summary>Options</summary>

- [ ] Artillery load test 
- [ ] Generate fake data 
- [ ] More data (20Gi) 
- [ ] Skip mobile build 
- [ ] Always build mobile 
- [ ] Stay up for 8 hours 
- [ ] Stay up for 24 hours 
- [ ] Stay up (no TTL) 
- [ ] Pause this deploy 

</details>

### Tests

- [ ] **Run E2E tests** 
- [ ] **Run DAST scan** 

### Review Hero

- [ ] **Run Review Hero** 

### Remember to...

- ...write or update tests


packages/settings/src/schema/central.ts
```diff
@@ -67,13 +72,15 @@ export const centralSettings = {
...
           defaultValue: '',
+          suggesterEndpoint: 'anthropicModel',
         },
         anthropicFastModel: {
           name: 'Anthropic fast model',
           description:
             'Optional faster Anthropic model for non-conversational tasks (PDF/image interpretation, structured tweaks, ProgramDefinition build). Falls back to anthropicModel when empty.',
           type: yup.string(),
           defaultValue: '',
+          suggesterEndpoint: 'anthropicVisionModel',
```
edmofro:
```suggestion
          suggesterEndpoint: 'anthropicFastModel',
```

packages/central-server/app/admin/adminRoutes.js
```diff
@@ -256,6 +257,8 @@ adminRoutes.put(
 
     await resolveSecretsForSave(Setting, settings, schema, scope, facilityId);
     await Setting.set('', settings, scope, facilityId);
+    // The suggester's list belongs to whichever key fetched it, so a key change must drop it.
+    if (getAtPath(settings, 'ai')) clearAnthropicModelCache();
```
edmofro: Don't love this injection of specific setting concerns into this wider zone

packages/central-server/app/suggestions.js
```diff
@@ -2,9 +2,14 @@ import { ensurePermissionCheck } from '@tamanu/shared/permissions/middleware';
 import { suggestions } from '@tamanu/shared/services/suggestions';
 import express from 'express';
 
+import { anthropicModelSuggestions } from './anthropicModelSuggestions.js';
+
 const suggestionsRoutes = express.Router();
 
 suggestionsRoutes.use(ensurePermissionCheck);
+// Central-only: it reads a central setting, and the shared router is mounted on
+// facility servers too.
```
edmofro: Comment feels unnecessary

packages/settings/src/schema/central.ts
```diff
@@ -67,13 +72,15 @@ export const centralSettings = {
           description: 'The Anthropic model to use for AI features',
           type: yup.string(),
           defaultValue: '',
+          suggesterEndpoint: 'anthropicModel',
         },
         anthropicFastModel: {
           name: 'Anthropic fast model',
           description:
             'Optional faster Anthropic model for non-conversational tasks (PDF/image interpretation, structured tweaks, ProgramDefinition build). Falls back to anthropicModel when empty.',
```
edmofro:
```suggestion
            'Optional faster Anthropic model for non-conversational tasks (PDF/image interpretation, structured tweaks, ProgramDefinition build). Falls back to anthropicModel when empty. A Haiku model is a good choice.',
```

packages/settings/src/schema/central.ts
```diff
@@ -67,13 +72,15 @@ export const centralSettings = {
           description: 'The Anthropic model to use for AI features',
```
edmofro:
```suggestion
          description: 'The Anthropic model to use for AI features. Opus is recommended, Sonnet is a cheaper alternative.',
```

packages/central-server/app/anthropicModelSuggestions.js
```diff
@@ -0,0 +1,117 @@
...
+  const models = await fetchModels(req.settings);
+  const model = models.find(({ id }) => id === req.params.id);
+  res.send(model ? { id: model.id, name: model.name } : { id: req.params.id, name: req.params.id });
+});
+
+anthropicModelSuggestions.get('/anthropicModel', listRoute(() => true));
+anthropicModelSuggestions.get('/anthropicModel/:id', lookupRoute);
+
+// The fast model serves the form builder's image context, so it has to read
+// images and PDFs; a model that can't would fail only at the point of use.
```
edmofro, line 110: Oh right this is why it's called vision. In that case I think it's misnamed in the UI, or there are two concepts being mixed together in a way they shouldn't...

edmofro (changes requested):
shouldn't have posted for review m8, I've still got opinions...

dannash100:
@edmofro no i love a human review 🙏

packages/central-server/app/anthropicModelSuggestions.js
edmofro: I dunno actually open to flexibility here, it does cut down the number of settings...but it's surprising to have it called one thing one place and another somewhere else.

packages/central-server/app/suggestions.js
dannash100: agree

packages/central-server/app/admin/adminRoutes.js
dannash100: Mabye something like this
[feat(settings): no-issue: apply AI settings without a restart](https://github.com/beyondessential/tamanu/pull/10807#top)

Move cache to ai store and listen to the valid settings, gives the bonus of not restart required. Bigger change tho

dannash100: Its a pattern we been wanting tho

